PR #2, opened 2019-06-21
Fix(messaging): dm channel create

tjerman:
There is no `members` field in request form, so the used value
was nil => only the creator is a member.


darh:
This is a fix of auto-generated file.

PR #9, opened 2019-08-07
Fix/message search before message

titpetric:
Please take a look.


titpetric:
New test code is related to the pagination bug (old deploy for crust chat),
Changes in drone/makefile re: unit tests are related to a sonyflake failing test (there were collisions within the same server due to how sonyflake ids are constructed),
Some changes are related to the sonyflake issue (message.Create), where I restructured the transaction to start/include mostly writes (reads happen before the transaction, which also limits the possibility of a deadlock)

titpetric:
And additionally gotest was removed due to overalls usage, so we didn’t have any benefit of gotest anymore, and we can revert to plain go test where needed.

darh:
I really not sure what was wrong here. I do not see any before-message-id related changes in the messaging service and/or repo besides the test. I've cherry-picking (c94a8a5c) `TestBeforeMessageID` code and it passes with the current code. 

----

I've also cherry-picked a33b0a7a and 58e45186.

Regarding other changes:
1. Makefile changes: if we're changing test tool we should change the variable, df17ddb8
2. changes re. message creation & transactions in service -- not sure why the change is needed. Should be part of a different PR.

PR #12, opened 2019-08-20
Migrate gplus -> google

tjerman:
External provider settings in DB should also be renamed

`auth.external.providers.gplus.*` -> `auth.external.providers.google.*`


tjerman:
@darh https://github.com/cortezaproject/corteza-server/pull/12/commits/1f6643bb002997c1838b272097aed787424038d9

system/db/schema/mysql/20190826085348.migrate-gplus-google.up.sql
```diff
@@ -0,0 +1,5 @@
+/* migrates existing credentials */
+UPDATE sys_credentials SET kind = 'google' WHERE kind = 'gplus';
+
+/* migrates existing settings. */
+UPDATE sys_settings SET name = REPLACE(name, '.gplus.', '.google.');
```
darh: Add `WHERE name LIKE 'auth.external.providers.gplus.%'`, just to be sure we're not breaking anything else.

PR #13, opened 2019-08-21
Allow channel/attach with URL to remote file

tjerman:


pkg/files.go
```diff
@@ -0,0 +1,57 @@
+package files
```
darh: Move this under `internal/store/download.go` for now. We'll move it under pkg later.

pkg/files.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if _, err := io.Copy(f, resp.Body); err != nil {
+		return nil, err
+	}
+
+	return f, nil
+}
+
+// ExtractExtFromURL extracts file's ext from given url
+func ExtractExtFromURL(url string) (string, error) {
```
darh: Let's not just trust external sources :)
Mimetype is a bit more reliable.

PR #16, opened 2019-08-26
WIP: Record import

tjerman:
Allow record import from structured files.


api/compose/spec.json
```diff
@@ -661,6 +661,66 @@
...
+            },
+            {
+                "name": "importRun",
+                "path": "/import",
+                "method": "PATCH",
+                "title": "Run record import",
+                "parameters": {
+                    "post": [
+                        {
+                            "name": "sessionID",
```
darh: move this to path, so `/import/{sessionID}`

api/compose/spec.json
```diff
@@ -661,6 +661,66 @@
...
+            },
+            {
+                "name": "importProgress",
+                "path": "/import",
+                "method": "GET",
+                "title": "Get import progress",
+                "parameters": {
+                    "get": [
+                        {
+                            "name": "sessionID",
```
darh: move this to path, so `/import/{sessionID}`

compose/internal/service/access_control.go
```diff
@@ -170,6 +170,17 @@ func (svc accessControl) CanRunAutomationTrigger(ctx context.Context, r *automat
 	return svc.can(ctx, types.AutomationScriptPermissionResource.AppendID(r.ID), "run", permissions.Allowed)
 }
 
+// @todo actual controll
+func (svc accessControl) CanReadImportSession(ctx context.Context) bool {
```
darh: No need for AC for the record importer.
If the user can create records (on module) she can import them.
Session access control is done solely on ownership.

compose/rest/record.go
```diff
@@ -167,6 +180,108 @@ func (ctrl *Record) Upload(ctx context.Context, r *request.RecordUpload) (interf
...
+	// determine decoder
+	switch strings.ToLower(ext) {
+	case "json", "jsonl", "ldjson", "ndjson":
+		recordDecoder = decoder.NewStructuredDecoder(json.NewDecoder(f))
+
+	case "csv":
+		recordDecoder = decoder.NewFlatReader(csv.NewReader(f))
+
+	// @todo...
+	case "xlss":
```
darh: `xlsx`?

compose/internal/service/session.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	sessionAccessController interface {
+		CanReadImportSession(context.Context) bool
+		CanCreateImportSession(context.Context) bool
+		CanDeleteImportSession(context.Context) bool
+	}
+
+	SessionService interface {
+		With(ctx context.Context) SessionService
```
darh: Please do not follow this pattern (`With(context.Context)`) anymore. This was a design mistake. Will be refactored soon.

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

More: https://golang.org/pkg/context/

See how AccessControl, AutomationScript, AutomationTrigger are done.

compose/decoder/decoder.go
```diff
@@ -0,0 +1,133 @@
...
+	flatReader struct {
+		r      FlatReader
+		header []string
+		more   bool
+	}
+
+	structuredDecoder struct {
+		header []string
+		d      StructuredDecoder
+		buf    []map[string]interface{}
```
darh: We're storing all entries from the file in this buffer? Do I understand this correctly?

If yes -- this is not how this should be implemented, no reason to hold it in a buffer.

compose/decoder/decoder.go
```diff
@@ -0,0 +1,133 @@
...
+
+	dec.Get(func(rtr []string) {
+		dec.header = rtr
+	})
+
+	return dec.header
+}
+
+// structured decoder
+func NewStructuredDecoder(d StructuredDecoder) *structuredDecoder {
```
darh: You'll probably need decoder (for converting one file line to structure, something like `json.Decoder`) AND reader (io.Reader). Then, structured decoder should pass reader through something that knows how to stop at new-line and feed what was read into decoder.

compose/rest/record.go
```diff
@@ -167,6 +180,108 @@ func (ctrl *Record) Upload(ctx context.Context, r *request.RecordUpload) (interf
...
+
+	// Make session
+	ses := types.RecordImportSession{
+		SessionID:    factory.Sonyflake.NextID(),
+		ImportFields: recordDecoder.Header(),
+		NamespaceID:  r.NamespaceID,
+		ModuleID:     r.ModuleID,
+	}
+
+	ctrl.session.SetByID(ses.SessionID, &recordImport{decoder: recordDecoder, session: ses})
```
darh: 1. Move the responsibility of generating session-IDs out of controller into service
2. passing non-exported structs and accept it as interface{}, not something I would be proud of. Pass params in as directly as possible (along with decoder)

compose/internal/service/service.go
```diff
@@ -48,11 +48,12 @@ var (
 	DefaultAutomationRunner automationRunner
 
 	DefaultNamespace NamespaceService
+	DefaultSession   SessionService
 	DefaultRecord    RecordService
 	DefaultModule    ModuleService
 	DefaultChart     ChartService
 	DefaultPage      PageService
-
+	// DefaultTrigger      TriggerService
```
darh: This looks like a relic from before latest automation feature. Remove.

compose/internal/service/session.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+type (
+	session struct {
```
darh: This is too general, rename to `ImportSession`

compose/internal/service/record.go
```diff
@@ -184,6 +218,49 @@ func (svc record) Find(filter types.RecordFilter) (set types.RecordSet, f types.
...
+	if ses.Progress.StartedAt != nil {
+		return errors.New("Unable to start import: Import session already active")
+	}
+
+	sa := time.Now()
+	ses.Progress.StartedAt = &sa
+	ssvc.SetRecordByID(svc.ctx, ses.SessionID, 0, 0, nil, &ses.Progress, nil)
+
+	return svc.db.Transaction(func() (err error) {
+		err = (*ses.Decoder).Records(ses.Fields, func(mod *types.Record) error {
```
tjerman: @darh is something like this (`(*ses.Decoder).Records`) ok; it doesn't seem ok to me

compose/internal/service/service.go
```diff
@@ -130,6 +130,7 @@ func Init(ctx context.Context, log *zap.Logger, c Config) (err error) {
 		)
 	}
 
+	DefaultSession = ImportSession()
```
darh: => `DefaultImportSession`

compose/internal/service/record.go
darh: nope.

pkg/helpers/files.go
```diff
@@ -0,0 +1,54 @@
+package helpers
```
darh: https://blog.golang.org/package-names

nope.

How about just mime?

compose/rest/record.go
```diff
@@ -167,6 +171,106 @@ func (ctrl *Record) Upload(ctx context.Context, r *request.RecordUpload) (interf
...
+	switch strings.ToLower(ext) {
+	case "json", "jsonl", "ldjson", "ndjson":
+		recordDecoder = decoder.NewStructuredDecoder(json.NewDecoder(f))
+
+	case "csv":
+		recordDecoder = decoder.NewFlatReader(csv.NewReader(f))
+
+	// @todo...
+	case "xlsx":
+		// recordDecoder = encoder.NewExcelizeDecoder()
```
darh: nil pointer error.

return error - not implemented or something.

compose/internal/service/import_session.go
```diff
@@ -0,0 +1,118 @@
...
+	var ris *RecordImportSession
+
+	if i >= 0 {
+		ris = svc.records[i]
+	} else {
+		ris = &RecordImportSession{
+			SessionID: factory.Sonyflake.NextID(),
+			CreatedAt: time.Now(),
+		}
+		svc.records = append(svc.records, ris)
```
darh: This (modifying same variable from multiple goroutines) is not safe.
Add locking (see internal/permissions/service.go for examples)

compose/internal/service/service.go
tjerman: Ee... slipped by

PR #29, opened 2019-10-08
Feature settings

tjerman:
Add settings services for messaging & compose.


api/compose/spec.json
```diff
@@ -1854,5 +1854,76 @@
...
+        "title": "Settings",
+        "path": "/settings",
+        "entrypoint": "settings",
+        "authentication": [],
+        "struct": [{
+            "imports": [
+                "sqlxTypes github.com/jmoiron/sqlx/types"
+            ]
+        }],
+        "apis": [{
```
darh: `list` endpoint is missing? Intentionally?

compose/service/settings.go
```diff
@@ -0,0 +1,97 @@
+package service
+
+import (
+	"context"
+
+	_ "github.com/joho/godotenv/autoload"
```
darh: You do not need this here anymore. Was used for auth settings autodiscovery on system (and this import shoud/cloud be moved there too)

PR #77, opened 2020-02-25
Feature rec val validators

darh:
Please take a look (all but generated event/*go) files

### DIFF
--- a/compose/service/record.go
+++ b/compose/service/record.go
@@ -826,18 +826,16 @@ func (svc record) setDefaultValues(m *types.Module, vv types.RecordValueSet) (ou
 
 	for _, f := range m.Fields {
 		if f.DefaultValue == nil {
-			return nil
+			continue
 		}
 
 		for i, dv := range f.DefaultValue {
 			// Default values on field are (might be) without field name and place
 			if !out.Has(f.Name, uint(i)) {
 				out = append(out, &types.RecordValue{
-					Name:      f.Name,
-					Value:     dv.Value,
-					Ref:       0,
-					Place:     uint(i),
-					DeletedAt: nil,
+					Name:  f.Name,
+					Value: dv.Value,
+					Place: uint(i),
 				})
 			}
 		}
--- a/compose/service/record.go
+++ b/compose/service/record.go
@@ -326,6 +326,43 @@ func (svc record) Export(filter types.RecordFilter, enc Encoder) error {
 }
 
 func (svc record) Create(new *types.Record) (rec *types.Record, err error) {
+	var invokerID = auth.GetIdentityFromContext(svc.ctx).Identity()
+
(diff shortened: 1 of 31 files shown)


compose/service/record.go
```diff
@@ -326,6 +335,43 @@ func (svc record) Export(filter types.RecordFilter, enc Encoder) error {
 }
 
 func (svc record) Create(new *types.Record) (rec *types.Record, err error) {
+	var invokerID = auth.GetIdentityFromContext(svc.ctx).Identity()
+
+	// Runs value sanitization, sets values that should used
```
tjerman: `should used` -> `should be used` :')

compose/service/values/sanitizer.go
```diff
@@ -86,42 +71,42 @@ func (s sanitizer) Run(m *types.Module, vv types.RecordValueSet) (out types.Reco
...
+			//	v = s.sRecord(v, f, m)
+			//case "select":
+			//	v = s.sSelect(v, f, m)
+			//case "string":
+			//	v = s.sString(v, f, m)
+			//case "url":
+			//	v = s.sUrl(v, f, m)
+			//case "user":
+			//	v = s.sUser(v, f, m)
 		}
```
tjerman: What's up with all of these?

compose/service/values/shared.go
```diff
@@ -0,0 +1,37 @@
...
+	"regexp"
+	"time"
+)
+
+const (
+	strBoolTrue  = "1"
+	strBoolFalse = "0"
+
+	datetimeInputFormatDate = "2006-01-02"
+	datetimeInputFormatTime = "15:04:05"
```
tjerman: should we not handle TZ?

compose/service/values/shared.go
```diff
@@ -0,0 +1,37 @@
...
+	fieldOpt_Datetime_onlyPastValues   = "onlyPastValues"
+
+	fieldOpt_Number_precision = "precision"
+
+	fieldOpt_Url_onlySecure = "onlySecure"
+)
+
+var (
+	// value resembles something that can be true
+	truthy = regexp.MustCompile(`^(t(rue)?|y(es)?|1)$`)
```
tjerman: add `i` (case insensitive) flag for good measure

compose/service/values/sanitizer.go
darh: placeholders.

compose/service/record.go
darh: ty.

compose/service/values/shared.go
darh: we do (in sanitizer), but these are internal datetime formats

compose/service/values/shared.go
darh: we do  it on the value.

PR #78, opened 2020-02-28
POC data migration system

tjerman:


pkg/app/options/corredor.go
```diff
@@ -45,7 +45,7 @@ func Corredor() (o *CorredorOpt) {
 		ListRefresh:        time.Second * 5,
 		Log:                false,
 
-		TlsCertEnabled: true,
+		TlsCertEnabled: false,
```
darh: No!

you have  `CORREDOR_CLIENT_CERTIFICATES_ENABLED` to control this.

pkg/migrate/main.go
```diff
@@ -0,0 +1,320 @@
+package migrate
```
darh: we have another thing we call "migrate" (for db migrations)
so, try to think of a better name .

also, how compose-oriented is this? 
we are dealing 99% with records and 1% with users, correct?
can it handle other types of data?
can it handle other formats (yaml)?

pkg/migrate/types/migrateable.go
```diff
@@ -0,0 +1,18 @@
...
+var (
+	SfDateTime = "2006-01-02 15:04:05"
+)
+
+type (
+	Migrateable struct {
+		Name string
+		Path string
+
+		Source *os.File
```
darh: how about io.Reader instead of os.File?

pkg/migrate/main.go
tjerman: At the moment it's as you said; mostly Compose oriented, but.. It should be fairly trivial to extend this to an arbitrary structure and arbitrary data source

PR #88, opened 2020-04-23
Respect Number field's specified precision

tjerman:
This enables the formatter to format numbers based on the field's precision.

### DIFF
--- a/compose/service/values/sanitizer.go
+++ b/compose/service/values/sanitizer.go
@@ -1,10 +1,11 @@
 package values
 
 import (
-	"github.com/cortezaproject/corteza-server/compose/types"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/cortezaproject/corteza-server/compose/types"
 )
 
 type (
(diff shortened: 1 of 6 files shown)


tjerman:
@darh below at line 210 the value's type was a bit random (sometimes string, sometimes float64) and I can't quite put my finger on why; any idea?

compose/service/values/sanitizer.go
```diff
@@ -192,6 +193,60 @@ func (sanitizer) sDatetime(v *types.RecordValue, f *types.ModuleField, m *types.
...
 }
 
+func (sanitizer) sNumber(v *types.RecordValue, f *types.ModuleField, m *types.Module) *types.RecordValue {
+	// No point in continuing
+	if v.Value == "" || f.Options == nil {
+		return v
+	}
+
+	// Default to 0 for consistency
+	if f.Options["precision"] == nil || f.Options["precision"] == "" {
```
darh, line 203: use `Int64Def()` fn on options

PR #90, opened 2020-05-18
Add support for batch record operations

tjerman:
Allow bulk record operations in a single request & transaction.
Feature required for the upcoming Record List feature.

The implementation is BC with "old" API clients.

TODO:
* [x] Checkup/cleanup,
* [x] tests.

### DIFF
--- a/compose/rest/record.go
+++ b/compose/rest/record.go
@@ -5,11 +5,12 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
-	"github.com/cortezaproject/corteza-server/compose/service/values"
-	"github.com/cortezaproject/corteza-server/pkg/payload"
 	"net/http"
 	"path"
 	"strings"
+
+	"github.com/cortezaproject/corteza-server/compose/service/values"
+	"github.com/cortezaproject/corteza-server/pkg/payload"
 
 	"github.com/titpetric/factory/resputil"
 
@@ -33,13 +34,15 @@ type (
 	recordPayload struct {
 		*types.Record
 
+		Records types.RecordSet `json:"records,omitempty"`
+
 		CanUpdateRecord bool `json:"canUpdateRecord"`
 		CanDeleteRecord bool `json:"canDeleteRecord"`
 	}
 
 	recordSetPayload struct {
-		Filter types.RecordFilter `json:"filter"`
-		Set    []*recordPayload   `json:"set"`
+		Filter *types.RecordFilter `json:"filter,omitempty"`
+		Set    []*recordPayload    `json:"set"`
 	}
 
 	Record struct {
@@ -103,7 +106,7 @@ func (ctrl *Record) List(ctx context.Context, r *request.RecordList) (interface{
 
 	rr, filter, err := ctrl.record.With(ctx).Find(rf)
 
-	return ctrl.makeFilterPayload(ctx, m, rr, filter, err)
+	return ctrl.makeFilterPayload(ctx, m, rr, &filter, err)
 }
 
 func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{}, error) {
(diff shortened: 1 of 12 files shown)


compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
 	return ctrl.makePayload(ctx, m, record, err)
 }
 
-func (ctrl *Record) Create(ctx context.Context, r *request.RecordCreate) (interface{}, error) {
+func (ctrl *Record) Create(ctx context.Context, rc *request.RecordCreate) (interface{}, error) {
```
darh, line 131: keep this named `r` -- for consistency across all controllers.

compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
...
+		})
+	}
+
+	// Initialize sub records for creation, if defined
+	if rc.Records != nil {
+		for _, br := range rc.Records {
+			for _, rr := range br.Set {
+				// Default to current module and namespace; can be overwritten
+				if rr.NamespaceID == 0 {
+					rr.NamespaceID = rc.NamespaceID
```
darh, line 162: There is no benefit in supporting cross-namespace behaviour right now.

compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
...
+			Values:      ru.Values,
+		}
+		batch = append(batch, &types.BulkRecordOperation{
+			Record:    rr,
+			Operation: types.OperationTypeUpdate,
+		})
+	}
+
+	// Initialize sub records for creation, if defined
+	if ru.Records != nil {
```
darh, line 238: no need to check for `nil`  as it does not reak `for`

compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
...
+		for _, br := range ru.Records {
```
darh, line 239: should/could this logic be part of `BulkRecordSet` (not as generated func, as additional custom on set)?

compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
...
+	}
+
+	rs, err := ctrl.record.With(ctx).Operate(batch...)
 
 	if rve, is := err.(*types.RecordValueErrorSet); is && !rve.IsValid() {
 		return ctrl.handleValidationError(rve), nil
 	}
 
-	return ctrl.makePayload(ctx, m, record, err)
+	if ru.Records == nil {
```
darh, line 284: This will hurt us. Bad.

Is there anyway to keep consistent return value?

compose/service/record.go
```diff
@@ -86,6 +86,7 @@ type (
 
 		Create(record *types.Record) (*types.Record, error)
 		Update(record *types.Record) (*types.Record, error)
+		Operate(pl ...*types.BulkRecordOperation) (types.RecordSet, error)
```
darh, line 89: `pl`? payload? better just `oo` (as-in operations )

compose/service/record.go
```diff
@@ -354,71 +355,225 @@ func (svc record) Export(filter types.RecordFilter, enc Encoder) error {
...
-			return
-		}
-
-		if !svc.ac.CanCreateRecord(svc.ctx, m) {
-			return ErrNoCreatePermissions.withStack()
-		}
+			// Handle any pre processing, such as defining parent recordID.
+			// This functions are defined by the system, so there is no need for validation
+			//
+			// @todo This should be moved into svc.create(...) and svc.update(...) functions,
```
darh, line 371: why it should be moved? pre-process is (as I understand it) part of bulk op)

compose/service/record.go
```diff
@@ -354,71 +355,225 @@ func (svc record) Export(filter types.RecordFilter, enc Encoder) error {
...
 
-func (svc record) Create(new *types.Record) (rec *types.Record, err error) {
-	var invokerID = auth.GetIdentityFromContext(svc.ctx).Identity()
-
+// Operate handles provided set of bulk record operations.
+// It's able to create or update records in a single transaction.
+//
+// Feature used mainly by Record Lines, but should be used when ever we wish to perform
+// multiple operations over records.
+func (svc record) Operate(pl ...*types.BulkRecordOperation) (rec types.RecordSet, err error) {
```
darh, line 363: how about just `Bulk`?

compose/rest/record.go
```diff
@@ -126,51 +128,163 @@ func (ctrl *Record) Read(ctx context.Context, r *request.RecordRead) (interface{
...
+				// parent record is created.
+				if br.RefField != "" {
+					if rc.Values == nil {
+						return nil, errors.New("missing parent record definition")
+					}
+
+					b.PreProcess = func(m, r *types.Record) (*types.Record, error) {
+						r.Values = r.Values.Set(&types.RecordValue{
+							Name:  br.RefField,
+							Value: strconv.FormatUint(m.ID, 10),
```
darh, line 188: don't forget to set `Ref`

compose/service/record.go
tjerman: This should become a post-process, that happens right before the values are saved to the DB.
I temporarily changed it to pre-process, because I can't think of an elegant way to do this at the moment.

compose/rest/record.go
tjerman: If we wish to return all the changed records (which makes sense), I can't think of a way to do so without breaking BC.
Not my proudest solution, but it did seem ok for now.

Because of this, I might prefer to define a new endpoint, that we can migrate to at a later point in time.

tjerman: Hmm... The situation is as follows. As is, the record is provided directly in the `response` on the very root.

If we just add `records RecordSet`, we can get a weird structure, that I don't agree on:
```
{
  response: {
    recordID: '...',
    moduleID: '...',
    values: [...],
    ...

    records: [...],
  }
}
```

tjerman: We could, however, provide either a single record (as we do now) or a list of records.
so either
```
{
  response: {
    recordID: '...',
    moduleID: '...',
    values: [...],
    ...
  }
}
```

or

```
{
  response: [{
    recordID: '...',
    moduleID: '...',
    values: [...],
  }]
}
```

But this is fairly similar to the original solution...

PR #94, opened 2020-05-28
Allow system field timezone formatting

tjerman:
Specified via `timezone` query param, that has to be a valid ISO timezone, such as `Pacific/Pago_Pago`.


darh (approved):

tjerman:
https://github.com/cortezaproject/corteza-server/pull/97

PR #96, opened 2020-05-29
Improve bulk record field error handling

tjerman:
# Flow

## client:

on form submit:
* generate a sequential set of ID's for all new records
* FE validation
* hit API

on errors:
* use id provided in the meta to determine if this item contains errors

## API:

on preproc:
* if record has no ID, create a sequential ID from the provided prefix and a counter
* if record has ID, use that

on error:
* identify the error with the above-mentioned ID

# Dev notes

* I refer to these ID's as an item ID, since they are not consistently record ID's
* I construct a fresh set on each form submit, to make the API's life easier when determining what ID a new record should have. Since errors are reset on each form submit, this is perfectly valid.


compose/service/record.go
```diff
@@ -537,6 +536,12 @@ func (svc record) Bulk(oo ...*types.BulkRecordOperation) (rr types.RecordSet, er
 		return nil
 	})
 
+	// no need for another round of action recording, since it was already
```
darh, line 539: What is recorded here is the fact that this was a bulk op. So, let's loose this check.

compose/types/rest.go
```diff
@@ -1,8 +1,14 @@
 package types
 
+import (
```
darh, line 3: Can we move this somewhere else? Why "rest"? This part of the app should not be aware of that layer at all.

compose/types/rest.go
```diff
@@ -1,8 +1,14 @@
...
+	"fmt"
+	"strconv"
+)
+
 type (
 	BulkRecord struct {
```
darh, line 9: `RecordBulk` please

PR #98, opened 2020-06-30
Fix page reordering on root node

tjerman:
When operating on root node, the parentID is 0; the page is not found so it fails.


darh (approved):

darh:
merged.

PR #108, opened 2020-09-10
Implement a base for the pairing process

tjerman:


federation/rest.yaml
```diff
@@ -5,79 +5,107 @@
...
-  - title: Node identity
-    path: "/node/identity"
-    entrypoint: identity
-    authentication: []
+  - title: Federation node handshake
+    path: "/nodes/{nodeIDA}/handshake"
+    entrypoint: nodeHandshake
+    parameters:
+      path:
+      - name: nodeIDA
```
darh: why bothering with A/B here?

federation/rest/node.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func (ctrl Node) Create(ctx context.Context, r *request.NodeCreate) (interface{}, error) {
+	if r.NodeURI == "" && (r.Domain == "" || r.Name == "") {
+		return nil, ErrInvalidNodeCreateParams
+	}
+
+	n := &types.Node{}
+	if r.NodeURI != "" {
```
darh, line 37: why are we doing this (biz logic) here? 
controllers should only translate payloads into internal structures and hand them off to service.

decision model: if code will need to be duplicated for CLI/gRPC/graphql/***... it does not belong here...

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+		Create(n *types.Node, sharedDomain string) (*types.Node, error)
+		Pair(nodeID uint64) error
+		HandshakeInit(sourceNodeID, nodeIDB uint64, nodeURI, token string) error
+		HandshakeConfirm(nodeID uint64) error
+		HandshakeComplete(nodeID uint64, token string) error
+	}
+)
+
+func Node(ctx context.Context) NodeService {
+	rand.Seed(time.Now().UnixNano())
```
darh: Why here? isn't this something we should do at boot time?

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+	return (&node{
+		store: DefaultNgStore,
+
+		actionlog:    DefaultActionlog,
+		tokenEncoder: auth.DefaultJwtHandler,
+	}).With(ctx)
+}
+
+func (svc node) With(ctx context.Context) NodeService {
```
darh: Remove this and follow the standard go-way with context.Context as 1st arg in the functions (where needed)
This (with(ctx) pattern is OBSOLETE, DANGEROUS and MUST NOT be used when implementing new services.

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+	// Handle fenedrated user
+	uHandle := "federation_" + strconv.FormatUint(n.ID, 10)
+	u, _ := svc.sysUser.FindByHandle(uHandle)
+
+	if u == nil {
+		var err error
+
+		// Create a system user to service this node
+		u, err = svc.sysUser.Create(&stypes.User{
+			Email:          strconv.FormatUint(n.ID, 10) + "@federation.corteza",
```
darh: how about if we use `"federation@"  + domainName`?

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+			Handle:         uHandle,
+			Kind:           stypes.FederationUser,
```
darh: No, let's not add user kinds like this.
IF we only need this to be diferentiated from other users, we'll solve this later with labels.

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+			EmailConfirmed: true,
```
darh: Really?

federation/service/node.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Handle fenedrated user
+	uHandle := "federation_" + strconv.FormatUint(n.ID, 10)
+	u, _ := svc.sysUser.FindByHandle(uHandle)
+
+	if u == nil {
+		var err error
+
+		// Create a system user to service this node
+		u, err = svc.sysUser.Create(&stypes.User{
```
darh: Same comments as above + can we tuck fed. user creation logic into one single fn?

federation/util/util.go
```diff
@@ -0,0 +1,89 @@
...
+	spew.Dump(u)
+
+	return u.String()
+}
+
+func GenerateOTT() string {
+
+	b := make([]rune, TokenLength)
+	for i := range b {
+		b[i] = tokenChars[rand.Intn(len(tokenChars))]
```
darh: see `pkg/rand`

store/rdbms/roles.go
```diff
@@ -18,7 +19,7 @@ func (s Store) convertRoleFilter(f types.RoleFilter) (query squirrel.SelectBuild
 	}
 
 	if f.MemberID > 0 {
-		query = query.Where(squirrel.Expr("rl.ID IN (SELECT rel_role FROM sys_role_member AS m WHERE m.rel_user = ?)", f.MemberID))
+		// query = query.Where(squirrel.Expr("rl.ID IN (SELECT rel_role FROM sys_role_member AS m WHERE m.rel_user = ?)", f.MemberID))
```
darh: yap, I know, fixed but needs to be pushed.

darh (changes requested):

federation/rest.yaml
tjerman: Made sense when I was writing this, but now it doesn't anymore...

federation/rest/node.go
tjerman: This is what I was trying to achieve, by converting some input into some standard structure. It didn't make sense to create another intermediate structure to store unified data that is later converted to a Node in the service.

federation/service/node.go
tjerman: I had that initially, but it would cause issues if you would run multiple instances from the same domain. That is why I've done it like so.

federation/service/node.go
tjerman: From my point of view, it would make sense to have a new user kind for such a thing. I think that the underlying logic would be cleaner if we solve this with user kinds instead of labels.

- we know what user is a federated user from the kind,
- we can prevent explicit logins for users that are used in federation.

store/rdbms/roles.go
tjerman: My bad -- a commit lapsus...

federation/service/node.go
darh: I'm not 100% sure yet what we're going to do with service users, bots (if ever) etc... 
For now, this kind "feature" can easily be dropped as we do not use it anywhere.

federation/rest/node.go
darh: but pairing URI is not something that only comes in via HTTP proto, correct?

PR #112, opened 2020-09-30
Implement support for labels on modules

darh:
This is POC and intended for brainstorming only

It introduces ability to add labels on resources and use labels as filters.
ATM it labels are only implemented on compose Modules, plan is to add it to almost every other resource (where reasonable)

Examples of possible usages:
 - group resources (so we do not have to reply on handle prefixes)
 - mark federated records
 - store import references (ext ID)
 - mark resources that could be preloaded in Corredor
 - marking dependencies by automation scripts, workflow
 - ....

I'm also experimenting with ModuleLabelResourceType format, not sure were it will land (this is a bigger problem that touches RBAC resources as well...) might just be that we'll end up (for now) with RBAC-like format for label resources.

### DIFF
--- a/compose/rest.yaml
+++ b/compose/rest.yaml
@@ -11,6 +11,7 @@ endpoints:
   path: "/namespace"
   authentication: []
   imports:
+    - github.com/cortezaproject/corteza-server/pkg/label
     - sqlxTypes github.com/jmoiron/sqlx/types
     - time
   apis:
@@ -31,6 +32,10 @@ endpoints:
       - type: uint
         name: limit
         title: Limit
+      - type: map[string]string
+        name: labels
+        title: Labels
+        parser: label.ParseStrings
       - type: string
         name: pageCursor
         title: Page cursor
@@ -47,6 +52,10 @@ endpoints:
         name: name
         required: true
         title: Name
+      - type: map[string]string
+        name: labels
+        title: Labels
+        parser: label.ParseStrings
       - type: string
         name: slug
         required: false
@@ -96,6 +105,10 @@ endpoints:
         name: meta
         required: true
         title: Meta data
+      - type: map[string]string
+        name: labels
+        title: Labels
+        parser: label.ParseStrings
       - type: "*time.Time"
         name: updatedAt
         required: false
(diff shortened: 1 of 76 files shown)


tjerman (commented):
In the `Module struct` you define labels as `Labels map[string]string 'json:"labels,omitempty"'`.
I would change this to `Labels []Label 'json:"labels,omitempty"'`.

For example, the second version enables me to define multiple values for the same label "type" -- a module can now be in multiple groups.

___

In the `store/labels.yaml` I would rename `Kind` to `Resource` (until I've read the comment, I thought this was a label kind such as a `module_group` or `federated_node`).

I would rename `Name` to `Kind` to indicate what this label defines.
This would force us to strictly define the supported set of labels and help us not abuse this feature.

darh:
@tjerman I have a feeling you're not reading this right -- might be due to badly chosen var/prop names or poor docs in this POC.

`Label` struct is general -- and will be reused for ALL kinds of resources -- so labels for modules will have kind=module and labels for records will have kind=record etc.
ResourceID is self explanatory.

Name and value are the actual pairs that go to a specific resource (module, record...). 
And there is intentional limitation of having unique labels per name per resource.

Having multiple labels with same name would introduce more complexity: we would need label IDs to properly manage the labels, updating/creating would not be as simple as using key-val struct etc.

Yes, we loose the ability to "define multiple groups" (from your example), but such examples might raise questions if we should no implement this "multiple groups" directly (w/o labels) in the first place.

Labels are just that -- simple pieces of data that allow us some filtering and labeling of our resources for distinction and faster/easier lookups.

tjerman:
@darh 
Do we then really need to have a `{ k: v }` pair for each label? Wouldn't a simple `v` be enough? Do you have a real world example where this would be useful? One place where this _might_ be helpful is for system label translations, but that could also be solved with the simplified structure.

With the additional simplification, you could have something in the lines of:

```js
// -----------
{
  "moduleID": "...",
  "name": "M1",
  "handle": "M1_h",
  "labels": [ "federated", "crm", "case management" ],
  ...
}
// -----------
{
  "moduleID": "...",
  "name": "M2",
  "handle": "M2_h",
  "labels": [ "crm", "case management" ],
  ...
}
// -----------
{
  "moduleID": "...",
  "name": "M3",
  "handle": "M3_h",
  "labels": [],
  ...
}
// -----------
```

To go back to the module, or rather resource grouping, we could then have this as a _pure_ front-end feature instead of something the server should care about.

PR #115, opened 2020-10-05
Implement federation node pairing flow

tjerman:
ref: https://github.com/cortezaproject/corteza-server/pull/108

The entire flow is described [in the docs](https://docs.cortezaproject.org/corteza-docs/2020.9/rfc/federation/index.html#rfc:federation:node-pair)


federation/service/service.go
```diff
@@ -38,6 +40,8 @@ var (
 
 	DefaultActionlog actionlog.Recorder
 
+	DefaultNode NodeService
+
 	// wrapper around time.Now() that will aid service testing
 	now = func() time.Time {
```
darh: I suggest you change rval to *time.Time, it's easier to convert to time.Time if needed then other way around...

federation/service/service.go
```diff
@@ -5,13 +5,15 @@ import (
 	"time"
 
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
+	"github.com/cortezaproject/corteza-server/pkg/auth"
 	"github.com/cortezaproject/corteza-server/pkg/healthcheck"
 	"github.com/cortezaproject/corteza-server/pkg/id"
 	"github.com/cortezaproject/corteza-server/pkg/objstore"
```
darh, line 11: Remove objectstore (file upload), no need for that in federation

federation/service/node.go
```diff
@@ -0,0 +1,213 @@
...
+	stypes "github.com/cortezaproject/corteza-server/system/types"
+	"github.com/davecgh/go-spew/spew"
+)
+
+var (
+	// This is temporary for local testing
+	tmpNodeStore = make(types.NodeSet, 0)
+
+	ErrorNodeNotFound   = errors.New("pair: node not found")
+	ErrorInvalidNodeURI = errors.New("pair: invalid node uri provided")
```
darh: all service-lever errors should be moved to action yaml def.

darh (commented):
minor things...

PR #116, opened 2020-10-06
WIP

tjerman:
`u` was never assigned, so it remained `nil`. I will take a better look for a proper solution/if there are other similar issues a bit later.


darh:
Fixed with 9b3ba316

PR #120, opened 2020-10-29
Fix record values with denied fields

darh:
Scenario:
 - values already exist on the record
 - record is updated with users that can not update all fields
 - only values for updatable fields are provided

Access control was moved from generalValueSetValidation to procCreate/procUpdate
to allow return of record value error set.

Existing values from fields with update-deny set are copied to updating set
to ensure proper handling by RVS's Merge fn().

This will be:
 - merged back to 2020.9.x branch
 - merged to 2020.12.x branch


tjerman (approved):
lgtm

PR #124, opened 2020-11-12
2020.12.x feature expr

darh:
Please take a look.


compose/types/module_field_options.go
```diff
@@ -13,8 +13,13 @@ type (
 )
 
 const (
+	moduleFieldOptionExpression         = "expression"
 	moduleFieldOptionIsUnique           = "isUnique"
 	moduleFieldOptionIsUniqueMultiValue = "isUniqueMultiValue"
+
+	moduleFieldNumberOptionPrecision         = "precision"
+	moduleFieldNumberOptionPrecisionMin uint = 0
+	moduleFieldNumberOptionPrecisionMax uint = 6
```
tjerman, line 23: Should we define this globally? I don't see why we would want to have different precision limits over different parts of the system

compose/service/values/expr.go
```diff
@@ -0,0 +1,113 @@
...
+func Expression(ctx context.Context, m *types.Module, r *types.Record, old *types.Record, rve *types.RecordValueErrorSet) {
+	var (
+		exprParser = parser()
+
+		scope = make(map[string]interface{})
+
+		reserved = map[string]bool{
+			"new": true,
+			"old": true,
+		}
```
tjerman, line 44: You could ditch this and just overwrite the reserved indexes with those values at the end

PR #127, opened 2020-11-25
Fix in-path sink signature matching

tjerman:
Merge into 2020.9.x, 2020.12.x

The eventbus payload contained the trailing /__sink=..., so path matching failed.


darh (approved):
:+1:

tjerman:
Merged

PR #132, opened 2020-12-11
Federation feature

petergrlica:
What has been currently done:
 - shared, exposed modules, nodes and module mappings REST endpoints
 - RBAC permissioning
 - structure and data sync

What to check more thoroughly:
 - sync_worker_structure and sync_worker_data
 - processer_structure and processer_data
 - sync service
 - syncer service


federation/importer/default.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"context"
+	"io"
+)
+
+// Import performs standard import procedure with default services
+func Import(ctx context.Context, ff ...io.Reader) (err error) {
+	return nil
+}
```
tjerman: You can just remove this file; importing was restructured

provision/000_base/roles.yaml
```diff
@@ -5,3 +5,4 @@ roles:
   # placeholder, admin role is created programmatically
   # with a fixed ID before configs are processed
   - admins
+  - federation
```
tjerman, line 8: Shouldn't we also define the ID inside `pkg/rbac/permissions.go`?
Also add it here `pkg/provision/roles.go`

PR #134, opened 2020-12-18
Round timestamps to a second accuracy

tjerman:
Different parts of the system (FE, API, store) may use different
standards so operations like IsStale check may fail.


darh (approved):
lgtm.

PR #136, opened 2021-01-26
Fix sys setting filtering lapsus

tjerman:
Settings with an owner were unable to be fetched in a "fetch all"
request.

Should also get merged to 2021.3.x


tjerman:
Intentional

PR #137, opened 2021-02-09
Base templating implementation; scion

tjerman:


store/rdbms/rdbms_schema.go
```diff
@@ -238,6 +239,23 @@ func (Schema) Labels() *Table {
...
+func (Schema) Templates() *Table {
+	return TableDef("templates",
+		ID,
+		ColumnDef("rel_owner", ColumnTypeIdentifier),
+		ColumnDef("handle", ColumnTypeVarchar, ColumnTypeLength(handleLength)),
+		ColumnDef("language", ColumnTypeText),
+		ColumnDef("type", ColumnTypeText),
+		ColumnDef("partial", ColumnTypeBoolean),
+		ColumnDef("meta", ColumnTypeJson),
+		ColumnDef("template", ColumnTypeText),
```
darh, line 251: Are we safe to assume that template will always be text?
Would we be better of with wrapping it into struct/doc and save it as json?

store/rdbms/templates.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/cortezaproject/corteza-server/pkg/filter"
+	"github.com/cortezaproject/corteza-server/system/types"
+)
+
+func (s Store) convertTemplateFilter(f types.TemplateFilter) (query squirrel.SelectBuilder, err error) {
+	query = s.templatesSelectBuilder()
+
+	query = filter.StateCondition(query, "tpl.deleted_at", f.Deleted)
+
+	// @todo labels?
```
darh: do it. takes a couple of copy-pastes and it works.

store/rdbms/templates.go
```diff
@@ -0,0 +1,30 @@
...
+
+	if f.TemplateID > 0 {
```
darh: Might be smart if just switch to slice of uint64 for (primary) ids.

store/templates.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+      It also returns deleted templates.
+  - fields: [ Handle ]
+    filter: { DeletedAt: nil }
+    uniqueConstraintCheck: true
+    description: |-
+        searches for template by the handle
+
+        It returns only valid templates (not deleted)
+  - fields: [ Language ]
```
darh: this returns only one resource (hence "lookup"). Is this really what you need?

darh (changes requested):

store/rdbms/rdbms_schema.go
tjerman: I... don't know...

We will be implementing some sort of a version control system for our templates so this part will need to be reworked. I didn't find much sense in complicating this part (yet) as only HTML and plaintext templates will be supported for now (plain text stuff).

I would leave it as is for now, and we can overengineer the structure a bit later when we need to.

store/templates.yaml
tjerman: Whoops

PR #141, opened 2021-03-01
ActivityStreams 2.0 and internal corteza formatter for federation, tests

petergrlica:
This PR adds support for activity streams on data and structure sync for federation.

Added:
 - activitystreams format for data and structure sync
 - internal corteza format refactored
 - moved structure and data endpoints
 - tests

### DIFF
--- a/federation/rest.yaml
+++ b/federation/rest.yaml
@@ -329,10 +329,38 @@ endpoints:
     path: "/nodes/{nodeID}/modules/exposed"
     authentication: []
     apis:
-      - name: readExposedAll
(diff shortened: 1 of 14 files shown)


federation/rest.yaml
```diff
@@ -353,6 +353,34 @@ endpoints:
             - type: string
               name: sort
               title: Sort items
+      - name: readExposedSocial
+        method: GET
+        title: List all exposed modules changes in activity streams format
+        path: "/social"
```
darh: can we call this what it is? activity-stream?

PR #146, opened 2021-03-11
Auth tests

petergrlica:
The handlers that were tested:
 - change-password
 - login
 - logout
 - mfa
 - oauth2
 - password-reset
 - profile
 - security
 - signup

Also, added a memory session package and some fixes to the auth package.

What needs to be done:
 - integration tests
 - authorized clients test
 - sessions test


auth/handlers/handle_login_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+		req = &http.Request{}
+
+		authService  authService
+		authHandlers *AuthHandlers
+		authReq      *request.AuthReq
+
+		authSettings = &settings.Settings{}
+
+		rq = require.New(t)
```
darh: should probably be moved inside sub-test fn. I had problems when I used one initialized on root.

darh (approved):

PR #150, opened 2021-03-16
Template rbac rules, auth moved to own file

petergrlica:
What:
 - added template rules to rbac
 - moved auth rules to own file
 - refactored the directory yaml loader to separate funcs


petergrlica:
Invalid, closing.

PR #155, opened 2021-03-20
Fixed go vet errors

vicpatel:
- Removed not used call of sval.String() from migrateEmailTemplates
- Added pointer as second argument of Unmarshal call in MarshalYAML
- Removed duplicate code for random bytes generation by using param value of makeNew in testAuthSessions  

Please take a look.


darh:
Excellent! TY @ervicky 

@tjerman this is mostly your work, so fell free to merge

PR #157, opened 2021-03-24
Rbac rules auth client provisioning

petergrlica:
Added partial import func support to different directories.
Tested with partial and full provisioning (templates and auth clients).

### DIFF
--- a/pkg/provision/partial.go
+++ b/pkg/provision/partial.go
@@ -1,0 +1,49 @@
+package provision
+
+import (
+	"context"
+
+	"github.com/cortezaproject/corteza-server/pkg/filter"
+	"github.com/cortezaproject/corteza-server/pkg/rbac"
+	"github.com/cortezaproject/corteza-server/store"
+	"github.com/cortezaproject/corteza-server/system/types"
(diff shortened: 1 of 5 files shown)


pkg/provision/partial.go
```diff
@@ -0,0 +1,64 @@
...
+type (
+	uConfigFn func(context.Context, store.Storer, *zap.Logger) bool
+	uConfig   struct {
+		dir string
+		fn  uConfigFn
+	}
+)
+
+// provisionPartialAuthClients checks for a specific set of auth client rbac rules
+func provisionPartialAuthClients(ctx context.Context, s store.Storer, log *zap.Logger) bool {
```
darh, line 22: I don't understand the logic here?
All you need to check if there is at least one rule for `system:auth-client:*`.
If there is, assume provision was done and don't run it again (for that part)

darh (changes requested):
Please clarifiy `provisionPartialAuthClients`. The rest is good.

pkg/provision/partial.go
petergrlica: True, fixed.

PR #160, opened 2021-04-02
WF type tweaks

tjerman:
# 1 Reference values

When assigning a missing value to a reference record value (record, user, file, ...), the sanitiser marked it as bad and failed.
This is a bit awkward to work with.

My solution makes the sanitiser ignore `"0"`'s (missing refs).
An alternative solution would be to filter these out within workflow when we cast to a `ComposeRecord`.
IMO either would do, your choice.

# 2 Zero values

Passing missing values and casting those causes wf to fail for some cases due to `nil`.
The modification to the code template prevents that by defining a fall back to the zero value of the type.

### DIFF
--- a/automation/automation/expr_types.go
+++ b/automation/automation/expr_types.go
@@ -2,6 +2,7 @@ package automation
 
 import (
 	"fmt"
+
 	"github.com/cortezaproject/corteza-server/pkg/expr"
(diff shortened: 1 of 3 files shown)


tjerman:
Uhh, the second part breaks tests... back to the drawing board

PR #163, opened 2021-04-07
Implemented store infrastructure for cockroachDb

vicpatel:
- Implemented store methods for cockroachDB
- Updated DDL generator for cockroachDB to create/update tables and indexes
- Workaround for a LIKE operator in cockroachDB
- Added sqlFunctionHandler and fieldToColumnTypeCaster handle for cockroachDB

Please take a look.

### DIFF
--- a/store/cockroach/cockroach.go
+++ b/store/cockroach/cockroach.go
@@ -1,0 +1,116 @@
+package cockroach
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"github.com/Masterminds/squirrel"
+	"github.com/cortezaproject/corteza-server/pkg/ql"
+	"github.com/cortezaproject/corteza-server/store"
+	"github.com/cortezaproject/corteza-server/store/rdbms"
+	"github.com/cortezaproject/corteza-server/store/rdbms/instrumentation"
+	"github.com/lib/pq"
+	"github.com/ngrok/sqlmw"
+	"go.uber.org/zap"
+	"net/url"
+	"strings"
+)
+
+type (
+	Store struct {
+		*rdbms.Store
+	}
+)
+
+const (
+	storeTypeName      = "cockroachdb"
+	storeTypeWithDebug = "cockroachdb+debug"
+)
+
+// For now, CockroachDB implementation is disabled due to some limitation from cockroachDB!
+//
+//	Here is why could not work on this further:
+//	* Prefix/Expression Indexes.
+//		I.e. we are not able to use "lower" expression while creating/updating indexes
+//	* Not able to use Prefix/Expression into upsert.
+//		I.e. also not able to use "lower"(for most case) expression in upsert
+//
+// @note we could enable this once cockroachDB overcome above limitations :)
+func init() {
+	store.Register(Connect, storeTypeName, storeTypeWithDebug)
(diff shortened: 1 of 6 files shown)


store/cockroach/sql_typecasters.go
```diff
@@ -0,0 +1,16 @@
...
+
+// fieldToColumnTypeCaster handles special ComposeModule field query representations
+// @todo Not as elegant as it should be but it'll do the trick until the #2 store iteration
+//
+// Return parameters:
+//   * full cast: query column + datatype cast
+//   * field cast tpl: fmt template to get query column
+//   * type cast tpl: fmt template to cast the compared to value
+func fieldToColumnTypeCaster(field rdbms.ModuleFieldTypeDetector, ident string) (string, string, string, error) {
+	return "", "", "", nil
```
darh: why is this empty? does none of the current implementations (my/pg/lite) fit?

store/cockroach/sql_functions.go
```diff
@@ -0,0 +1,10 @@
+package cockroach
+
+import (
+	"github.com/cortezaproject/corteza-server/pkg/ql"
+)
+
+func sqlFunctionHandler(f ql.Function) (ql.ASTNode, error) {
+	// @todo ^^
```
darh: why is this empty? does none of the current implementations (my/pg/lite) fit?

darh (commented):
Great! Getting closer to working version.

store/cockroach/sql_functions.go
vicpatel: Added it, yes it does!

store/cockroach/sql_typecasters.go
vicpatel: Added it, It does!

store/cockroach/cockroach.go
```diff
@@ -0,0 +1,113 @@
...
+//
+//	Here is why could not work on this further:
+//	* Prefix/Expression Indexes.
+//		I.e. we are not able to use "lower" expression while creating/updating indexes
+//	* Not able to use Prefix/Expression into upsert.
+//		I.e. also not able to use "lower"(for most case) expression in upsert
+//
+// @note we could enable this once cockroachDB overcome above limitations :)
+func init() {
+	//store.Register(Connect, storeTypeName, storeTypeWithDebug)
```
darh: How come this is commented out?
Can we leave it enabled? Since the cockroachdb package (this) itself is not imported it does not matter what we have in the init fn()...

store/cockroach/cockroach.go
```diff
@@ -0,0 +1,113 @@
...
+
+type (
+	Store struct {
+		*rdbms.Store
+	}
+)
+
+const (
+	storeTypeName = "cockroachdb"
+	// storeTypeWithDebug = "cockroachdb+debug"
```
darh: same as the comment below for the init() fn.

darh (changes requested):

store/cockroach/cockroach.go
vicpatel: Yes, I Will add it back!

store/cockroach/cockroach.go
vicpatel: Sure

PR #172, opened 2021-04-20
Implements websocket along with in memory websocket sessions.

vicpatel:
- Implements route for websocket
- Adds default config and encoder for websocket messaging
- Sends pending prompts via websocket

Please take a look.


pkg/auth/jwt.go
```diff
@@ -140,3 +140,7 @@ func claimsToIdentity(c jwt.MapClaims) (i *Identity) {
 
 	return
 }
+
+func ClaimsToIdentity(c jwt.MapClaims) (i *Identity) {
```
darh: Just export `claimsToIdentity` fn, no need to have this intermediate fn..

pkg/options/websocket.yaml
```diff
@@ -0,0 +1,19 @@
+imports:
+  - time
+
+docs:
+  title: Websocket server
```
darh, line 5: Can we be a bit more verbose? Besides title, you can add an intro that's used for the documentation gen.

app/servers.go
```diff
@@ -63,6 +63,8 @@ func (app *CortezaApp) mountHttpRoutes(r chi.Router) {
 			r.Route("/automation", automationRest.MountRoutes)
 			r.Route("/compose", composeRest.MountRoutes)
 
+			app.WsServer.MountRoutes(r)
```
darh: Can we move the "/websocket" route root here so that we follow the same pattern of route definitions?

pkg/payload/incoming/jwt.go
```diff
@@ -0,0 +1,25 @@
+package incoming
+
+import (
+	"github.com/dgrijalva/jwt-go"
+	"github.com/pkg/errors"
+)
+
+type (
+	Token struct {
```
darh: Can we be more specific here about what kind of token are we talking about? So `Auth` would probably be a better choice for struct name. And for the file name as well.

pkg/payload/incoming.go
```diff
@@ -0,0 +1,12 @@
+package payload
```
darh: I would prefer if we keep this `pkg/payload/...` removed - it's an obsolete pattern that was part of messaging components and and old mindset. If you really really need it, move it under websocket package. But really.. let's discuss this on a call and find a better way to solve this.

pkg/payload/outgoing/payload.go
```diff
@@ -0,0 +1,16 @@
+package outgoing
+
+type (
+	Payload struct {
```
darh: Just repeating this for the sake of not missing it :)

I would prefer if we keep this pkg/payload/... removed - it's an obsolete pattern that was part of messaging components and and old mindset. If you really really need it, move it under websocket package. But really.. let's discuss this on a call and find a better way to solve this.

websocket/router.go
```diff
@@ -0,0 +1,26 @@
+package websocket
+
+import (
+	"github.com/go-chi/chi"
+	"net/http"
+)
+
+func middlewareAllowedAccess(next http.Handler) http.Handler {
```
darh: no need for that at all as we agreed, everyone can access ws endpoint

websocket/router.go
```diff
@@ -0,0 +1,26 @@
...
+		//}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (ws Websocket) MountRoutes(r chi.Router) {
+	r.Group(func(r chi.Router) {
+		r.Route("/websocket", func(r chi.Router) {
+			//r.Use(middlewareAllowedAccess)
```
darh: Just add a comment or two about why there are not any auth-related middlewares here and where/how access control is done.

websocket/session_incoming_channel.go
```diff
@@ -0,0 +1,31 @@
...
+		return err
+	}
+
+	// Get identity using JWT claims
+	identity := auth.ClaimsToIdentity(claims)
+
+	// Update the existing ws sessions if exists or create new one
+	if store.CountConnections(identity.Identity()) > 0 {
+		store.Walk(func(session *Session) {
+			session.user = identity
```
darh: This looks wrong? Are we setting a new identity on ALL sessions regardless? Or am I misinterpreting this?

websocket/session_store.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/cortezaproject/corteza-server/pkg/id"
+	"sync"
+)
+
+var store *Store
+
+type (
+	Store struct {
```
darh: there's no need for this to be published, right? Plus... might be best if we find a better name for this, since this was originally implemented, we introduced "Store" for something else.

Even `sessions` (so, plural) is good enough.

websocket/session_store.go
```diff
@@ -0,0 +1,78 @@
...
+
+		Sessions map[uint64]*Session
+	}
+)
+
+func init() {
+	store = NewStore()
+}
+
+func NewStore() *Store {
```
darh: when you rename the struct, change this from `NewStore` to `Store`

pkg/options/websocket.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+props:
```
darh, line 8: Maybe we should add `logEnabled` (same as auth has it) to explicitly control logging in ws?

websocket/websocket.go
```diff
@@ -0,0 +1,59 @@
...
+		ReadBufferSize:  1024,
+		WriteBufferSize: 1024,
+
+		// Allow connections from any Origin
+		CheckOrigin: func(r *http.Request) bool { return true },
+	}
+)
+
+type (
+	Websocket struct {
```
darh: unexport/lowercase please.

websocket/websocket.go
```diff
@@ -0,0 +1,59 @@
...
+		config *Config
+	}
+)
+
+func New(config *Config) *Websocket {
```
darh: Please no `New`. Should be `Websocket`.

websocket/websocket.go
```diff
@@ -0,0 +1,59 @@
...
+
+		// Allow connections from any Origin
+		CheckOrigin: func(r *http.Request) bool { return true },
+	}
+)
+
+type (
+	Websocket struct {
+		config *Config
+	}
```
darh: Add logger

websocket/websocket.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return ws
+}
+
+func (ws Websocket) Open(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if _, ok := err.(websocket.HandshakeError); ok {
+		fmt.Println("ws: need a websocket handshake")
```
darh: use logger.

websocket/websocket.go
```diff
@@ -0,0 +1,59 @@
...
+	} else if err != nil {
+		fmt.Println("ws: failed to upgrade connection")
+		api.Send(w, r, errors.Wrap(err, "ws: failed to upgrade connection"))
+		return
+	}
+
+	session := (&Session{}).New(ctx, ws.config, conn)
+
+	if err := session.Handle(); err != nil {
+		logger.Default().
```
darh: bring in logger via "constructor" fn and user that.

darh (changes requested):
Needs a few tweaks & improvements. But on a good path!

pkg/auth/jwt.go
vicpatel: Sure

pkg/options/websocket.yaml
vicpatel: Yes

automation/service/session.go
```diff
@@ -272,6 +274,29 @@ func (svc *session) Watch(ctx context.Context) {
 		//svc.suspendAll(ctx)
 	}()
 
+	// Prompt routine
+	go func() {
```
darh: Merge with the goroutine above.

automation/service/session.go
```diff
@@ -272,6 +274,29 @@ func (svc *session) Watch(ctx context.Context) {
...
+		defer sentry.Recover()
+		defer promptTicker.Stop()
+		defer svc.log.Info("stopped")
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-promptTicker.C:
+				activeSessions := websocket.GetActiveSessions()
```
darh: this is still too coupled with WS... but I guess you plan to refactor this further.

automation/service/session.go
```diff
@@ -272,6 +274,29 @@ func (svc *session) Watch(ctx context.Context) {
...
+				for _, s := range activeSessions {
+					var ctxr = context.Background()
+					pp := svc.PendingPrompts(auth.SetIdentityToContext(ctxr, s.User()))
+					if len(pp) > 0 {
+						_ = s.Send(websocket.Message(websocket.StatusOK, websocket.WorkflowApplication, pp), s.User().Identity())
```
darh: the more I think about this application filter, the less sense it makes... 
either way, this info should come from the prompt itself and you should not hard coded it like this.

websocket/router.go
```diff
@@ -2,25 +2,15 @@ package websocket
...
-}
-
-func (ws Websocket) MountRoutes(r chi.Router) {
+// MountRoutes initialize route for websocket
+// No middleware used, since anyone can open connection and
+// send first message with valid JWT token,
+// If it's valid then we keep the connection open or close it
+func (ws *websocket) MountRoutes(r chi.Router) {
+	// Initialize handlers & controllers.
 	r.Group(func(r chi.Router) {
```
darh: do we still need group?

darh (changes requested):

websocket/router.go
vicpatel: No

websocket/encoder.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	StatusOK    = "ok"
+	StatusError = "error"
+)
+
+type (
+	message struct {
+		Status      string      `json:"status"`
+		Application uint64      `json:"application"`
```
darh: uint64? tell me how do you see this used

automation/service/session.go
```diff
@@ -264,6 +265,18 @@ func (svc *session) Watch(ctx context.Context) {
 				// @todo cleanup pool when sessions are complete
 
 			case <-gcTicker.C:
+				// Sends pending prompt via websocket
+				ws := websocket.Session(ctx, nil, nil)
```
darh: this does not look right...

websocket/session.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func (s *session) Walk(callback func(*session)) {
+	for _, sess := range sessions {
+		callback(sess)
+	}
+}
+
+// GetActiveUserIDs return userIDs from active ws sessions
+func (s *session) GetActiveUserIDs() (ids []uint64) {
```
darh: Do we care about duplictes?

websocket/session.go
```diff
@@ -0,0 +1,325 @@
...
+	return nil
+}
+
+func (s *session) Save() *session {
+	if s.id == 0 {
+		s.id = id.Next()
+	}
+
+	if s.user != nil {
+		if _, ok := sessions[s.user.Identity()]; !ok {
```
darh: so... one session per user? will this work?

darh (changes requested):

websocket/session.go
vicpatel: No, Fixed it!

websocket/sessionImpl.go
```diff
@@ -0,0 +1,20 @@
+package websocket
```
darh: why did you name this file `sessionImpl.go`? What does impl mean? Can we at least use snake-case please.

websocket/encoder.go
```diff
@@ -0,0 +1,32 @@
...
+	"encoding/json"
+)
+
+const (
+	StatusOK    = "ok"
+	StatusError = "error"
+)
+
+type (
+	message struct {
```
darh: is this whole thing even used anymore?

darh (changes requested):
2 more things and good to go.

websocket/sessionImpl.go
vicpatel: It supposes to be the session implementation file for incoming messages!

websocket/encoder.go
vicpatel: Yes, only for sending messages to users. but not renamed it to response and moving it to types.

websocket/websocket.go
```diff
@@ -0,0 +1,88 @@
...
+// Send delivers message to user to ones we want to
+// if len(userIDs) == 0 -- it delivers to everyone
+func (ws *websocket) Send(kind string, payload interface{}, userIDs ...uint64) error {
+	pb, err := Response(kind, payload).Marshal()
+	if err != nil {
+		return err
+	}
+
+	sendsToAll := len(userIDs) == 0
+	userIDMap := make(map[uint64]uint64)
```
darh: if you do map[uint64]bool this makes the rest of the code even lighter.

websocket/types.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"encoding/json"
+	"github.com/dgrijalva/jwt-go"
+	"github.com/pkg/errors"
+)
+
+type (
+	// OAuth is JWT token provided by client as first message,
+	// and will be passed whenever it changes
+	OAuth struct {
```
darh: Auth made more sense, sorry:)

darh (changes requested):

PR #173, opened 2021-04-20
2021.6.x feature messagebus

petergrlica:
Messagebus featuring:
 - store consumer (using default settings)
 - eventbus consumer
 - producers support (each producer has a separate funcionality, currently only eventbus works via workflows)
 - tests


pkg/messagebus/eventbus.go
```diff
@@ -0,0 +1,45 @@
...
+
+	return h
+}
+
+func (cq *EventbusConsumer) Write(ctx context.Context, p []byte) error {
+	cq.dispatcher.Dispatch(ctx, event.QueueOnMessage(makeEvent(cq.queue, p)))
+	return nil
+}
+
+func (cq *EventbusConsumer) SetStorer(s QueueStorer) {}
```
darh: Even if you are setting (and getting, in another case) a `....Storer` (an interface, so -er suffix), you are actually setting/getting a store (a struct, so no -er suffix)

pkg/messagebus/client.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/cortezaproject/corteza-server/pkg/filter"
+	"github.com/cortezaproject/corteza-server/pkg/id"
+)
+
+type (
+	Client interface {
```
darh, line 12: Defining an interface (especially in such case) with unpublished functions makes little sense - how are you ever going to define a client outside of this package?

pkg/messagebus/client.go
```diff
@@ -0,0 +1,76 @@
...
+		add(ctx context.Context, q string, p []byte) (err error)
+		get(ctx context.Context, q string) (list QueueMessageSet, err error)
+		process(ctx context.Context, m QueueMessage) (err error)
+	}
+
+	SqlClient interface {
```
darh: Might be a misunderstanding when we had a call... Was this RDBMS client before? We discussed about renaming it to StoreClient, right?

.gitattributes
```diff
@@ -0,0 +1,3 @@
+**/*gen.go linguist-generated=true
+**/*gen_test.go linguist-generated=true
+**/*gen_test.go.tpl linguist-generated=true
```
darh: Like we discusssed; `.tpl` files should not be excluded.

btw - when you fix this, feel free to cherry-pick this change directly to `2021.6.x` branch if possible

auth/request/manager.go
```diff
@@ -48,12 +49,12 @@ func (m *SessionManager) Search(ctx context.Context, userID uint64) (set []*type
 	return
 }
 
-// Returns all users sessions
+// DeleteByUserID removes session by user ID
```
darh, line 52: Nice catch; next time (really, next time) feel free to make this changes directly on the version branch

automation/automation/queue_handler.go
```diff
@@ -0,0 +1,39 @@
...
+	return h
+}
+
+func (h queueHandler) write(ctx context.Context, args *queueWriteArgs) (err error) {
+	if !args.hasQueue {
+		return fmt.Errorf("could not send message to queue, queue missing")
+	}
+
+	if !args.hasPayload {
+		return fmt.Errorf("could not send message to queue, payload missing")
```
darh, line 31: Is payload really necessary? Could producer signal consumer with an empty/without payload?

pkg/messagebus/consumer.go
```diff
@@ -0,0 +1,39 @@
...
+const (
+	ConsumerCorteza  ConsumerType = "corteza"
+	ConsumerNoop     ConsumerType = "noop"
+	ConsumerRedis    ConsumerType = "redis"
+	ConsumerStore    ConsumerType = "store"
+	ConsumerEventbus ConsumerType = "eventbus"
+)
+
+type (
+	ConsumerType string
```
darh, line 16: What is the benefit of having consumer types defined? Isn't implementation/interface enough?

store/messagebus_queuemessage.yaml
```diff
@@ -0,0 +1,27 @@
+import:
```
darh: Can we split queuemessage (referring to name of the file) so that it is properly converted to CamelCase?

store/messagebus_queuesettings.yaml
```diff
@@ -0,0 +1,37 @@
+import:
```
darh: Can we split queuesettings (referring to name of the file) so that it is properly converted to CamelCase?

store/messagebus_queuesettings.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - github.com/cortezaproject/corteza-server/pkg/messagebus
+
+types:
+  package: messagebus
+  type: messagebus.QueueSettings
+  filterType: messagebus.QueueSettingsFilter
+
+fields:
```
darh: remove sortable on `ID` and add it to `Queue`, and all three `....At` fields

store/messagebus_queuemessage.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - { field: Payload }
+  - { field: Processed, type: "time.Time" }
+  - { field: Created,   type: "time.Time" }
+
+rdbms:
+  alias: mqm
+  table: queue_messages
+  customFilterConverter: true
+
+search:
```
darh: you probably don't need sorting here?

pkg/messagebus/producer.go
```diff
@@ -0,0 +1,27 @@
...
+		Subscriber
+		Poller
+		Storer
+	}
+
+	Subscriber interface {
+		Notification(ctx context.Context) <-chan interface{}
+	}
+
+	Poller interface {
```
darh, line 20: Why not a Ticker? :)

pkg/messagebus/producer.go
```diff
@@ -0,0 +1,27 @@
...
+
+type (
+	Producer interface {
+		Reader
+		Subscriber
+		Poller
+		Storer
+	}
+
+	Subscriber interface {
```
darh, line 16: Why not a Notifier? Or should fn. be named Subscribe? :)

pkg/messagebus/queue.go
```diff
@@ -0,0 +1,28 @@
...
+
+		// settings are used in messagebus for handling
+		// throttling, polling settings
+		settings QueueSettings
+	}
+
+	QueueSet map[string]*Queue
+)
+
+func (qs *QueueSet) toSlice() []*Queue {
```
darh: overkill?

system/service/queue.go
```diff
@@ -0,0 +1,252 @@
...
+		CanReadMessagebusQueue(ctx context.Context, c *messagebus.QueueSettings) bool
+		CanUpdateMessagebusQueue(ctx context.Context, c *messagebus.QueueSettings) bool
+		CanDeleteMessagebusQueue(ctx context.Context, c *messagebus.QueueSettings) bool
+
+		CanReadFromMessagebusQueue(ctx context.Context, c *messagebus.QueueSettings) bool
+		CanWriteToMessagebusQueue(ctx context.Context, c *messagebus.QueueSettings) bool
+	}
+)
+
+func Queue() QueueService {
```
darh: Please model this service more closely to how Application service is designed.
So, constructor (Queue fn.) should return a ptr to a struct and not an interface.

`...Service` interface inside service pkg (like the `QueueService` above) is an old pattern we need to get rid off across all codebase.

darh (changes requested):
Excellent. A few minor things to improve + some questions.

.gitattributes
petergrlica: :+1:

automation/automation/queue_handler.go
petergrlica: Must be present, can be empty was my idea here...

pkg/messagebus/client.go
petergrlica: Fixed

pkg/messagebus/producer.go
petergrlica: refactored

pkg/messagebus/producer.go
petergrlica: refactored

darh (approved):
lgtm. 🚀

PR #180, opened 2021-05-07
Workflows; type signatures; function reference;

vicpatel:
- Updates path for generated docs
-  Adds adoc codegen for expression types
-  Adds adoc codegen for automatic functions

Please take a look.

### DIFF
--- a/pkg/codegen/afunc.go
+++ b/pkg/codegen/afunc.go
@@ -216,3 +216,17 @@ func genAutomationFunctions(tpl *template.Template, dd ...*aFuncDefs) (err error
 
 	return nil
 }
+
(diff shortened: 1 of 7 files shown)


tjerman (changes requested):
1. Filter out any **non-struct type**, so only the ones that have the `struct:` property should be included (a simple if will do).
2. Use the `exprType` property as the data type instead of the `goType`
3. Include the function reference

tjerman:
I've tweaked the template files; lgtm.

@darh do you have any comments? If not; rebase, (probably) cleanup commits, and merge.

darh:
Nope. just -- rebase, cleanup, merge.

PR #181, opened 2021-05-10
Extends SCIM implementation to support role membership management

vicpatel:
- Updates request payload for patching group
- Updates tests

Please take a look.


darh (approved):
LGTM. Merge + patch version branches as instructed.

PR #182, opened 2021-05-11
Add a call stack to prevent infinite WF invocations

tjerman:


pkg/options/workflow.yaml
```diff
@@ -11,3 +11,8 @@ props:
     type: bool
     default: false
     description: Enables verbose logging for workflow execution
+
+  - name: callStackSize
+    type: int
+    default: 16
+    description: Defines the maximum call stack size
```
darh: can we be a bit more explicit here? that this means call stack between workflows, not steps in the workflow.

pkg/wfexec/session.go
```diff
@@ -552,8 +559,9 @@ func (s *Session) exec(ctx context.Context, st *State) (err error) {
 			// so we need to pull it from state owner and add it to new context
 			// that is set to step exec function
 			ctxWithIdentity := auth.SetIdentityToContext(ctx, st.owner)
+			ctxWithStack := SetContextCallStack(ctxWithIdentity, s.callStack)
```
darh: `stepCtx` or `subCtx`... no need that we create a new var for each variation

tjerman:
Merged

PR #183, opened 2021-05-12
Adds expr functions for KV, KVV and Vars manipulation

vicpatel:
Adds expr functions for KV, KVV, Vars manipulation

- Adds/Refactors methods for Set, Merge, Filter, Delete
- Renames expr/Vars.Merge to MustMerge, updates its usage
- Appends kvFunctions to parser in pkg/expr
- Update/Fixes tests and example testable

Please take a look.

### DIFF
--- a/pkg/expr/expr_types.go
+++ b/pkg/expr/expr_types.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/PaesslerAG/gval"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -23,14 +24,69 @@ type (
 		Type
 		ResolveTypes(func(string) Type) error
 	}
+
+	merger interface {
+		Merge(...Iterator) (TypedValue, error)
+	}
+
+	filterer interface {
+		Filter(...string) (TypedValue, error)
+	}
+
+	deleter interface {
+		Delete(...string) (TypedValue, error)
+	}
 )
 
+func KvFunctions() []gval.Language {
+	return []gval.Language{
+		gval.Function("set", set),
+		gval.Function("merge", merge),
+		gval.Function("filter", filter),
+		gval.Function("omit", omit),
+	}
+}
+
+func EmptyKV() *KV {
+	return &KV{value: make(map[string]string)}
+}
+
+func EmptyKVV() *KVV {
+	return &KVV{value: make(map[string][]string)}
+}
+
 func EmptyVars() *Vars {
-	return &Vars{value: map[string]TypedValue{}}
+	return &Vars{value: make(map[string]TypedValue)}
 }
 
 func ResolveTypes(rt resolvableType, resolver func(typ string) Type) error {
 	return rt.ResolveTypes(resolver)
+}
+
+func set(m merger, key string, val TypedValue) (out TypedValue, err error) {
+	out, err = m.Merge()
+	if err != nil {
+		return
+	}
+
+	err = Assign(out, key, val)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+func merge(m merger, nn ...Iterator) (TypedValue, error) {
+	return m.Merge(nn...)
+}
+
+func filter(f filterer, ss ...string) (TypedValue, error) {
+	return f.Filter(ss...)
+}
+
+func omit(d deleter, ss ...string) (TypedValue, error) {
+	return d.Delete(ss...)
 }
 
 // Typify detects input type and wraps it with expression type
@@ -301,7 +357,7 @@ func CastToString(val interface{}) (out string, err error) {
 	}
 }
 
-func CastStringSlice(val interface{}) (out []string, err error) {
+func CastToStringSlice(val interface{}) (out []string, err error) {
 	return cast.ToStringSliceE(UntypedValue(val))
 }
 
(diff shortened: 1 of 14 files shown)


pkg/expr/expr_types.go
```diff
@@ -23,6 +23,26 @@ type (
 		Type
 		ResolveTypes(func(string) Type) error
 	}
+
+	iterator interface {
+		Each(func(string, interface{}) error) error
+	}
+
+	setter interface {
```
darh: could an existing `FieldAssigner` interface be used instead of this, new one?

pkg/expr/expr_types.go
```diff
@@ -23,6 +23,26 @@ type (
 		Type
 		ResolveTypes(func(string) Type) error
 	}
+
+	iterator interface {
```
darh: could an existing `Iterator` be used instead of this, new one?

darh:
btw, regarding interface `filter`

https://golang.org/doc/effective_go#interface-names

and you would not be the first one to go with `filterer`, no matter how quirky it sounds
https://pkg.go.dev/github.com/msample/logfu

:)

pkg/expr/expr_types.go
vicpatel: Yes

pkg/expr/values.go
```diff
@@ -30,6 +30,7 @@ type (
 
 	FieldAssigner interface {
 		AssignFieldValue(string, TypedValue) error
+		Set(string, interface{}) error
```
darh: Why? :) AssignFieldValue does exact same thing already.

pkg/expr/expr_types.go
```diff
@@ -404,6 +432,85 @@ func CastToKV(val interface{}) (out map[string]string, err error) {
 	}
 }
 
+func (t *KV) Each(fn func(k string, v TypedValue) error) (err error) {
+	if t == nil || t.value == nil {
+		return
+	}
+
+	for k, v := range t.value {
+		if err = fn(k, Must(NewString(v))); err != nil {
```
darh: Must will panic. You use it when it's ok to panic (and it's not in most cases).
You return error from Each anyway, so better to try and do NewString, and return if it yields an error.

pkg/expr/expr_types.go
```diff
@@ -404,6 +432,85 @@ func CastToKV(val interface{}) (out map[string]string, err error) {
...
+	if t.value == nil {
+		t.value = make(map[string]string)
+	}
+
+	nn = append([]Iterator{t}, nn...)
+
+	for _, i := range nn {
+		if err = i.Each(func(k string, v TypedValue) error {
+			t.value[k] = cast.ToString(v.Get())
+			return err
```
darh: What error are you returning here exactly?

pkg/expr/expr_types.go
```diff
@@ -404,6 +432,85 @@ func CastToKV(val interface{}) (out map[string]string, err error) {
...
+func (t *KV) Merge(nn ...Iterator) (err error) {
+	if t.value == nil {
+		t.value = make(map[string]string)
+	}
+
+	nn = append([]Iterator{t}, nn...)
+
+	for _, i := range nn {
+		if err = i.Each(func(k string, v TypedValue) error {
+			t.value[k] = cast.ToString(v.Get())
```
darh: you can use `cast.ToStringE` to get the fn variant that does not suppress your errors

pkg/expr/expr_types.go
```diff
@@ -404,6 +432,85 @@ func CastToKV(val interface{}) (out map[string]string, err error) {
...
+// Merge combines the given KVs into KV
+func (t *KV) Merge(nn ...Iterator) (err error) {
+	if t.value == nil {
+		t.value = make(map[string]string)
+	}
+
+	nn = append([]Iterator{t}, nn...)
+
+	for _, i := range nn {
+		if err = i.Each(func(k string, v TypedValue) error {
```
darh: please keep _lambdas_ out of ifs

pkg/expr/vars.go
```diff
@@ -409,3 +410,36 @@ func CastToVars(val interface{}) (out map[string]TypedValue, err error) {
...
+// Filter take keys returns KV with only those key value pair
+func (t *Vars) Filter(keys ...string) (err error) {
+	if t.value == nil {
+		return
+	}
+
+	// Delete key from t.value if it is not from keys
+	for k := range t.value {
+		if !contains(keys, k) {
+			delete(t.value, k)
```
darh: no need to pre-check for key for delete fn.

pkg/expr/expr_types.yaml
```diff
@@ -37,6 +37,9 @@ types:
     as: 'string'
     default: '""'
 
+  StringSlice:
```
darh: do we need this?

darh (changes requested):

pkg/expr/expr_types.yaml
vicpatel: No

pkg/expr/expr_types.go
vicpatel: Fixed it

darh (changes requested):
these new functions still need to be aded to `expr` pkg, see existing function mappings (via `gval.Function`) in `pkg/expr/func_*.go`

pkg/expr/expr_types_test.go
```diff
@@ -8,6 +8,149 @@ import (
...
+			"kvv":   &KVV{},
+			"key":   "foo",
+			"value": Must(NewString("bar")),
+		}
+	)
+
+	eval(`set(kvv, key, value)`, p)
+
+	// output:
+	// <nil>
```
darh: Is this correct? Should test return <nil>? Or modified input?
So, is original input mutated here or does it stay unchanged after call? Would be very nice if we could prove that it's unchanged with the test?

pkg/expr/expr_types_test.go
```diff
@@ -8,6 +8,149 @@ import (
...
+			"kv":    &KV{},
+			"key":   "k1",
+			"value": Must(NewString("v11")),
+		}
+	)
+
+	eval(`set(kv, key, value)`, p)
+
+	// output:
+	// <nil>
```
darh: same as for Example_set_kvv

pkg/expr/vars_test.go
```diff
@@ -8,6 +8,69 @@ import (
...
+			"vars":  &Vars{},
+			"key":   "foo",
+			"value": &String{value: "foo"},
+		}
+	)
+
+	eval(`set(vars, key, value)`, p)
+
+	// output:
+	// <nil>
```
darh: Same as for Example_set_kvv

pkg/expr/vars_test.go
```diff
@@ -8,6 +8,69 @@ import (
...
+			"foo": &Vars{value: map[string]TypedValue{
+				"k1": &String{value: "v1"},
+			}},
+			"bar": &Vars{value: map[string]TypedValue{
+				"k2": Must(NewString("v2")),
+			}},
+		}
+	)
+
+	eval(`merge(vars, bar)`, p)
```
darh: Missing output?

pkg/expr/vars.go
```diff
@@ -58,8 +58,26 @@ func (t Vars) ResolveTypes(res func(typ string) Type) (err error) {
...
+		})
+	}
+
+	return vars, nil
+}
+
+// Deprecated: MustMerge combines the given Vars(es) into Vars
+// It will be now Merge(nn ...Iterator) (TypedValue, error)
+// Which does the same except it will return TypedValue and error(if necessary)
+func (t *Vars) MustMerge(nn ...Iterator) *Vars {
```
darh, line 79: Why is it deprecated? :)
Let's call it a shorthand function.

darh (changes requested):
Almost there...

pkg/expr/expr_types_test.go
vicpatel: Sure

PR #192, opened 2021-05-24
Prevent error handling from ending the WF exec session

tjerman:
(`2021.3.x`, `2021.6.x`)

When the worker finished with the step execution, it only checked if the error was set and not if the error was handled.
This fixes it and it worked for my test cases; please check it out if I've missed anything.


tjerman:
Rebased & merged

PR #194, opened 2021-05-26
Adds validation for corredor steps in workflow

vicpatel:
- Adds check at workflow step level for corredorExec function
- Adds `disabled` flag for corredor related functions from `workflows/functions` api response if corredor is disabled

Please take a look.


darh (approved):
lgtm.

PR #197, opened 2021-06-07
Improves record-value manipulation

vicpatel:
This will fix WF issue for accessing record's value via selectors & gval expressions

Please take a look.

### DIFF
--- a/compose/automation/expr_types.go
+++ b/compose/automation/expr_types.go
@@ -4,13 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
 	"github.com/PaesslerAG/gval"
 	"github.com/cortezaproject/corteza-server/compose/types"
 	"github.com/cortezaproject/corteza-server/pkg/expr"
+	"github.com/cortezaproject/corteza-server/pkg/slice"
 	"github.com/spf13/cast"
-	"strconv"
-	"strings"
-	"time"
 )
 
 func CastToComposeNamespace(val interface{}) (out *types.Namespace, err error) {
@@ -99,7 +101,8 @@ func (t *ComposeRecord) AssignFieldValue(kk []string, val expr.TypedValue) error
 	switch kk[0] {
 	case "values":
 		return assignToComposeRecordValues(t.value, kk[1:], val)
-		// @todo deep setting labels
+	// case "labels":
+	// @todo deep setting labels
 	default:
 		return assignToComposeRecord(t.value, kk[0], val)
 	}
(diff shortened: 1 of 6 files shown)


compose/automation/expr_types.go
```diff
@@ -171,25 +174,65 @@ func CastToComposeRecordValues(val interface{}) (out types.RecordValueSet, err e
...
+			out = out.Set(&types.RecordValue{Name: k, Value: val[k]})
 		}
 
 		return
 
 	case map[string][]string:
-		out = types.RecordValueSet{}
-		for k, vv := range val {
-			for i, v := range vv {
+
```
tjerman: remove empty line

tjerman (commented):
This seems fine to me. I'll take another closer look before we merge it tho (not before next week!)

compose/automation/expr_types.go
```diff
@@ -171,26 +174,61 @@ func CastToComposeRecordValues(val interface{}) (out types.RecordValueSet, err e
...
 				out = out.Set(&types.RecordValue{Name: k, Value: v, Place: uint(i)})
 			}
 		}
-
 		return
+	case map[string]interface{}:
+		for _, k := range slice.Keys(val) {
+			if val[k] == nil {
+				continue
+			}
```
tjerman: `nil` can be a bit tricky when working with `interface{}`.
I would prefer to see something in the lines of https://github.com/cortezaproject/corteza-server/blob/2021.6.x-fix-wf-copy-record/pkg/expr/func_generic.go#L42 with some additional tests to cover this specific execution path.
You can copy-paste that function into the `compose/automation` package.

Ref: https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/

compose/automation/expr_types.go
```diff
@@ -171,26 +174,61 @@ func CastToComposeRecordValues(val interface{}) (out types.RecordValueSet, err e
...
+				switch v := value.(type) {
+				case bool:
+					if !v {
+						// a small exception for boolean fields, don't add false values
+						continue
+					} else {
+						// when true, cast to 1
+						// legacy ...
+						value = "1"
+					}
```
tjerman: I would prefer to see these comments explaining the booleans below the `case bool:` line; all together

tjerman (changes requested):
In addition to my comments, can you please add some additional tests to cover cases with missing values?

compose/automation/expr_types_test.go
```diff
@@ -341,21 +341,25 @@ func TestCastToComposeRecordValues(t *testing.T) {
 			&types.RecordValue{Name: "strings", Value: "val"},
 			&types.RecordValue{Name: "strings", Value: "val2", Place: 1},
 			// false booleans will not be set and that is ok!
-			//&types.RecordValue{Name: "bools", Value: "false", Place: 1},
+			// &types.RecordValue{Name: "bools", Value: "false", Place: 1}, 9879201458
```
tjerman: ?

tjerman (commented):
What is that number for? Other then that, lgtm, merge

compose/automation/expr_types_test.go
vicpatel: My bad, will remove it!

PR #200, opened 2021-06-11
Adds before/after suspend to system:user resource

vicpatel:
Please take a look.

### DIFF
--- a/system/service/user.go
+++ b/system/service/user.go
@@ -594,11 +594,19 @@ func (svc user) Suspend(ctx context.Context, userID uint64) (err error) {
 			return UserErrNotAllowedToSuspend()
 		}
 
+		// Clone u to oldUser
+		oldUser := *u
 		u.SuspendedAt = now()
+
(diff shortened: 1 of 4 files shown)


tjerman (changes requested):
This is the first step; the next is to make it actually work.

Take a look on how `system/service/user.go` -> `Update(...) {...}` function works (note the `svc.eventbus.WaitFor(...)` lines)

Lastly define a workflow that has a trigger listening on that event and performs some action, like a prompt.

system/service/user.go
```diff
@@ -594,11 +594,20 @@ func (svc user) Suspend(ctx context.Context, userID uint64) (err error) {
 			return UserErrNotAllowedToSuspend()
 		}
 
+		// Clone u to oldUser
+		oldUser := *u
+		// Modify u.SuspendedAt to current time
```
tjerman: Remove this comment; it's not needed

tjerman (commented):
Address my comment, then merge. lgtm

PR #204, opened 2021-06-21
Fix field access control for default values

tjerman:
If a module field defined a default value and RBAC rules that denied some user from creating a specific field; any create/update requests would fail, because the field contained some value (the default).

This fix ignores default values for fields where the user doesn't have access.
I think this solution is completely valid; if the field with a default value needs to be populated, it should be marked as required.

Another solution would be to assign the defaults after all the AC and such is ran.
I am not so sure how well such an approach would hold.

**todo** tests


petergrlica (approved):
todo - add ac to tests later

PR #207, opened 2021-06-24
Adds check before dismissing the reminder of other user

vicpatel:
Please take a look.


system/service/reminder_actions.yaml
```diff
@@ -61,3 +61,9 @@ errors:
 
   - error: notAllowedToAssign
     message: "not allowed to assign reminders to other users"
+
+  - error: notAllowedToDismiss
+    message: "not allowed to dismiss reminders of other users"
+
+  - error: notAllowedToRead
+    message: "not allowed to check reminders of other users"
```
darh: should be `"not allowed to **read** reminders of other users"` (so, read, not check)

darh (commented):
Would be nice to have test for scenario for reading reminders of other users...

darh (approved):

PR #209, opened 2021-06-29
2021.6.x feature new ci v2

formaldehid:
Please take a look.


.gitignore
```diff
@@ -13,3 +13,5 @@ profile.coverprofile
 /gin-bin
 /.vscode
 /federation/etc/.env*
+/.act*
+/workflow
```
darh, line 17: why are we ignoring this?

darh (commented):
One tiny question in the .gitignore.
plus, branch is lagging behind `2021.6.x` a bit.

Other than that, looks ok (did not do any in-depth sanity check of the YAML files...)

.gitignore
formaldehid: I'm using https://github.com/nektos/act to test pipelines locally. This produces /workflow folder as output. I ignored /.act* type of files because you can send events in json to this simulator... $act -b -e .act-push-event.json (for instance)

darh: ok.. no problems with .act... 
but why do we ignore /workflow?

formaldehid: running act produces workflow folder where it stores output of jobs

darh: Understood. 

_on a sidenote.... our ignore files are getting bulkier, might be good to put a comment or two on some lines, especially when they use keywords (workflow) that are used in the product itself._

PR #210, opened 2021-06-29
Adds admin API for user session enhancement

vicpatel:
- log-out users from all sessions
- unauthorize all auth clients

Please take a look.

### DIFF
--- a/system/rest.yaml
+++ b/system/rest.yaml
@@ -694,6 +694,16 @@ endpoints:
         type: string
         title: Script to execute
         required: true
+  - name: sessionsRemove
+    method: DELETE
+    title: Remove all auth sessions of user
+    path: "/{userID}/sessions"
+    parameters:
+      path:
+        - type: uint64
+          name: userID
+          required: true
+          title: ID
 - title: Applications
   path: "/application"
   entrypoint: application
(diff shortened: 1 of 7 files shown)


auth/handlers/handle_sessions.go
```diff
@@ -99,14 +106,42 @@ func (h *AuthHandlers) sessionsProc(req *request.AuthReq) error {
...
+		return err
+	}
+
+	ss, err := h.getSessionsByUser(req.Context(), userID, userSessionParams{})
+	if err != nil {
+		return err
+	}
+
+	for _, s := range ss {
+		if err = h.SessionManager.DeleteByID(req.Context(), s.id); err != nil {
```
darh: how about if we just use DeleteByUserID() instead of loading+looping+deleting-each-one?

auth/handlers/handle_authorized-clients.go
```diff
@@ -69,19 +70,39 @@ func (h *AuthHandlers) clientsProc(req *request.AuthReq) error {
...
+	}
+
+	ss, err := h.getAuthorizedClients(req.Context(), userID)
+	if err != nil {
+		return err
+	}
+
+	for _, s := range ss {
+		clientID := s.ID
+		if err = h.TokenService.DeleteByID(req.Context(), clientID); err != nil {
```
darh: how about if we just use DeleteByUserID() instead of loading+looping+deleting-each-one?

darh (changes requested):

vicpatel:
> well... `DELETE .../logout` does not make much sense. How could this be improved?

Yes, DELETE `../authentication`

darh:
How about `.../sessions`? This way we are future-proof if/when we split this to 2 endpoints (one for sessions, one for tokens) as we discussed.

darh:
How about `DELETE .../sessions`? This way we'll be future-proof when we decide to break this into two endpoints (one for sessions, other for oa2 tokens)

PR #216, opened 2021-07-09
Fake data

vicpatel:
Adds data gen for records and user.

It has cli integration with we can create users, record as many we want.

Please take a look.


seeder/dataGen.go
```diff
@@ -0,0 +1,250 @@
...
+	var users []*sTypes.User
+	var labels []*lTypes.Label
+
+	for i := 0; i < opt.getLimit(); i++ {
+		var user sTypes.User
+		user.ID = id.Next()
+		user.Email, _ = gen.faker.fakeValueByName("Email")
+		user.Name, _ = gen.faker.fakeValueByName("Name")
+		user.Handle = gen.faker.fakeUserHandle(user.Name)
+		user.Kind = sTypes.BotUser
```
darh: why bot user? switch this to normal.

seeder/faker.go
```diff
@@ -0,0 +1,85 @@
...
+		val = gofakeit.LoremIpsumWord()
+		break
+	case valueKind == "int":
+		break
+	}
+	return
+}
+
+func (f faker) fakeUserHandle(s string) string {
+	return gofakeit.Generate("??????") + "_handle "
```
darh: how about ???????_seeded?

seeder/commands/faker.go
```diff
@@ -0,0 +1,169 @@
...
+	"context"
+	"fmt"
+	"github.com/cortezaproject/corteza-server/seeder"
+	"strconv"
+
+	"github.com/cortezaproject/corteza-server/pkg/cli"
+	"github.com/spf13/cobra"
+)
+
+func Faker(app serviceInitializer) *cobra.Command {
```
darh: Is it seeder, faker? Can we be more consistant?

seeder/commands/faker.go
```diff
@@ -0,0 +1,169 @@
...
+	// Fake data generation commands.
+	cmd := &cobra.Command{
+		Use:   "faker",
+		Short: "Fake data generation",
+	}
+
+	// Create users
+	createUserCmd := &cobra.Command{
+		Use:   "addUser",
```
darh: this command names are inconsistent:
 - in faker (plural/singular) 
 - with the rest of the system (camelCase, all on one level instead of sub-commands  (addUser vs. user).
 - are you adding or generating or seeding or faking?

seeder/commands/faker.go
```diff
@@ -0,0 +1,169 @@
...
+			fmt.Fprintf(
+				cmd.OutOrStdout(),
+				"                     Cleared    all    fake    records\n",
+			)
+		},
+	}
+
+	// Clear users
+	clearAllFakeDataCmd := &cobra.Command{
+		Use:   "clearData",
```
darh: why data? why not all?

seeder/dataGen.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+var (
+	DefaultStore store.Storer
+)
+
+const (
+	fakeDataLabel       = "generated"
+	fakeUserLabelName   = "generatedUser"
+	fakeRecordLabelName = "generatedRecord"
```
darh: this "generatedUser" or "generatedRecord" does not add any value, do we need to be so explicit?

seeder/dataGen.go
```diff
@@ -0,0 +1,250 @@
...
+	return &lTypes.Label{
+		Kind:       kind,
+		ResourceID: resourceID,
+		Name:       name,
+		Value:      fakeDataLabel,
+	}
+}
+
+// MakeMeSomeFakeUserPlease creates given no of users into DB
+func (gen dataGen) MakeMeSomeFakeUserPlease(opt GenOption) (IDs []uint64, err error) {
```
darh: This needs to be named better.

seeder/dataGen.go
```diff
@@ -0,0 +1,250 @@
...
+	err = gen.store.DeleteUser(gen.ctx, users...)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+// MakeMeSomeFakeRecordPlease creates given no of record into DB
+func (gen dataGen) MakeMeSomeFakeRecordPlease(mod *cTypes.Module, opt GenOption) (IDs []uint64, err error) {
```
darh: This needs to be named better.

seeder/dataGen.go
```diff
@@ -0,0 +1,250 @@
...
+	err = gen.store.UpsertLabel(gen.ctx, labels...)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+// ClearFakeUsers clear all the fake user from DB
+func (gen dataGen) ClearFakeUsers() (err error) {
```
darh: What other kinds of users than fake users are could you be removing here?

seeder/seeder.go
```diff
@@ -0,0 +1,66 @@
+package seeder
```
darh, line 1: Is this even used?

darh (changes requested):
Not a happy camper.

PR #218, opened 2021-07-20
Fix deferred workflow execution

darh:
This fix addresses execution of deferred workflows
and access control when setting run-as user.

Some additional standardisation on how we write
mutex/lock code.

This will also be ported to `2021.9.x`.

### DIFF
--- a/pkg/scheduler/service.go
+++ b/pkg/scheduler/service.go
@@ -19,14 +19,14 @@ type (
 		dispatcher dispatcher
 
 		// Read & write locking
-		l *sync.RWMutex
+		l sync.RWMutex
 
 		// Simple chan to control if service is running or not
 		ticker *time.Ticker
 	}
 
 	dispatcher interface {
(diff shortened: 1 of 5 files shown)


pkg/scheduler/service.go
```diff
@@ -19,14 +19,14 @@ type (
...
 
 		// Read & write locking
-		l *sync.RWMutex
+		l sync.RWMutex
 
 		// Simple chan to control if service is running or not
 		ticker *time.Ticker
 	}
 
 	dispatcher interface {
```
tjerman, line 28: Should the interface be also renamed then?

darh: sorry, what interface?

pkg/scheduler/service.go
```diff
@@ -173,6 +173,11 @@ func (svc service) dispatch(ctx context.Context) {
 	defer svc.l.RUnlock()
```
tjerman, line 173: Move this defer next to the lock

PR #219, opened 2021-07-23
2021.3.x non OIDC missing fix

splack:
I did some other refactoring to use a common constant for the prefix and moved the logic for checking if the provider is valid into types.

I did a bit of cleaning on updateAuthSettings to initialise rather than assign.

I was beaten to the update by 2 mins and didn't have tests.

Feel free to reject this PR or copy useful bits.

### DIFF
--- a/system/types/app_settings.go
+++ b/system/types/app_settings.go
@@ -5,6 +5,10 @@ import (
 	"strings"
 )
 
+const (
+	oidcProviderPrefix = "openid-connect." // must match const in "github.com/cortezaproject/corteza-server/auth/external" external.go
+)
+
 type (
 	// AppSettings type is structured representation of all application settings
 	//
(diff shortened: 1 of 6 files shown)


auth/external/goth.go
```diff
@@ -44,7 +45,7 @@ func SetupGothProviders(log *zap.Logger, redirectUrl string, ep ...settings.Prov
 			redirect = strings.Replace(redirectUrl, "{provider}", pc.Handle, 1)
 		}
 
-		if strings.HasPrefix(pc.Handle, OIDC_PROVIDER_PREFIX) {
+		if strings.HasPrefix(pc.Handle, types.OIDC_PROVIDER_PREFIX) {
```
darh: I don't want this to be moved and have cross component (auth <-> system) dependency.
If there is a better location for this constant, I'm all ears.

system/types/app_settings.go
```diff
@@ -177,6 +181,12 @@ type (
 	}
 )
 
+func (set *ExternalAuthProvider) ValidConfiguration() bool {
+	return (set.Enabled && set.Handle != "" && set.Key != "" && set.Secret != "" &&
```
darh: break this into a series of `if ... { false }` checks for readability please.

darh (changes requested):
There is a comment regarding location of the constant that affects majority of changes here.
I'm ok with initialisation vs. assignment change and emailOTP, TOTP struct types.

Feel free to rebase with the change I made and refactor the new tests and update the PR.

auth/external/goth.go
splack: If there is a common package then it should go there. I'm new to the code base so I've just kept it private to the package and added a comment referencing the other constant.

system/types/app_settings.go
splack: Done

darh:
I'm touching this part of the codebase for another feature. Will merge this (parts of it) through that.

darh:
Merged.

PR #223, opened 2021-07-26
Improve default auth client handling

tjerman:
Prevent default auth clients from changing their handle, being deleted, or de-activated.

Auth app PR: https://github.com/cortezaproject/corteza-webapp-admin/pull/59


tjerman:
Merged; 3 and 9

PR #224, opened 2021-07-27
Add role membership workflow handlers

tjerman:


system/automation/roles_handler.yaml
```diff
@@ -76,6 +86,50 @@ functions:
     results:
       role: *rvRole
 
+  lookupMembers:
```
darh: you either search for members (plural) or lookup for a member (singular)
so this is probably searchMembers, correct?

tests/workflows/testdata/S1001_role_members/data_model.yaml
```diff
@@ -0,0 +1,7 @@
+users:
```
darh: the idea for test numeration was to just do +1, nothing more than that.

darh (changes requested):

system/automation/roles_handler.yaml
tjerman: True

tests/workflows/testdata/S1001_role_members/data_model.yaml
tjerman: Noted

darh (approved):
:shipit:

tjerman:
Merged; 3 and 9

PR #225, opened 2021-07-30
Fix setting record with default values

darh:
Please take a look.

### DIFF
--- a/compose/service/record.go
+++ b/compose/service/record.go
@@ -821,6 +821,26 @@ func (svc record) Create(ctx context.Context, new *types.Record) (rec *types.Rec
 func (svc record) procCreate(ctx context.Context, invokerID uint64, m *types.Module, new *types.Record) (rve *types.RecordValueErrorSet) {
 	new.Values.SetUpdatedFlag(true)
 
+	new.Values.Walk(func(v *types.RecordValue) error {
+		f := m.Fields.FindByName(v.Name)
+		if f == nil {
+			return nil
+		}
+
+		d := f.DefaultValue.Get("", v.Place)
+		if d == nil {
+			// just so that we do not miss any defaults that MIGHT have
+			// field name set to it
+			// this is highly unlikely but it does not hurt to try
+			d = f.DefaultValue.Get(v.Name, v.Place)
+		}
+
+		// Mark as updated ONLY if value set is different from the default one
+		v.Updated = d == nil || d.Value != v.Value
+
+		return nil
+	})
+
 	// Reset values to new record
 	// to make sure nobody slips in something we do not want
 	new.ID = nextID()
--- a/compose/service/record_test.go
+++ b/compose/service/record_test.go
@@ -297,9 +297,145 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
 
 		req.NotNil(recChecked.Values.Get("bool", 0), "should checked again")
 		req.Equal("1", recChecked.Values.Get("bool", 0).Value, "should checked again")
+	}
+}
 
+func TestRecord_defValueFieldPermissionIssue(t *testing.T) {
+	var (
+		req    = require.New(t)
+		ctx    = context.Background()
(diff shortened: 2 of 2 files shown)


compose/service/record_test.go
```diff
@@ -297,9 +297,167 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
...
 		req.Equal("1", recChecked.Values.Get("bool", 0).Value, "should checked again")
+	}
+}
+
+func TestRecord_defValueFieldPermissionIssue(t *testing.T) {
+	var (
+		req = require.New(t)
+
+		// uncomment to enable sql conn debugging
+		//ctx = logger.ContextWithValue(context.Background(), logger.MakeDebugLogger())
```
tjerman: leftover

compose/service/record_test.go
```diff
@@ -297,9 +297,167 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
...
+
+		recPartial *types.Record
+
+		//valChecked = types.RecordValueSet{
+		//	&types.RecordValue{Name: "string", Value: "abc"},
+		//}
+		//
+		//valUnchecked = types.RecordValueSet{
+		//	&types.RecordValue{Name: "string", Value: "abc"},
+		//}
```
tjerman: rm

compose/service/record_test.go
```diff
@@ -297,9 +297,167 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
...
+					out += "NULL"
+				}
+				out += ">"
+			}
+
+			return
+		}
+	)
+
+	t.Log("creating namespace, module and fields")
```
tjerman, line 361: are these logs something we want to keep?

compose/service/record.go
```diff
@@ -620,6 +620,30 @@ func RecordUpdateOwner(invokerID uint64, r, old *types.Record) *types.Record {
 	return r
 }
 
+func RecordValueCreateOpCheck(ctx context.Context, ac recordValueAccessController, m *types.Module, vv types.RecordValueSet) *types.RecordValueErrorSet {
```
tjerman: This is unused as far as I can tell?

compose/service/record_test.go
```diff
@@ -297,9 +297,167 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
...
+
+		recPartial.Values = types.RecordValueSet{&types.RecordValue{Name: "writable", Value: "w2"}}
+
+		recPartial, err = svc.Update(ctx, recPartial)
+		req.NoError(errors.Unwrap(err))
+		req.Equal("<w2><NULL>", valueExtractor(recPartial, "writable", "readable"))
 	}
 
+	{
+		t.Log("creating record with w/o editor role (expecting defaults")
```
tjerman: remove `w/o`

compose/service/record_test.go
```diff
@@ -297,9 +297,167 @@ func TestRecord_boolFieldPermissionIssueKBR(t *testing.T) {
...
+		t.Log("updating record removing one of the values")
+
+		recPartial.Values = types.RecordValueSet{
+			&types.RecordValue{Name: "writable", Value: "def-w"},
+			&types.RecordValue{Name: "readable", Value: "r"},
+		}
+
+		recPartial, err = svc.Update(ctx, recPartial)
+		req.NoError(errors.Unwrap(err))
+		req.Equal("<def-w><r>", valueExtractor(recPartial, "writable", "readable"))
```
tjerman: No value is removed here

tjerman (changes requested):
Why are we doing a `t.Log("...test message details here...")` instead of `t.Run("...", ...)`?

darh:
> Why are we doing a `t.Log("...test message details here...")` instead of `t.Run("...", ...)`?

because I'm telling a story of a complex test case and not trying to run test in parallel.

PR #227, opened 2021-08-03
Added goja lib, transformer lib and jsenv pkg

petergrlica:
Added jsenv support to core pkgs
Support for:
 - custom vars in js
 - custom function definition and wrapper for context


petergrlica:
Already merged

PR #229, opened 2021-08-05
2021.3.x feature split cred check

darh:
Please take a look.

### DIFF
--- a/auth/handlers/handle_login.go
+++ b/auth/handlers/handle_login.go
@@ -1,6 +1,8 @@
 package handlers
 
 import (
+	"fmt"
+
 	"github.com/cortezaproject/corteza-server/auth/request"
 	"github.com/cortezaproject/corteza-server/pkg/auth"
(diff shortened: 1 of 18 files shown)


auth/handlers/handle_login.go
```diff
@@ -17,26 +19,66 @@ type (
...
+				// User w/o the password
+				// In case there is one single IdP automatically redirect user there
+				req.RedirectTo = fmt.Sprintf("%s/%s", GetLinks().External, h.Settings.Providers[0].Handle)
+
+				// for the clarity of the flow
+				// keeping this optional return here
+				return
+			}
+
+			// This user has existing password credentials or with more than one IdP
```
petergrlica, line 69: log debug that there is more than one idp?

petergrlica (approved):
looks good, just one comment/question

auth/handlers/handle_login.go
darh: Fair point. tnx

darh:
merged.

PR #230, opened 2021-08-05
Fix sink request processing for workflow handlers

tjerman:
What is done:

* added missing automation type definitions
* added support for `TypedValue` -> HTTP reponse encoding

Questions:

* should we support a wider range of types when encoding the HTTP response? Currently we just do strings and bytes. We can probably also do maps and such?
* there is a manual tweak in one of the generated files (a cast to `int` from `int64`). I am not sure how we should handle that one.

### DIFF
--- a/system/automation/expr_types.yaml
+++ b/system/automation/expr_types.yaml
@@ -48,6 +48,28 @@ types:
       - { name: 'archivedAt',      exprType: 'DateTime',   goType: '*time.Time', mode: ro }
       - { name: 'deletedAt',       exprType: 'DateTime',   goType: '*time.Time', mode: ro }
 
+  SinkResponse:
+    as: '*types.SinkResponse'
+    struct:
+      - { name: 'status',          exprType: 'Integer',   goType: 'int64'          }
+      - { name: 'header',          exprType: 'KVV',       goType: 'map[string][]string'}
+      - { name: 'body',            exprType: 'Any',       goType: 'interface{}'}
+
+  SinkRequest:
+    as: '*types.SinkRequest'
+    struct:
+      - { name: 'method',         exprType: 'String', goType: 'string'}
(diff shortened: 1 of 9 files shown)


system/automation/expr_types.yaml
```diff
@@ -48,6 +48,28 @@ types:
       - { name: 'archivedAt',      exprType: 'DateTime',   goType: '*time.Time', mode: ro }
       - { name: 'deletedAt',       exprType: 'DateTime',   goType: '*time.Time', mode: ro }
 
+  SinkResponse:
```
darh, line 51: If we go with HttpRequest and HttpResponse we might manage to have future compatibility here. Maybe?

system/automation/expr_types.yaml
```diff
@@ -48,6 +48,28 @@ types:
...
+      - { name: 'method',         exprType: 'String', goType: 'string'}
+      - { name: 'path',           exprType: 'String', goType: 'string'}
+      - { name: 'host',           exprType: 'String', goType: 'string'}
+      - { name: 'header',         exprType: 'KVV',    goType: 'map[string][]string'}
+      - { name: 'query',          exprType: 'KVV',    goType: 'map[string][]string'}
+      - { name: 'username',       exprType: 'String', goType: 'string'}
+      - { name: 'password',       exprType: 'String', goType: 'string'}
+      - { name: 'remoteAddr',     exprType: 'String', goType: 'string'}
+      - { name: 'postForm',       exprType: 'KVV',    goType: 'map[string][]string'}
+      - { name: 'body',           exprType: 'Bytes',  goType: '[]byte'}
```
darh, line 70: Could this be io.Reader?  Please take a look at `automation/automation/http_request_handler.yaml`.

darh (commented):
I'm sure comments need extra explanation. Feel free to nudge me.

PR #232, opened 2021-08-11
API Gateway feature

petergrlica:
Added api gateway to server.

Please check mostly pkg/apigw, as the system CRUD is fairly common.


automation/automation/jsenv_handler.yaml
```diff
@@ -0,0 +1,25 @@
...
+    meta:
+      short: Process arbitrary data in jsenv
+    params:
+      scope:
+        required: true
+        types:
+          - { wf: Any }
+          - { wf: String, suffix: String }
+          - { wf: Reader, suffix: Stream }
+      function:
```
darh: would "source" be a better name for this here?

automation/automation/jsenv_handler.yaml
```diff
@@ -0,0 +1,25 @@
+imports:
+  - io
+
+functions:
+  process:
```
darh: maybe `execute` instead of process?

automation/automation/jsenv_handler.go
```diff
@@ -0,0 +1,66 @@
...
+	out, err := fn.Exec(vm.New(args.scopeString))
+
+	if err != nil {
+		err = fmt.Errorf("could not exec jsenv function: %s", err)
+		return
+	}
+
+	switch vv := out.(type) {
+	case uint64, int64:
+		res.ResultInt = int64(vv.(uint64))
```
darh, line 61: can you just add a quick comment to these two that they might be removed (if I understand you correctly, this is the case here, right?)

darh (commented):
non obligatory suggestions for improvements

automation/automation/jsenv_handler.yaml
petergrlica: it would, thanks

automation/automation/jsenv_handler.go
petergrlica: :+1:

PR #235, opened 2021-08-16
Fix encoding on save & re-encoding on fetch

darh:
Please take a look.

### DIFF
--- a/system/rest/actionlog.go
+++ b/system/rest/actionlog.go
@@ -2,6 +2,8 @@ package rest
 
 import (
 	"context"
+	"strconv"
+	"strings"
 
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
 	"github.com/cortezaproject/corteza-server/pkg/filter"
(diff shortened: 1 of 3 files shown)


system/rest/actionlog.go
```diff
@@ -115,6 +122,24 @@ func (ctrl Actionlog) makeFilterPayload(ctx context.Context, ee []*actionlog.Act
 	return &actionlogPayload{Filter: f, Set: pp}, nil
 }
 
+// Making sure JS can read our values
+func sanitizeMapStringInterface(m map[string]interface{}) {
+	for k := range m {
+		fmt.Printf("%s: (%T) %s\n", k, m[k], m[k])
```
tjerman: Remove

tjerman (commented):
I think that the mix and matching of numbers and strings is not the play here (stringify if too large, keep as is if small enough).
I think the better play here is to stringify all of the numbers to keep them consistent.

PR #236, opened 2021-08-17
Allow store functions to define codegen imports

tjerman:
Please take a look.


darh (approved):
What I want(ed) to avoid is putting this kind of logic into template (where it could be handled on data prep. side)

_I'm ok with this for the old codegen._

PR #238, opened 2021-08-19
Adds expr functions

vicpatel:
modWeek, random for number, sub for date

Please take a look.

### DIFF
--- a/pkg/expr/func_num.go
+++ b/pkg/expr/func_num.go
@@ -1,9 +1,10 @@
 package expr
 
 import (
-	"math"
-
 	"github.com/PaesslerAG/gval"
+	"github.com/pkg/errors"
+	"math"
+	"math/rand"
 )
 
 func NumericFunctions() []gval.Language {
@@ -19,6 +20,7 @@ func NumericFunctions() []gval.Language {
 		gval.Function("sqrt", math.Sqrt),
 		gval.Function("sum", sum),
 		gval.Function("average", average),
+		gval.Function("random", random),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/expr/func_num.go
```diff
@@ -112,3 +115,24 @@ func average(v ...interface{}) float64 {
...
+	case 1:
+		if to = v[0]; to < 0 {
+			return 0, errors.New("unexpected input type")
+		}
+	case 2:
+		if from, to = v[0], v[1]; from < 0 || to < 0 {
+			return 0, errors.New("unexpected input type")
+		}
+	default:
+		return 0, errors.New("insufficient input")
```
tjerman: I'd prefer to see something in the lines of:

> Expecting 1 or 2 parameter, got <count here>

pkg/expr/func_num.go
```diff
@@ -112,3 +115,24 @@ func average(v ...interface{}) float64 {
...
+	}
+
+	rand.Seed(time.Now().UnixNano())
```
tjerman: I think that seeding is already done somewhere in initialization; try to see if you can find instance, else ignore this comment.

pkg/expr/func_num.go
```diff
@@ -112,3 +115,24 @@ func average(v ...interface{}) float64 {
...
+
+	switch len(v) {
+	case 1:
+		if to = v[0]; to < 0 {
+			return 0, errors.New("unexpected input type")
+		}
+	case 2:
+		if from, to = v[0], v[1]; from < 0 || to < 0 {
+			return 0, errors.New("unexpected input type")
+		}
```
tjerman: I would do this with multiple ifs to avoid repeating logic.
Something in the lines of:

```
if len(v) > 0 {
  if v[0] < 0 { error... }
}
if len(v) > 1 {
  if v[1] < 0 {error ... }
}
...
```

Personal preference; feel free to ignore.

pkg/expr/func_time.go
```diff
@@ -195,3 +196,27 @@ func strfTime(base interface{}, f string) (string, error) {
...
+	var duration time.Duration
+	if t1.Before(*t2) {
+		duration = t2.Sub(*t1)
+	}
+
+	if t1.After(*t2) {
+		duration = t1.Sub(*t2)
+	}
+
+	return duration.Microseconds(), nil
```
tjerman: The comment states miliseconds but here you do microseconds; do milliseconds as further accuracy is not needed.

pkg/expr/func_time.go
```diff
@@ -195,3 +196,27 @@ func strfTime(base interface{}, f string) (string, error) {
 	o, _ := strftime.Format(f, *t, strftime.WithMilliseconds('b'))
 	return o, nil
 }
+
+// sub returns difference between two date into milliseconds
+func sub(from interface{}, to interface{}) (out int64, err error) {
```
tjerman, line 201: I would prefer to see that we either throw an error if `from` is less then `to`, or return a negative number.

tjerman (changes requested):

tjerman (approved):
lgtm.

Please make sure that the function reference in the documentation mentions that the result of the sub function is in milliseconds.

PR #243, opened 2021-08-31
Added support for async routes

petergrlica:


pkg/apigw/pipeline/pipeline.go
```diff
@@ -46,56 +52,64 @@ func (pp *Pl) ErrorHandler(ff types.ErrorHandlerFunc) {
...
+				err = hh.Handler(rw, r)
+				log.Debug("finished processing", zap.Duration("duration", time.Since(start)))
+				return
+			}
 
-			pp.log.Debug("finished processing",
-				zap.String("filter", hh.Name),
-				zap.Duration("duration", time.Since(start)))
+			if hh.Async {
+				go fn()
```
darh: do we need to log the error here?

darh: how do we handle max-conncurrency?

darh: add 2 sep. task for max-concurrency
 - general API rate limiting (chi)
 - per-route rate limiting (chi)
 - goroutine concurrency w/ channles (??)

darh (approved):
lgtm + add tasks

PR #246, opened 2021-09-01
CLI user command update

vicpatel:
Extends `user add` CLI command

- Adds flag `make-password-link` to generate temp token to create users password and provide link

Please take a look.

Related locale PR# https://github.com/cortezaproject/corteza-locale/pull/2

### DIFF
--- a/auth/handlers/handle_create-password.go
+++ b/auth/handlers/handle_create-password.go
@@ -1,0 +1,117 @@
+package handlers
+
+import (
+	"github.com/cortezaproject/corteza-server/auth/request"
+	"github.com/cortezaproject/corteza-server/pkg/auth"
+	"github.com/cortezaproject/corteza-server/system/service"
+	"github.com/cortezaproject/corteza-server/system/types"
+	"go.uber.org/zap"
+)
+
+var (
+	loginLink          = GetLinks().Login
+	profileLink        = GetLinks().Profile
+	createPasswordLink = GetLinks().CreatePassword
+)
+
+func (h *AuthHandlers) createPasswordForm(req *request.AuthReq) (err error) {
+	h.Log.Debug("password create form")
+	req.Template = TmplCreatePassword
+
+	// user not set, expecting valid token in URL
+	if token := req.Request.URL.Query().Get("token"); len(token) > 0 {
+		var (
+			user *types.User
+			// If user has password or not
+			passwordSet bool
+		)
+
+		user, err = h.AuthService.ValidatePasswordCreateToken(req.Context(), token)
+		if err == nil {
+			// If user does not have password (or in case there is no user with such email)
+			passwordSet = h.AuthService.PasswordSet(req.Context(), user.Email)
+			if !passwordSet {
+				// login user
+				req.AuthUser = request.NewAuthUser(h.Settings, user, false, h.Opt.SessionLifetime)
+
+				// redirect back to self (but without token and with user in session)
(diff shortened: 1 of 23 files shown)


system/service/auth.go
```diff
@@ -1371,3 +1425,23 @@ func (svc auth) LoadRoleMemberships(ctx context.Context, u *types.User) error {
 func (svc auth) GetProviders() types.ExternalAuthProviderSet {
 	return CurrentSettings.Auth.External.Providers
 }
+
+// randomStringWithRestrictions generates a random ASCII string with at least one digit and one special character
+func randomStringWithRestrictions() string {
```
darh: Could this work if we move it to `pkg/rand:Password()`? You can design it in a similar way as `rand.Bytes()`

system/service/auth.go
```diff
@@ -960,6 +1003,17 @@ func (svc auth) createUserToken(ctx context.Context, u *types.User, kind string)
 
 			// random number, 6 chars
 			token = fmt.Sprintf("%06d", rand2.Int())[0:6]
+		case credentialsTypeCreatePasswordToken:
+			// todo?: svc.settings.Auth.Internal.PasswordCreate.Expires
```
darh: sure.

auth/handlers/handle_create-password.go
```diff
@@ -0,0 +1,54 @@
+package handlers
+
+import (
+	"github.com/cortezaproject/corteza-server/auth/request"
+	"github.com/cortezaproject/corteza-server/pkg/auth"
+	"github.com/cortezaproject/corteza-server/system/service"
+	"go.uber.org/zap"
+)
+
+func (h *AuthHandlers) createPasswordForm(req *request.AuthReq) error {
```
darh: We could/should/must(?) protect both these endpoints a bit.
Can user that already has password set use this? 
There is `PasswordSet()` fn on auth service that checks this.

system/types/app_settings.go
```diff
@@ -50,6 +50,12 @@ type (
 				// Can users reset their passwords
 				PasswordReset struct{ Enabled bool } `kv:"password-reset"`
 
+				// Can users create their passwords
```
darh: _This little thing did not get submitted when I reviewed this yesterday._

Who creates passwords for whom? This comment does not really describes it right...

darh (approved):

system/service/auth_notification.go
```diff
@@ -63,6 +64,10 @@ func (svc authNotification) PasswordReset(ctx context.Context, emailAddress stri
 	})
 }
 
+func (svc authNotification) PasswordCreate(_ context.Context, _ string, token string) (string, error) {
```
darh: why keep the 2st two args if you do not need them?

system/types/app_settings.go
```diff
@@ -50,6 +50,15 @@ type (
 				// Can users reset their passwords
 				PasswordReset struct{ Enabled bool } `kv:"password-reset"`
 
+				// PasswordCreate setting for create password for user via generated link with token
+				// If user has no password then link redirects to create password page
+				// Otherwise it redirects to profile page of that user
+				// link can be generated through useradd cli command with `make-password-link` flag
+				PasswordCreate struct {
```
darh, line 57: This needs to be supported in the admin panel as well.
Can you add a task for next sprint with instructions so one of the frontend devs can pick it up (as an alternative to you doing it ;))

darh (changes requested):

PR #250, opened 2021-09-06
2021.9.x feature cc

tjerman:


store/configurable_content.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+types:
+  type: types.ConfigurableContent
+
+fields:
+  - { field: ID               }
+
+  - { field: Lang             }
+  - { field: Resource         }
+  - { field: K                }
```
darh: I would go with key and message (msg) here. No extreme abbreviations.

PR #251, opened 2021-09-08
Updates expr function `count` to return count based on params

vicpatel:
- count(foo) counts all elements
- count(foo, 'bar', 'baz') counts number of bars and bazs in foo

Please take a look.


darh (approved):
LGTM. Please do not forget to update the docs.

PR #254, opened 2021-09-17
Adds JWT token generation methods

vicpatel:
- Generate creates token encoded with user and save it to store
- Updated token generation method usage in  Corredor and Impersonalization

Please take a look.

### DIFF
--- a/app/boot_levels.go
+++ b/app/boot_levels.go
@@ -406,6 +406,8 @@ func (app *CortezaApp) InitServices(ctx context.Context) (err error) {
 		return
 	}
 
+	auth.SetJWTStore(app.Store)
+
 	corredor.Service().SetUserFinder(sysService.DefaultUser)
 	corredor.Service().SetRoleFinder(sysService.DefaultRole)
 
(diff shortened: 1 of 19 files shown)


pkg/auth/jwt_token.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type (
+	ExtraReqInfo struct {
+		RemoteAddr string
+		UserAgent  string
+	}
+)
+
+func (t *token) Generate(ctx context.Context, s tokenStore, i Identifiable) (tokenString string, err error) {
```
darh: How about if we keep the `tokenStore` directly on the `token` struct? This will allow us to call Generate without store and keep it as lightweight as possible + no passing store to the corredor's Connect.

Would that make sense?

darh (changes requested):

pkg/auth/jwt_token.go
vicpatel: Yes

darh (approved):

PR #255, opened 2021-09-17
 APIGw removed permissions for filters

vicpatel:
Please take a look.


system/service/apigw_filter.go
```diff
@@ -15,44 +14,37 @@ type (
...
 	}
 
-	functionAccessController interface {
-		CanSearchApigwFilters(context.Context) bool
-
-		CanCreateApigwFilter(context.Context) bool
-		CanReadApigwFilter(context.Context, *types.ApigwFilter) bool
-		CanUpdateApigwFilter(context.Context, *types.ApigwFilter) bool
-		CanDeleteApigwFilter(context.Context, *types.ApigwFilter) bool
+	filterAccessController interface {
```
tjerman: make it `routeAccessController`

system/service/apigw_filter.go
```diff
@@ -233,14 +225,14 @@ func (svc *apigwFilter) UndeleteByID(ctx context.Context, ID uint64) (err error)
...
-			return ApigwFilterErrNotAllowedToDelete(qProps)
-		}
-
 		if r, err = svc.route.FindByID(ctx, q.Route); err != nil {
 			return err
 		}
 
+		if !svc.ac.CanDeleteApigwRoute(ctx, r) {
+			return ApigwRouteErrNotAllowedToUndelete()
+		}
```
tjerman: The undelete function should check if user can **update** the route, not if they can delete it.

system/service/apigw_filter.go
```diff
@@ -268,10 +260,6 @@ func (svc *apigwFilter) Search(ctx context.Context, filter types.ApigwFilterFilt
 
 	// For each fetched item, store backend will check if it is valid or not
 	filter.Check = func(res *types.ApigwFilter) (bool, error) {
-		if !svc.ac.CanReadApigwFilter(ctx, res) {
-			return false, nil
-		}
-
```
tjerman: If the rest of the operations check access based on the API gateway route, this should also do that.
That said, it might be a bit annoying as we need to fetch the route as well.

You could preload all of the routes, throw them in a hashmap and then use that here.
Alternatively, we could also limit route filtering to allow searching only over a specific route which would make this trivial to handle.

The same applies to the following two deleted blocks.

tjerman (changes requested):
A few changes; other then that it seems fine to me :+1:

PR #258, opened 2021-09-20
2021.3.x feature invoker runner

darh:
When workflow loads we need to set one or two new variables to scope (both of type `system.User`):

`invoker` for the user that triggered the workflow. For events that are not triggered by a user (on interval, on timestamp, on request) but as consequence of some scheduled or external event this value is NOT set.

`runner` for user that executed the workflow. If runAs is configured on the workflow it points to the run-as user. Otherwise it is the same as `invoker`

This will also be applied to 2021.9.x.


darh:
Merged.

PR #266, opened 2021-09-29
Fixes apigw prefilter for string with hyphen

vicpatel:
Please take a look.

### DIFF
--- a/pkg/apigw/filter/prefilter.go
+++ b/pkg/apigw/filter/prefilter.go
@@ -100,13 +100,16 @@ func (h header) Handler() types.HandlerFunc {
 
 		// get the request data and put it into vars
 		out, err := expr.NewVars(vv)
-
 		if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/apigw/filter/prefilter.go
```diff
@@ -89,17 +89,19 @@ func (h header) Handler() types.HandlerFunc {
 	return func(rw http.ResponseWriter, r *http.Request) error {
 		var (
 			ctx = r.Context()
+			vv  = map[string]interface{}{}
```
darh: Why are we solving this with `map[string]interface`? why not http.Header that can be converted into `expr.KVV`?

darh (changes requested):

tjerman (changes requested):
Please don't modify my existing tests but add 2 new ones; one for query params and one for headers.

PR #267, opened 2021-09-30
Implements LOCALE_RESOURCE_TRANSLATIONS_ENABLED

darh:


compose/service/locale.go
```diff
@@ -183,3 +191,42 @@ func (svc resourceTranslationsManager) loadChart(ctx context.Context, s store.St
...
+			return nil
+		}
+
+		if !lsvc.Locale().SupportedLang(contentLang) {
+			// unsupported language
+			return errors.InvalidData("unsupported language")
+		}
+
+		if !ac.CanManageResourceTranslations(ctx) {
+			return errors.Unauthorized("resource translation not supported")
```
tjerman: shouldn't this be "not allowed to translate resources", not unsupported?

compose/service/locale.go
```diff
@@ -183,3 +191,42 @@ func (svc resourceTranslationsManager) loadChart(ctx context.Context, s store.St
 	_, m, err = loadChart(ctx, s, namespaceID, chartID)
 	return m, err
 }
+
+func updateTranslations(ctx context.Context, ac localeAccessControl, lsvc ResourceTranslationsManagerService, tt ...*locale.ResourceTranslation) error {
```
tjerman: shouldn't this be placed in the generated service?

tjerman (changes requested):

compose/service/locale.go
darh: correct.

tjerman (approved):

PR #276, opened 2021-10-08
Fixes Wf duplicate issues steps

vicpatel:
Prevents the steps from verifying again by fixing return logic for converting those step definitions into workflow.Step instances

Please take a look.


darh (changes requested):
There are a few `fmt.Println` leftorvers. Let's wait for @tjerman feedback too.

PR #278, opened 2021-10-12
Fix create and update with unique constraints

darh:
When creating or updating resource that did not match unique constraint
filters, check wrongly reported not-unique error when matching (and
valid) resource was found in the store.

This fix adds explicit check if resource to be stored does not match
constraint filters and skips the rest of the constraint checking
procedure.


tjerman (approved):
lgtm

PR #280, opened 2021-10-13
Decouple messagebus package

petergrlica:
What has been done:
 - removed system layer out of pkg/messagebus
 - decoupled rest and store from pkg
 - added a new Bytes type to use with queue message


darh (approved):

darh:
Kudos. Merge please.

PR #283, opened 2021-10-26
Support recording and searching actionlogs

darh:
Please take a look.


tjerman (approved):
We've also said we'll be logging the actual invoker alongside the actorID (to cover cases where runAs is defined).
Is this not the case anymore?

Other then that this seems fine to me :+1:

PR #286, opened 2021-11-02
Fixes workflow preprocessing

vicpatel:
Adds counter for dependent step verification to prevent pending dependent workflow step recursion

Please take a look.

### DIFF
--- a/automation/service/workflow_converter.go
+++ b/automation/service/workflow_converter.go
@@ -68,6 +68,7 @@ func (svc workflowConverter) makeGraph(def *types.Workflow) (*wfexec.Graph, type
 		return true, nil
 	})
 
+	counter := 0
 	makeGraphSteps = func(steps types.WorkflowStepSet) {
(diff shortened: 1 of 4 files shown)


tjerman (changes requested):
Can you please add a set of tests for error handler steps, similar to how the gateway tests look like.
Since the initial issue was due to multiple outbound paths, this case should also be handled.
If all that passes, it is fine by me.

tjerman (changes requested):
It seems that I read the test cases too quickly the last time, apologies.
Anyway, this is what needs to happen.

1. Create a new test file for testing that the workflow gets properly preprocessed/converted -- currently it dies in an infinite recursion for some cases. These tests to collect step issues are fine.
2. Use more descriptive handles for your workflows inside the `.yaml` file.

Additional test cases that must be covered:

Screenshot from 2021-11-03 17-11-54

The above-shown changing of branches kills the preprocessing for me.
Construct a similar workflow test case for changing error handler steps.

PR #287, opened 2021-11-03
Refactor role security, support it for ext. auth flow

darh:
Refactors role security as it was implemented on auth-clients into more reusable function.
Implement same/similar role-security structure on external auth providers and allow permitted/prohibited/forbidden roles to be applied when user authenticates via IdP.

### DIFF
--- a/pkg/auth/role-security.go
+++ b/pkg/auth/role-security.go
@@ -1,0 +1,41 @@
+package auth
+
+import (
+	"sort"
+
+	"github.com/cortezaproject/corteza-server/pkg/slice"
+)
+
+// ApplyRoleSecurity takes role security params (set of permitted, prohibited and forced roles)
+// and applies these rules to the set of given roles
+//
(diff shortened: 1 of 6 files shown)


pkg/auth/role-security.go
```diff
@@ -0,0 +1,41 @@
...
+
+	out = make([]uint64, 0, len(mForced))
+	for forcedRoleID := range mForced {
+		out = append(out, forcedRoleID)
+	}
+
+	// for stable output
+	sort.Slice(out, func(i, j int) bool {
+		return out[i] < out[j]
+	})
```
tjerman, line 38: For numerical arrays, IIRC, you have a built-in sort function.
Perhaps worth swapping it out with that, if it exists.

tjerman (commented):
Other then the comment, lgtm

petergrlica (approved):

darh:
merged manually.

PR #292, opened 2021-11-18
Corteza discovery

vicpatel:
- It provides mappings of corteza resources(Namespaces, Modules, Records, Users) along with values to corteza discovery indexer.
- It also save recordLogs for create, update, deletion of resources.
- Extend settings to hold discovery enabled/disabled flag
- Adds URL of compose resources for namespace and record and for module still pending, added todo with notes to code

Please take a look.


discovery/rest/feed.go
```diff
@@ -0,0 +1,20 @@
...
+	feed struct {
+	}
+)
+
+func Feed() *feed {
+	return &feed{}
+}
+
+func (ctrl feed) Changes(ctx context.Context, r *request.FeedChanges) (interface{}, error) {
+	return nil, fmt.Errorf("not implemented")
```
darh, line 19: how are we fetching the changes then?

provision/001_settings/discovery.yaml
```diff
@@ -0,0 +1,12 @@
+settings:
```
darh, line 1: move that under 400_discovery dir please

PR #293, opened 2021-11-19
Fix bool/false value sanitization

darh:
Boolean value sanitization MUST return empty string for false boolean
value to ensure consisten behaviour when empty or false value for field
is send back and that field is read-only (via RBAC).

Please take a look.


tjerman (approved):
LGTM ship it :+1:

PR #294, opened 2021-11-20
Implement support for base password constraints

tjerman:
Please note the added JSON tags in the settings struct; that was required for the current settings response.
I am not sure if we want to proceed with such an approach or do we wish to define a wrapper structure to make sure we are only exposing the settings we must.

Please take a look.

### DIFF
--- a/pkg/options/auth.yaml
+++ b/pkg/options/auth.yaml
@@ -10,6 +10,17 @@ props:
     description: |-
       Enable extra logging for authentication flows
 
+  - name: passwordSecurity
+    type: bool
+    default: true
+    description: |-
+      Password security allows you to disable constraints to which passwords must conform to.
+      
+      [CAUTION]
+      ====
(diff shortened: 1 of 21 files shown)


pkg/options/auth.yaml
```diff
@@ -10,6 +10,17 @@ props:
...
 
+  - name: passwordSecurity
+    type: bool
+    default: true
+    description: |-
+      Password security allows you to disable constraints to which passwords must conform to.
+      
+      [CAUTION]
+      ====
+      Password security should not be disabled on production environments as it is insecure.
```
darh, line 21: >  ....environments as it is insecure.

Could probably be written a little bit better? Maybe:

"Keep enabled in production environment to avoid security incidents."

darh (changes requested):
Not really sure why there was a need to expose settings as you did? Where is this needed?

PR #296, opened 2021-11-25
Add SECURITY.md

vicpatel:
Please take a look.

### DIFF
--- a/security.md
+++ b/security.md
@@ -1,0 +1,37 @@
+# Security
+
+At Corteza, the security of our products and services is important to us.
+All of our source code repositories are managed through GitHub organisations.
+Here is the list of [Corteza Repositories](https://github.com/orgs/cortezaproject/repositories)
+
+If you believe you have found a security vulnerability in any Corteza repository, please report it to us as described below.
+
+## Reporting Security Issues
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+Instead, please send email to [info@cortezaproject.org](mailto:info@cortezaproject.org).
+
+You should receive a response within 24 business hours. If for some reason you do not,
+please follow up via email to ensure we received your original message. 
+
+Please include the requested information listed below (as much as you can provide)
+to help us better understand the nature and scope of the possible issue:
+
+* Type of issue,
+* full paths of source file(s) related to the manifestation of the issue,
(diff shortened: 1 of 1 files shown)


security.md
```diff
@@ -0,0 +1,39 @@
...
+which includes all source code repositories managed through our GitHub organizations.
+Here is the list of [Corteza Repositories](https://github.com/orgs/cortezaproject/repositories)
+
+If you believe you have found a security vulnerability in any Corteza repository, please report it to us as described below.
+
+## Reporting Security Issues
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+Instead, please send email to [security@cortezaproject.org](mailto:security@cortezaproject.org).
```
tjerman: Is this a valid email address?
Please validate.

security.md
```diff
@@ -0,0 +1,39 @@
...
+
+If you prefer to submit via forum, please report them to [Corteza Forum](https://forum.cortezaproject.org/).
```
tjerman: I would exclude the forum as it is public also, plus I think it's easier for us to track it if its via email (for now)

security.md
```diff
@@ -0,0 +1,39 @@
...
+Please include the requested information listed below (as much as you can provide)
+to help us better understand the nature and scope of the possible issue:
+
+* Type of issue
+* Full paths of source file(s) related to the manifestation of the issue
+* The location of the affected source code (tag/branch/commit or direct URL)
+* Any special configuration required to reproduce the issue
+* Step-by-step instructions to reproduce the issue
+* Proof-of-concept or exploit code (if possible)
+* Impact of the issue, including how an attacker might exploit the issue
```
tjerman: Have the first bullet start with a capital letter (leave as is), the rest should start with a lower-case letter (change).
All lines except for the last one should end with a comma (,) and the last with a period (.)

tjerman (changes requested):
Please follow the one sentence per line rule.
Other then that it seems fine to me.

security.md
darh: validated. :)

security.md
vicpatel: Yeah sure.

zmija: @vicpatel we'll have to use info@cortezaproject.org here instead of security@

PR #299, opened 2021-11-26
Add route to copy permissions to role

vicpatel:
It copies all RBAC rules from one role to another, but also removes all existing rules from role.

Please take a look.

### DIFF
--- a/pkg/rbac/service.go
+++ b/pkg/rbac/service.go
@@ -279,3 +279,33 @@ func (svc service) String() (out string) {
 
 	return
 }
+
+// CloneRulesByRoleID clone all rules of a Role S to a specific Role T by removing its existing rules
+func (svc *service) CloneRulesByRoleID(ctx context.Context, fromRoleID uint64, toRoleID ...uint64) (err error) {
+	var (
+		updatedRules RuleSet
+	)
+
+	// Make sure rules of fromRoleID stays intact
+	rr := svc.FindRulesByRoleID(fromRoleID)
+
+	for _, roleID := range toRoleID {
+		// Remove existing rules
+		existingRules := svc.FindRulesByRoleID(roleID)
+		for _, rule := range existingRules {
+			// Make sure to remove existing rules
+			rule.Access = Inherit
+		}
+		updatedRules = append(updatedRules, existingRules...)
+
+		// Clone rules from role S to role T
+		for _, rule := range rr {
+			// Make sure everything is properly set
+			r := *rule
+			r.RoleID = roleID
+			updatedRules = append(updatedRules, &r)
+		}
+	}
+
+	return svc.Grant(ctx, updatedRules...)
+}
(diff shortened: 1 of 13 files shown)


system/rest.yaml
```diff
@@ -944,6 +944,20 @@ endpoints:
         type: rbac.RuleSet
         required: true
         title: List of permission rules to set
+  - name: copy
+    path: "/{roleID}/rules/{originRoleID}"
```
darh: Path is a bit misleading... Give it another look please

system/rest/permissions.go
```diff
@@ -61,3 +61,29 @@ func (ctrl Permissions) Update(ctx context.Context, r *request.PermissionsUpdate
 
 	return api.OK(), ctrl.ac.Grant(ctx, r.Rules...)
 }
+
+func (ctrl Permissions) Copy(ctx context.Context, r *request.PermissionsCopy) (interface{}, error) {
```
darh: This is excellent -- but it's too much logic (that should be part of the service) in the controller.

system/service/user_actions.gen.go
```diff
@@ -11,13 +11,12 @@ package service
...
 	"fmt"
-	"strings"
-	"time"
-
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
 	"github.com/cortezaproject/corteza-server/pkg/errors"
 	"github.com/cortezaproject/corteza-server/pkg/locale"
 	"github.com/cortezaproject/corteza-server/system/types"
+	"strings"
+	"time"
```
darh, line 19: this was probably a mistake?

system/service/auth_actions.gen.go
```diff
@@ -11,13 +11,12 @@ package service
...
 	"fmt"
-	"strings"
-	"time"
-
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
 	"github.com/cortezaproject/corteza-server/pkg/errors"
 	"github.com/cortezaproject/corteza-server/pkg/locale"
 	"github.com/cortezaproject/corteza-server/system/types"
+	"strings"
+	"time"
```
darh, line 19: this was probably a mistake?

darh (changes requested):

system/service/auth_actions.gen.go
vicpatel: yes

pkg/rbac/service.go
```diff
@@ -279,3 +279,31 @@ func (svc service) String() (out string) {
 
 	return
 }
+
+// CloneRulesByRoleID clone all rules of a Role S to a specific Role T
+//      returns set of role T rules by replacing it's existing rules with role S rules
+//
+// This function is auto-generated
```
darh: is it? :)

darh (changes requested):
excellent. small (comment) fix, tests & you are done!

PR #301, opened 2021-11-29
 2021.9.x feature wf attachments

darh:


compose/automation/expr_types.yaml
```diff
@@ -51,6 +51,38 @@ types:
...
+#      - { name: 'image',  exprType: 'AttachmentMetaImage',       goType: 'string' }
+#
+#  AttachmentMetaImage:
+#    as: '*types.AttachmentImageMeta'
+#    struct:
+#      - { name: 'weight',          exprType: 'Integer',      goType: 'int',  mode: ro }
+#      - { name: 'height',          exprType: 'Integer',      goType: 'int',  mode: ro }
+#      - { name: 'animated',        exprType: 'Boolean',      goType: 'bool', mode: ro }
+
+
```
tjerman, line 85: What's up with this?
Should we just remove it if we don't want it (yet)?

tjerman (commented):
See comment; else lgtm

PR #304, opened 2021-12-06
Generate JWT workflow function

petergrlica:
JWT generate handler with:
 - payload
 - headers
 - scope
 - cert

Also added UNIX strftime support to gval functions

### DIFF
--- a/pkg/expr/func_num.go
+++ b/pkg/expr/func_num.go
@@ -1,10 +1,12 @@
 package expr
 
 import (
-	"github.com/PaesslerAG/gval"
-	"github.com/pkg/errors"
 	"math"
 	"math/rand"
+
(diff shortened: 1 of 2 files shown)


automation/automation/jwt_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	h.register()
+	return h
+}
+
+func (h jwtHandler) generate(ctx context.Context, args *jwtGenerateArgs) (res *jwtGenerateResults, err error) {
+	var (
+		secret interface{}
+		auxp   map[string]interface{} = make(map[string]interface{})
```
darh: redundant types.

darh (changes requested):
Add a test in `tests/workflows`

Issue #295, opened 2021-11-22
minio storage not initialising


andre-vspry:
**Describe the bug**
Set the various minio_ variables pointing to a bucket and corteza serve-api exits with error.

**To Reproduce**

```go
package main

import (
	"bytes"
	"log"

	"github.com/cortezaproject/corteza-server/pkg/objstore/minio"
)

func main() {
	m, err := minio.New("test-4/system", minio.Options{
		Endpoint:        "storage.googleapis.com",
		Secure:          true,
		AccessKeyID:     "GOOG1EXXXXXX",
		SecretAccessKey: "XXXXXXXX",
	})
	if err != nil {
		log.Fatal(err)
	}

	b := bytes.NewBuffer([]byte("test"))
	if err := m.Save("test", b); err != nil {
		log.Fatal(err)
	}
}
```

**Expected behavior**
corteza starts serving the API.

**Screenshots**
```
Nov 23 09:52:44 test-4 systemd[1]: Started Corteza low code server.
Nov 23 09:52:44 test-4 corteza-start[7205]: {"level":"warn","ts":1637625164.3658867,"msg":"loading plugins","paths":""}
Nov 23 09:52:44 test-4 corteza-start[7205]: {"level":"warn","ts":1637625164.3659208,"msg":"loading plugins","paths":[]}
Nov 23 09:52:44 test-4 corteza-start[7205]: Error: Bucket name contains invalid characters
Nov 23 09:52:44 test-4 corteza-start[7205]: Usage:
Nov 23 09:52:44 test-4 corteza-start[7205]:   corteza-server serve-api [flags]
Nov 23 09:52:44 test-4 corteza-start[7205]: Aliases:
Nov 23 09:52:44 test-4 corteza-start[7205]:   serve-api, serve
Nov 23 09:52:44 test-4 corteza-start[7205]: Flags:
Nov 23 09:52:44 test-4 corteza-start[7205]:   -h, --help   help for serve-api
Nov 23 09:52:44 test-4 corteza-start[7205]: Bucket name contains invalid characters
Nov 23 09:52:44 test-4 systemd[1]: corteza.service: Main process exited, code=exited, status=1/FAILURE
```

**Additional context**
Looks like the path is being added to the bucket name which causes it to fail the validity checks.

https://github.com/cortezaproject/corteza-server/blob/c5b730b524a91d5cac997c7aa252c8f276825854/system/service/service.go#L132

[...]

PR #306, opened 2021-12-15
Refactor min.io object storage

vicpatel:
- Introduces env variable for `MINIO_PATH_PREFIX` for more flexibility over bucket naming
- Provides method for generating bucket name from bucket related env variables
- Decouples minio client from New method
- Adds tests

 Related to Issue #295

Please take a look.

### DIFF
--- a/pkg/objstore/minio/store.go
+++ b/pkg/objstore/minio/store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	minio "github.com/minio/minio-go/v6"
 	"github.com/minio/minio-go/v6/pkg/encrypt"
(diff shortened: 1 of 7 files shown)


pkg/objstore/minio/store.go
```diff
@@ -121,6 +133,17 @@ func (s store) Open(name string) (io.ReadSeeker, error) {
 	})
 }
 
-func (s *store) Healthcheck(ctx context.Context) error {
+func (s *store) Healthcheck(_ context.Context) error {
 	return nil
 }
+
+// GenerateBucketName return bucket name based on path prefix, separator, bucketName
+func GenerateBucketName(pathPrefix, separator, name, componentName string) (out string) {
```
darh: please explain why are you mixing bucket and path concepts here? 

We've discussed approach where 2 env variables (options) would be introduced 
`MINIO_BUCKET_NAME`
`MINIO_PATH_PREFIX`

Both would go through simple string replacement, replacing `{component}` (if it would exist of corse) with system|compose|....

Then the both options would be passed to the client where bucket would be used same as today and path-prefix would be applied within save/remove/open.

What am I missing here?

darh (changes requested):

pkg/objstore/minio/store.go
vicpatel: No, you are right, my bad I will fix it.

darh (approved):

PR #308, opened 2021-12-20
Allow format QL nodes to properly format their arguents

tjerman:
The argument formatting function was not applied to the arguments so pgsql date formatting didn't work properly.
Should also get ported to `.3`


darh (commented):
How much effort to construct an integration or store test for this?

darh:
Other than tests -- LGTM.

tjerman:
Done

PR #312, opened 2021-12-29
Add expr function for sorting an array

vicpatel:
Please take a look.

### DIFF
--- a/pkg/expr/func_arr.go
+++ b/pkg/expr/func_arr.go
@@ -3,6 +3,7 @@ package expr
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/PaesslerAG/gval"
@@ -17,6 +18,7 @@ func ArrayFunctions() []gval.Language {
 		gval.Function("has", has),
 		gval.Function("hasAll", hasAll),
 		gval.Function("find", find),
+		gval.Function("sort", sortSlice),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


darh (approved):

pkg/expr/func_arr.go
```diff
@@ -220,3 +222,59 @@ func slice(arr interface{}, start, end int) interface{} {
...
+// sortSlice sorts slice
+func sortSlice(arr interface{}, desc bool) (out interface{}, err error) {
+	if arr, err = toSlice(arr); err != nil {
+		return
+	}
+
+	// sort slice of typed values
+	if stv, is := arr.([]TypedValue); is {
+		return sortSliceStable(stv, func(i, j int) bool {
+			c, _ := stv[i].(Comparable).Compare(stv[j])
```
tjerman: This expression scares me; a typed value does not necessarily implement the `Comparable` interface so this line will panic because you are not "respecting" that fact (for lack of a better word).

To avoid a panic you need to do something like this:

```go
casted, ok := thing.(Comparable)
if !ok {
    // Does not conform; don't try to compare it
}
```

(omitting the `ok` panics if it doesn't conform).

tjerman (changes requested):
Alongside the comment; a few things to consider:

1. should the dedicated logic that handles comparable typed values be moved down to where you handle primitive ones? To me, it would make sense to have everything in one place -- up to you.
2. what should happen if a value is not comparable; primitive (ints and strings) or typed (Records and stuff)? Should we error out or treat them as the same? In my opinion I would prefer an error so that we don't have this "strange undefined" state; probably something to discuss with Denis but from my point of view; up to you.

tjerman (changes requested):
The error management is a little off and won't work correctly for some scenarios (see comments down below).
Other then that, lgtm

pkg/expr/func_arr.go
```diff
@@ -220,3 +222,81 @@ func slice(arr interface{}, start, end int) interface{} {
...
+
+	// sort slice of native type values
+	var (
+		v  = reflect.ValueOf(arr)
+		vi = v.Interface()
+	)
+
+	// sort the array if error is nil;
+	// we trap the error(s) in the outer scope from the less function,
+	// and returning the last error at a time since all error will the same in almost every scenario.
```
tjerman, line 243: The error management is a bit off --

The `err` variable used in `c, err = casted.Compare(stv[j])`, `err = fmt.Errorf("cannot compare %s and %s: unknown state",...)`, and `err = fmt.Errorf("cannot compare %s and %s: unknown state", getKind(nVal), getKind(mVal))` is defined in the outer scope and is the same.

With that said, if (for example) `err = fmt.Errorf("cannot compare %s and %s: unknown state", getKind(nVal), getKind(mVal))` sets the error, the upper line (`c, err = casted.Compare(stv[j])`) can clear the error and make it seem like nothing happened.

What I would do is this:

```go
out, sortErr := sortSliceStable(vi, func(i, j int) bool {
	if err != nil {
		return false
	}

	var (
		nVal = v.Index(i)
		mVal = v.Index(j)
	)
	if desc {
		nVal = v.Index(j)
		mVal = v.Index(i)
	}

	switch getKind(vi) {
	// ...
	// ...
	// ...
	default:
		// sort slice of typed values
		if stv, is := arr.([]TypedValue); is {
			if casted, ok := stv[i].(Comparable); ok {
				var c int
				c, err = casted.Compare(stv[j])
				if err != nil {
					return false
				}
				// ...
			} else {
				err = fmt.Errorf("cannot compare %s and %s: unknown state", stv[i].Type(), stv[j].Type())
				return false
			}
		}
		err = fmt.Errorf("cannot compare %s and %s: unknown state", getKind(nVal), getKind(mVal))
		return false
	}
})
```

The above snippet adds a little guard statement to the very top to skip the entire processing in case something went wrong.

pkg/expr/func_arr.go
```diff
@@ -220,3 +222,81 @@ func slice(arr interface{}, start, end int) interface{} {
...
+					err = fmt.Errorf("cannot compare %s and %s: unknown state", stv[i].Type(), stv[j].Type())
+					return false
+				}
+			}
+			err = fmt.Errorf("cannot compare %s and %s: unknown state", getKind(nVal), getKind(mVal))
+			return false
+		}
+	})
+	if sortErr != nil {
+		return arr, err
```
tjerman: This should probably return `sortErr` and not `err`?

pkg/expr/func_arr.go
```diff
@@ -220,3 +222,81 @@ func slice(arr interface{}, start, end int) interface{} {
...
+// sortStable sorts the slice x using the provided less
+// function, keeping equal elements in their original order.
+func sortSliceStable(arr interface{}, less func(i, j int) bool) (out interface{}, err error) {
+	if arr, err = toSlice(arr); err != nil {
+		return
+	}
+
+	sort.SliceStable(arr, less)
+	return arr, err
+}
```
tjerman: I would personally remove this function and merge the contents to the caller as there is only one call; up to you

tjerman (changes requested):
The error management is a little off and won't work correctly for some scenarios (see comments down below).
Other then that, lgtm

PR #315, opened 2022-01-04
Allow version in DB_DSN for dev mode

vicpatel:
If server is executed through Makefile we set BUILD_VERSION in .env and use it to replace {version} in DB_DSN.

This would help developers to automatically use different db when switching between versions.

Please take a look.

### DIFF
--- a/store/connect.go
+++ b/store/connect.go
@@ -3,6 +3,9 @@ package store
 import (
 	"context"
 	"fmt"
+	"go.uber.org/zap"
+	"os"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -108,6 +108,7 @@ $(RELEASE_PKEY):
 
 watch: $(GIN)
 	$(GIN) $(GIN_ARGS) --build cmd/corteza run -- serve
+	@echo BUILD_VERSION=$(BUILD_VERSION) >> .env
```
darh: this looks intrusive... let's not append to .env AFTER (!!) every watch... :P

store/connect.go
```diff
@@ -14,7 +16,13 @@ var (
 	registered = make(map[string]ConnectorFn)
 )
 
-func Connect(ctx context.Context, dsn string) (s Storer, err error) {
+func Connect(ctx context.Context, dsn string, opt options.EnvironmentOpt) (s Storer, err error) {
+	if opt.IsDevelopment() {
+		expr := regexp.MustCompile(`[.\-]+`)
+		version := expr.ReplaceAllString(opt.BuildVersion, "_")
```
darh: how about if we just do a quick & dirty `os.Getenv('BUILD_VERSION')` instead of all this option acrobatics?

darh (changes requested):
kiss ;)

pkg/options/environment.gen.go
```diff
@@ -10,7 +10,8 @@ package options
 
 type (
 	EnvironmentOpt struct {
-		Environment string `env:"ENVIRONMENT"`
+		Environment  string `env:"ENVIRONMENT"`
+		BuildVersion string `env:"BUILD_VERSION"`
```
darh: re-run codegen

store/mysql/mysql.go
```diff
@@ -46,6 +46,8 @@ func Connect(ctx context.Context, dsn string) (store.Storer, error) {
 	cfg.CastModuleFieldToColumnType = fieldToColumnTypeCaster
 	cfg.SqlSortHandler = SqlSortHandler
 
+	fmt.Println("DB_DSN: ", cfg.DataSourceName)
```
darh: leftovers?

darh (changes requested):

Makefile
```diff
@@ -108,6 +108,7 @@ $(RELEASE_PKEY):
 
 watch: $(GIN)
 	$(GIN) $(GIN_ARGS) --build cmd/corteza run -- serve
+	@echo BUILD_VERSION=$(BUILD_VERSION)
```
darh: missed this one?

darh (changes requested):
One thing in Makefile that looks like another leftover?

Before you squash & merge:
 - copy (adjusted) the comment from .env.example to `Connect()` fn
 - add one or two examples in the comment how versions are transformed
 - maybe additional warning that it's not bulletproof and might break your connection.

darh (approved):

PR #317, opened 2022-01-10
Added support for SAML HTTP binding

petergrlica:
What:
 - added a setting for AuthNRequest

### DIFF
--- a/auth/saml/sp.go
+++ b/auth/saml/sp.go
@@ -40,6 +40,8 @@ type (
 		SignRequests    bool
 		SignatureMethod string
 
+		Binding string
+
 		// user meta from idp
 		IdentityPayload IdpIdentityPayload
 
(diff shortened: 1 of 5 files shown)


auth/saml/sp.go
```diff
@@ -72,6 +74,13 @@ func NewSamlSPService(log *zap.Logger, args SamlSPArgs) (s *SamlSPService, err e
 		sp.SignatureMethod = args.SignatureMethod
 	}
 
+	// default to GET
+	if args.Binding == "" {
+		args.Binding = saml.HTTPRedirectBinding
+	}
+
+	// sp.AuthnNameIDFormat = "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
```
darh: leftovers?

darh (commented):
lgtm

auth/saml/sp.go
petergrlica: A reminder, but will remove

PR #320, opened 2022-02-02
Update argument type of set expr function

vicpatel:
It changes 3rd argument type os set expr function from TypedValue to interface, since GvalParser parses argument to native types before evaluating the expression.

Please take a look.

### DIFF
--- a/tests/workflows/testdata/S0016_set_expression_issue/workflow.yaml
+++ b/tests/workflows/testdata/S0016_set_expression_issue/workflow.yaml
@@ -1,0 +1,28 @@
+workflows:
+  set_expression:
+    enabled: true
+    trace: true
+    triggers:
+      - enabled: true
+        stepID: 1
+
+    steps:
+      - stepID: 1
+        kind: expressions
+        arguments: [
+          { target: aux, type: Vars, value: null },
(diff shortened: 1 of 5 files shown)


tests/workflows/testdata/S0016_set_expression_issue/workflow.yaml
```diff
@@ -0,0 +1,28 @@
...
+        kind: expressions
+        arguments: [
+          { target: aux, type: Vars, value: null },
+          { target: testString, type: String, value: "testString" },
+          { target: foo, type: String, value: "testing string" },
+          { target: testInt, type: String, value: "testInt" },
+          { target: bar, type: Integer, expr: "40" },
+          { target: testVar, type: String, value: "testVar" },
+          { target: foobar, type: Vars, expr: "{\"testString\": \"testing string\", \"testFloat\": 50}" },
+          { target: out, type: Vars, expr: "set(aux, testString, foo)" },
```
tjerman, line 20: Can we have one more `set` call with all three params being constant?
Something like:

```
set({}, "something", "here")
```

tjerman (commented):
If I recall correctly we were supposed to support these assignment functions for records and record values also.
I can't see from the code if that is also now working.

PR #323, opened 2022-02-05
2022.3.x auth limiting

tjerman:
* Reject request in case new tokens were issues but old ones were used
* Reject request if exceeds rate limit

All of the above is done from the existing state so no DB modifications are needed.

### DIFF
--- a/system/service/auth.go
+++ b/system/service/auth.go
@@ -64,10 +64,17 @@ const (
 
 	passwordMinLength = 8
 	passwordMaxLength = 256
+
+	tokenReqMaxCount  = 5
+	tokenReqMaxWindow = time.Minute * 15
 )
 
 var (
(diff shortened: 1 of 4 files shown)


darh (approved):

system/service/auth.go
```diff
@@ -1024,6 +1024,19 @@ func (svc auth) loadUserFromToken(ctx context.Context, token, kind string) (u *t
...
+	// Check if they used their last issued token of the given kind
+	cc, _, err := store.SearchCredentials(ctx, svc.store, types.CredentialsFilter{
+		OwnerID: c.OwnerID,
+		Kind:    c.Kind,
+		Deleted: filter.StateExcluded,
+	})
+	if err != nil {
+		return
+	}
+	if c.Credentials != cc[len(cc)-1].Credentials {
```
darh: This is a bit brave -- we're not explicitly sorting this..

tjerman (commented):
LGTM :ship:

PR #324, opened 2022-02-15
Apigw request body read + refactor

petergrlica:
What:
 - refactored jsenv
 - added `pkg/http.Request` with body buffering capabilities
 - fixed workflow loading on processer
 - refactored expr.HttpRequest out of `system/` into `expr/`


automation/automation/apigw_body_handler.go
```diff
@@ -0,0 +1,41 @@
...
+	return h
+}
+
+func (h apigwBodyHandler) read(ctx context.Context, args *apigwBodyReadArgs) (res *apigwBodyReadResults, err error) {
+	res = &apigwBodyReadResults{}
+
+	if !args.hasRequest {
+		err = fmt.Errorf("could not read body, contents missing")
+		return
+	}
```
tjerman: Is this needed if the param is marked as required?

pkg/apigw/registry/registry.go
```diff
@@ -74,7 +72,5 @@ func (r *Registry) Preload() {
 }
 
 func NewWorkflow() (wf filter.WfExecer) {
-	// implementation assumes that Corredor & Workflow options can not be changed
-	// in the run-time.
-	return service.Workflow(logger.Default().Named("workflow"), *options.Corredor(), *options.Workflow())
+	return service.DefaultWorkflow
```
tjerman, line 75: Should the system boot procedure push this into here?

pkg/expr/expr_types.go
```diff
@@ -654,6 +655,36 @@ func CastToReader(val interface{}) (out io.Reader, err error) {
...
+func CastToUrl(val interface{}) (out *url.URL, err error) {
+	switch val := UntypedValue(val).(type) {
+	case *url.URL:
+		return val, nil
+	case nil:
+		return &url.URL{}, nil
+	default:
+		return &url.URL{}, fmt.Errorf("unable to cast type %T to %T", val, out)
+	}
+}
```
tjerman: Should `strings` and `[]byte` also be handled here?

pkg/http/request.go
```diff
@@ -0,0 +1,91 @@
...
+		Body io.Reader
+	}
+
+	BufferedReader struct {
+		buffer []byte
+	}
+)
+
+// NewBufferedReader creates a new reader from readcloser
+func NewBufferedReader(r io.ReadCloser) (b *BufferedReader, err error) {
```
tjerman: You can use a `io.Reader` since you're not closing

pkg/http/request.go
```diff
@@ -0,0 +1,91 @@
...
+				break
+			}
+
+			p[n] = bb.buffer[n]
+			n++
+		}
+	}
+
+	return
+}
```
tjerman: Won't this read from the start every time you call this `Read`?
This is something we do want but it's not what `Read` generally does so it could cause issues.
Should we conform to `ReadSeeker` instead?

pkg/http/request_test.go
```diff
@@ -0,0 +1,40 @@
...
+func Test_requestReadMultiple(t *testing.T) {
+	var req = require.New(t)
+	r, _ := h.NewRequest("POST", "/foo", strings.NewReader(`foo body`))
+
+	rs, err := NewBufferedReader(r.Body)
+
+	req.NoError(err)
+	req.Equal(`foo body`, must(io.ReadAll(rs)))
+	req.Equal(`foo body`, must(io.ReadAll(rs)))
+}
```
tjerman: Can we have a test case with a larger payload?
I'm looking at `io.ReadAll` implementation and it reads in `512 byte` chunks; so if your payload exceeds that (and I do not miss interpret the implementation) it should start breaking.

automation/automation/expr_types.go
```diff
@@ -29,3 +32,25 @@ func CastToEmailMessage(val interface{}) (out *emailMessage, err error) {
...
+	var (
+		b   []byte
+		err error
+	)
+
+	switch val := in.(type) {
+	case *http.Request:
+		b, err = io.ReadAll(val.Body)
+	case io.Reader:
+		b, err = io.ReadAll(val)
```
tjerman, line 46: If `val` would be one of the expanded IO interfaces (`io.ReadCloser`, `io.ReadCloseSeeker`, ...) would this case still catch them (I'm not entirely sure)?
Can those even happen? I would personally handle them in a better safe than sorry kind of thing.

pkg/http/request.go
```diff
@@ -0,0 +1,91 @@
...
+		for n < c {
+			if len(bb.buffer) <= n {
+				err = io.EOF
+				break
+			}
+
+			p[n] = bb.buffer[n]
+			n++
+		}
+	}
```
tjerman: Could this be implemented with the `copy` function?

pkg/http/request_test.go
petergrlica: Nice catch, will check how to implement Seeker, it does look like the Read limitation will cause trouble.

pkg/http/request.go
petergrlica: Responded below, will check Seeker

automation/automation/expr_types.go
petergrlica: `io.ReadCloser` is a `io.Reader` and a `io.Closer`, so it would catch that

pkg/http/request.go
petergrlica: Will check how Seeker would be implemented here

pkg/http/request.go
petergrlica: Wanted to limit it to `io.ReadCloser` as it will be used only for `http.Request` (for now)

pkg/expr/expr_types.go
petergrlica: Yes

pkg/apigw/registry/registry.go
petergrlica: I don't see the benefit here, used the DefaultWorkflow as it handles the reloads (and wf caching) properly

automation/automation/apigw_body_handler.go
petergrlica: Sure, workflow probably handles that, will remove

darh (approved):

tjerman (approved):
Now, this is what gets me excited :ok_hand: 

The only thing I'm not too sure about is that the Reader automatically seeks to the start.
I would personally prefer to see the read and seek as two separate steps, or wrapped in some helper function.

Personally, I think I would change that, but this is also fine by me; ship it :+1:

PR #325, opened 2022-02-16
Extend set/omit expr function

vicpatel:
It updates 1st parameter of set expr function from merger type to interface, so It will accept constant values, also extend set/omit usage for ComposeRecordValues.

Please take a look.

### DIFF
--- a/compose/automation/expr_types.go
+++ b/compose/automation/expr_types.go
@@ -518,3 +518,105 @@ func CastToAttachment(val interface{}) (out *types.Attachment, err error) {
 		return nil, fmt.Errorf("unable to cast type %T to %T", val, out)
 	}
 }
+
+func EmptyComposeRecordValues() *ComposeRecordValues {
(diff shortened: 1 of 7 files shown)


darh:
Looks ok. @tjerman does this solve your issues?

compose/automation/expr_types.go
```diff
@@ -518,3 +518,104 @@ func CastToAttachment(val interface{}) (out *types.Attachment, err error) {
...
+
+			bb, err = json.Marshal(v.Get())
+			if err != nil {
+				return err
+			}
+
+			err = json.Unmarshal(bb, &rVal)
+			if err != nil {
+				return err
+			}
```
tjerman, line 570: I'm not entirely sure how I feel about this, perhaps add a `@todo` if we don't have a better fix atm?

tjerman (commented):
This is what I wanted, yes; see my comment and then call it a day :+1:

PR #326, opened 2022-02-16
Add user import/export

tjerman:


app/boot_levels.go
```diff
@@ -507,6 +508,10 @@ func (app *CortezaApp) InitServices(ctx context.Context) (err error) {
 		}
 	}
 
+	// Register reporters
+	// @todo additional datasource providers; generate?
+	sysService.DefaultReport.RegisterReporter("composeRecords", cmpService.DefaultRecord)
```
darh, line 513: Not sure why is this here?

darh (commented):
how hard it would be to add support for uncompressed import?

app/boot_levels.go
tjerman: I cherry-picked a few bits from the gig branch; this will all get `--ff` to the version branch so I just left it here for the time being

PR #327, opened 2022-02-17
2022.3.x feature web console

darh:
Please take a look.


.env.example
```diff
@@ -170,6 +170,26 @@
...
+# Default: admin
+# HTTP_SERVER_WEB_CONSOLE_USERNAME=admin
+
+###############################################################################
+# Password for the web console endpoint. When running in dev environment, password is not required.
+# 
+# Corteza intentionally sets default password to random chars to prevent security incidents.
+# Type:    string
+# Default: <no value>
+# HTTP_SERVER_WEB_CONSOLE_PASSWORD=<no value>
```
tjerman, line 191: Would it make sense to allow multiple credentials here?
Perhaps something in the lines of:

```
HTTP_SERVER_WEB_CONSOLE_CREDENTIALS="cup:cake pine:apple chocolate:cookies"
```

This way you could revoke access to specific users while preserving others.

pkg/api/server/handlers.go
```diff
@@ -110,14 +117,40 @@ func activeRoutes(log *zap.Logger, mountable []func(r chi.Router), envOpt option
...
 
-	mountServiceHandlers(r, log, httpOpt)
+	mountServiceHandlers(r, log, httpOpt, active)
 	return
 }
 
-func mountServiceHandlers(r chi.Router, log *zap.Logger, opt options.HttpServerOpt) {
+func mountServiceHandlers(r chi.Router, log *zap.Logger, opt options.HttpServerOpt, state uint32) {
+	if opt.WebConsoleEnabled {
+		path := "/!console"
```
tjerman, line 128: Should we use `/__console` to be consistent with other `/__profiler` and similar routes?
Is the `!` something special here?
Do we even want to prefix it; is `/console` not enough?

pkg/options/HTTPServer.go
```diff
@@ -1,5 +1,15 @@
 package options
 
+func (o *HttpServerOpt) Defaults() {
+	if Environment().IsDevelopment() {
+		// enable web console and remove username, password defaults
+		// if this is explicitly via ENV, it will override these defaults
+		o.WebConsoleEnabled = true
+		o.WebConsoleUsername = ""
+		o.WebConsolePassword = ""
+	}
```
tjerman, line 10: Above, in case there is no username/password the console is still mounted with a logged warning.
I would prefer to see it is either disabled by default or mounted with a random username/password which would be logged in the console.

I can see cases where someone is going to upgrade to 2022.3 and not set these parameters.

tjerman (commented):
Should we move the web console to a different repository and then embed the built thing?
It might be easier for our front-end people to work on it.

.env.example
darh: Good idea, maybe later.

pkg/api/server/handlers.go
darh: Would be good if I would add some explanation there, yes. 

The double lo-dash endpoints always felt a bit off and would be moved under console and add a nice UI around it
Why exclamation mark or a prefix at all? Just to mark a special meaning for the route.

Does that makes sense?

pkg/options/HTTPServer.go
darh: As explained in the docs:
> Console is enabled and open to everyone by default in development environment (`ENVIRONMENT=dev`) or can be explicitly enabled by setting `HTTP_SERVER_WEB_CONSOLE_ENABLED` environmental variable to `true`. Default values for username is "admin" and random characters for password to avoid potential security issues.

So -- this is DISABLED by default and is only enabled, and open to everyone if you are running public server in dev mode. Not much we can help you there.

pkg/api/server/handlers.go
tjerman: I suppose, but I would still consider having it as `/console` just like we have for `/api` and `/api-gw`

pkg/options/HTTPServer.go
tjerman: I suppose I should have read the docs :see_no_evil:

vicpatel (approved):

pkg/api/server/handlers.go
vicpatel: I would say keeping all routes consistent is a good idea but wouldn't it be better for the future, If we introduce some more routes like this then any route that starts with `!` will have a special use case.

darh: /console it is.

PR #331, opened 2022-02-23
Prevent workflow worker from stopping in case of delayed steps

tjerman:
Backport to 2021.9

In such case
Screenshot from 2022-02-23 14-24-55
where there were two parallel branches; if the two branches used some delay/prompt step, the first terminated the whole thing without waiting for the second scheduled step.

This occurred due to us not checking for delayed/prompted steps


darh (commented):
would love to see some tests, but I guess we'll see soon enough :)

PR #333, opened 2022-03-01
Improve KV related expr types, add r/w locking

vicpatel:
Also extended filter for compose record values

Please take a look.


pkg/expr/expr_types.go
```diff
@@ -463,12 +463,12 @@ func CastToUnsignedInteger(val interface{}) (out uint64, err error) {
 }
 
 func (t *KV) Has(k string) bool {
-	_, has := t.value[k]
-	return has
+	_, is := t.value[k]
```
darh, line 466: why change this?

pkg/expr/expr_types.go
```diff
@@ -463,12 +463,12 @@ func CastToUnsignedInteger(val interface{}) (out uint64, err error) {
...
+	return is
 }
 
 func (t *KV) Select(k string) (TypedValue, error) {
-	if v, has := t.value[k]; has {
+	if v, is := t.value[k]; is {
```
darh, line 471: why change this?

darh (changes requested):

pkg/expr/expr_types.go
vicpatel: Reverting this change, can't recall the reason but must be changed to keep all bool receivers same throughout the file 🙈

vicpatel:
Merged into `2022.9`

PR #338, opened 2022-03-19
Solve the "in" sql syntax error

danniel2016:
Solve the "in" syntax error


darh:
Thanks for this but it judging from the branches used you are fixing an old version and 2022.3 is just around the corner. On top of that we are replacing `squirrel`. 

I will however make sure that this bug does not creep back in.

PR #339, opened 2022-03-21
Integration gateway profiler

petergrlica:
**What the feature does**

Adds a service and apigw subsystem that carries info on:
 - missed hits on registered routes
 - missed hits on non-existing routes
 - aggregations + hit details per-route, per-hit

**Please check**
Is the service and aggregation on the right layer (service)?
Can sorting be optimized?

Note, there is no hit purging yet, so the performance implications on that area are known.


system/types/app_settings.go
```diff
@@ -207,6 +207,12 @@ type (
 			Enabled bool `kv:"-" json:"enabled"`
 		} `kv:"federation" json:"federation"`
 
+		// Federation settings
```
darh: leftover.

pkg/apigw/registry/registry.go
```diff
@@ -8,19 +8,22 @@ import (
...
-		h map[string]types.Handler
+		opts *options.ApigwOpt
+		h    map[string]types.Handler
 	}
 
 	secureStorageTodo struct{}
 )
 
-func NewRegistry() *Registry {
+func NewRegistry(opts *options.ApigwOpt) *Registry {
```
darh: Options should b (unwritten Corteza rule) passed to the functions as values (not refs).

pkg/apigw/helpers.go
```diff
@@ -21,3 +25,35 @@ func helperDefaultResponse(opt *options.ApigwOpt) http.HandlerFunc {
...
 	}
 }
+
+func helperMethodNotAllowed(opt *options.ApigwOpt, pr *profiler.Profiler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		addToProfiler(opt, pr, r, http.StatusMethodNotAllowed)
+
+		if opt.LogEnabled {
+			// Say something friendly when logging is enabled
+			http.Error(w, devHelperResponseBody, http.StatusTeapot)
```
darh, line 37: This is nice :)
But why not properly returning 405 even when logging? The fact that logging is enabled does not affect (other) behaviour of the handler and its responses , correct?

pkg/apigw/helpers.go
```diff
@@ -21,3 +25,35 @@ func helperDefaultResponse(opt *options.ApigwOpt) http.HandlerFunc {
...
+func addToProfiler(opt *options.ApigwOpt, pr *profiler.Profiler, r *http.Request, status int) {
+	if !(opt.ProfilerEnabled && opt.ProfilerGlobal) {
+		return
+	}
+
+	// add to profiler
+	ar, err := h.NewRequest(r)
+
+	if err != nil {
+		panic(err)
```
darh: Can this situation be caused by a non-developer (so anyone w/o access to server source)?
If yes, then this should be converted to a log (zap).

pkg/apigw/profiler/hit.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	h "github.com/cortezaproject/corteza-server/pkg/http"
+)
+
+type (
+	Hit struct {
```
darh, line 12: what was the rationale for abbreviating fields in this struct?

app/options/apigw.cue
```diff
@@ -25,6 +25,14 @@ apigw: schema.#optionsGroup & {
 			type:        "bool"
 			description: "Enable extra logging"
 		}
+		profiler_enabled: {
+			type:        "bool"
+			description: "Enable profiler"
+		}
+		profiler_global: {
+			type:        "bool"
+			description: "Profiler enabled for all routes"
```
darh: do we need to explain in more detail gow enabling this affects profilers on routes?

app/options/apigw.cue
```diff
@@ -25,6 +25,14 @@ apigw: schema.#optionsGroup & {
 			type:        "bool"
 			description: "Enable extra logging"
 		}
+		profiler_enabled: {
```
darh, line 28: should this be enabled by default?

pkg/apigw/ctx/ctx.go
```diff
@@ -3,12 +3,14 @@ package ctx
...
 	"context"
 
+	"github.com/cortezaproject/corteza-server/pkg/apigw/profiler"
 	"github.com/cortezaproject/corteza-server/pkg/apigw/types"
 )
 
 type ContextKey string
 
 const ContextKeyScope ContextKey = "scope"
+const ContextKeyProfiler ContextKey = "profiler"
```
darh: can you change this (both?) to follow "empty-struct" approach (see `identityCtxKey`) we use for context key?

app/options/apigw.cue
darh: and, should we log this into stdout (warn/info/debug) that global profiler is enabled?

darh (changes requested):

pkg/apigw/ctx/ctx.go
petergrlica: :+1: agree

app/options/apigw.cue
petergrlica: > should this be enabled by default?

It should be on by default and global off, will act accordingly

app/options/apigw.cue
petergrlica: > do we need to explain in more detail gow enabling this affects profilers on routes?

I'll add a little more info on server logs, most will be in docs

> and, should we log this into stdout (warn/info/debug) that global profiler is enabled?

yes

pkg/apigw/profiler/hit.go
petergrlica: None other than self-explanatory

pkg/apigw/helpers.go
petergrlica: Did not want to wrap with logger and the rationale was that something really wrong should go when creating a wrapper request from `http.Request`, but there could be some issues with ie headers or body, so I'll revert that panic, :+1:

pkg/apigw/helpers.go
petergrlica: I just wanted a teapot, I agree with you though

pkg/apigw/registry/registry.go
petergrlica: :+1:

PR #341, opened 2022-03-21
Fixes record field validation translations

vicpatel:
Please take a look.

Related Locale PR# https://github.com/cortezaproject/corteza-locale/pull/39/files

### DIFF
--- a/compose/service/values/validator.go
+++ b/compose/service/values/validator.go
@@ -3,6 +3,7 @@ package values
 import (
 	"context"
 	"fmt"
+	"github.com/cortezaproject/corteza-server/pkg/locale"
 	"math/big"
 	"net/mail"
 	"net/url"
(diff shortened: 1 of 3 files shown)


compose/service/values/validator.go
```diff
@@ -26,38 +27,72 @@ type (
...
+	}
 }
 
-func makeEmptyErr(field *types.ModuleField) types.RecordValueError {
-	return types.RecordValueError{Kind: "empty", Meta: map[string]interface{}{"field": field.Name}}
+func makeEmptyErr(ctx context.Context, field *types.ModuleField, ls localeService) types.RecordValueError {
+	return types.RecordValueError{
+		Kind:    "empty",
+		Meta:    map[string]interface{}{"field": field.Name},
+		Message: ls.T(ctx, ComposeResourceTranslationType, "record-field.errors.empty"),
```
darh: Value you used for translation namespace (`ComposeResourceTranslationType`) does not look right. Isn't this used for resource translations only? And you are not translating resources here (field labels on module for example) but static texts. 

See `RecordErrValueInput` and how translations are constructed there. Namespace is `compose`. See `src/en/corteza-server/compose/record.yaml` in `corteza-locale`.

darh (changes requested):

compose/service/values/validator.go
vicpatel: MB, I will fix the variable name, I introduced new var `ComposeResourceTranslationType = "compose"`.

PR #342, opened 2022-03-23
2022.3.x wftweaks

tjerman:
Please take a look.


app/options/workflow.cue
```diff
@@ -21,6 +21,11 @@ workflow: schema.#optionsGroup & {
 			defaultGoExpr: "16"
 			description:   "Defines the maximum call stack size between workflows"
 		}
+		stack_trace_disabled: {
```
darh: I suggest we flip this to enabled and follow the established pattern.

darh (changes requested):
Just one thing: DISABLED => ENABLED.

PR #347, opened 2022-04-08
Strftime on expressions input string

petergrlica:
Receive string as an input parameter to the strftime expression function also.

### DIFF
--- a/pkg/expr/func_time.go
+++ b/pkg/expr/func_time.go
@@ -173,6 +173,14 @@ func prepMod(base interface{}, mod interface{}) (*time.Time, int, error) {
 		t = &auxt
 	case *time.Time:
 		t = auxt
+	case string:
+		tt, err := cast.ToTimeE(auxt)
+
+		if err != nil {
+			return nil, 0, err
+		}
+
+		t = &tt
 	default:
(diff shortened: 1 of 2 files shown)


pkg/expr/func_time.go
```diff
@@ -173,6 +173,14 @@ func prepMod(base interface{}, mod interface{}) (*time.Time, int, error) {
 		t = &auxt
 	case *time.Time:
 		t = auxt
+	case string:
+		tt, err := cast.ToTimeE(auxt)
```
tjerman: Should we use the same functions we do for record values and string-date conversion?

petergrlica: These functions, we use all over the `expr` package, so I wanted to be consistent.
Can you provide an example of the conversions you mention?

tjerman: https://github.com/cortezaproject/corteza-server/blob/2022.9.x-crs/compose/service/values/sanitizer.go#L193

petergrlica: @tjerman I would like to leave the `compose/service` methods out of `pkg` area so as not to pollute (all that we include in this package is external libs / `pkg/*` as I think should be the case)

PR #348, opened 2022-04-11
Add txRetryErrHandler to postgres driver

vicpatel:
Fixes panic on server startup with Postgres DB.

Please take a look.


darh:
fixed in a different commit.

PR #353, opened 2022-04-19
Fix mimeType detection for CSV file

vicpatel:
Please take a look.

### DIFF
--- a/tests/compose/record_test.go
+++ b/tests/compose/record_test.go
@@ -796,11 +796,21 @@ func TestRecordAttachment(t *testing.T) {
 			},
 		},
 		&types.ModuleField{Name: "str", Kind: "String"},
+		&types.ModuleField{
+			Name: "csv_only",
+			Kind: "File",
+			Options: types.ModuleFieldOptions{
+				"mimetypes": "text/csv",
+			},
+		},
 	)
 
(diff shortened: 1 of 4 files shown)


tests/compose/record_test.go
```diff
@@ -796,23 +796,31 @@ func TestRecordAttachment(t *testing.T) {
...
 	defer func() {
 		// reset settings after we're done
 		systemService.CurrentSettings.Compose.Record.Attachments.MaxSize = 0
 		systemService.CurrentSettings.Compose.Record.Attachments.Mimetypes = nil
 	}()
 
 	systemService.CurrentSettings.Compose.Record.Attachments.MaxSize = maxSizeLimit
-	systemService.CurrentSettings.Compose.Record.Attachments.Mimetypes = []string{
-		"application/octet-stream",
-	}
```
vicpatel, line 815: Removing this; No need to add fallback allowed mime-type after updating [mime-type detection func](https://github.com/cortezaproject/corteza-server/pull/353/files#diff-50c3677675e238dcb0bcf922e2a0bb629c232c27666ec4794c199627a873f26bR513), It extracts file mime-type more accurately now.

tjerman (approved):

PR #354, opened 2022-04-19
Track compose record changes

vicpatel:
Add env options to enable/disable Compose record action log  

It will enable/disable action log entry for compose record create, update or delete.

Please take a look.


darh (changes requested):
address `@todo`s

darh (approved):

PR #355, opened 2022-04-26
 Charts yaxis translations

vicpatel:
Please take a look.

### DIFF
--- a/compose/service/chart.go
+++ b/compose/service/chart.go
@@ -2,7 +2,9 @@ package service
 
 import (
 	"context"
+	"github.com/cortezaproject/corteza-server/pkg/locale"
 	"reflect"
+	"strconv"
 
 	"github.com/cortezaproject/corteza-server/compose/types"
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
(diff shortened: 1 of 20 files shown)


compose/service/chart.go
```diff
@@ -249,6 +270,10 @@ func (svc chart) updater(ctx context.Context, namespaceID, chartID uint64, actio
 			}
 		}
 
+		if err = updateTranslations(ctx, svc.ac, svc.locale, c.EncodeTranslations()...); err != nil {
```
tjerman, line 278: Updater should assign a reportID in case it is not provided; same as with create but in an if that checks if it should

tjerman (commented):
Rest seems fine

compose/service/chart.go
```diff
@@ -172,7 +172,7 @@ func (svc chart) Create(ctx context.Context, new *types.Chart) (*types.Chart, er
 
 		// Ensure page-block IDs
 		for i := range new.Config.Reports {
-			new.Config.Reports[i].ReportID = uint64(i) + 1
+			new.Config.Reports[i].ReportID = nextID() + uint64(i)
 		}
```
tjerman: If you're going to do it like this, just use nextID without the increment

pkg/provision/migrations_202203_resource_translations.go
```diff
@@ -86,8 +87,8 @@ func convertComposeChartReportsTranslations(res *cmpTypes.Chart) (sysTypes.Resou
 	var tt sysTypes.ResourceTranslationSet
 
 	for i, r := range res.Config.Reports {
-		reportID := i + 1
-		res.Config.Reports[i].ReportID = uint64(reportID)
+		reportID := id.Next() + uint64(i)
```
tjerman: same here

tjerman (commented):
Tiny request; else is fine :+1:

PR #360, opened 2022-05-17
2022.9.x feature privacy console

vicpatel:
Add data privacy request route and implementation

- Introduces new role for data-privacy-officer(Role allows user to manage data privacy requests)
- along with a new system resource for data privacy requests and its access control
- Routes as per access control to create data privacy request, list request and filter it based on their kind and status, and update data privacy request status

Please take a look.

### DIFF
--- a/compose/service/locale.go
+++ b/compose/service/locale.go
@@ -272,6 +272,45 @@ func (svc resourceTranslationsManager) pageExtended(ctx context.Context, res *ty
 	return
 }
 
+func (svc resourceTranslationsManager) chartExtended(_ context.Context, res *types.Chart) (out locale.ResourceTranslationSet, err error) {
+	var (
+		yAxisLabelK  = types.LocaleKeyChartYAxisLabel
+		metricLabelK = types.LocaleKeyChartMetricsMetricIDLabel
+	)
+
+	for _, report := range res.Config.Reports {
+		for _, tag := range svc.locale.Tags() {
+			out = append(out, &locale.ResourceTranslation{
+				Resource: res.ResourceTranslation(),
+				Lang:     tag.String(),
+				Key:      yAxisLabelK.Path,
+				Msg:      svc.locale.TResourceFor(tag, res.ResourceTranslation(), yAxisLabelK.Path),
+			})
+		}
+
+		for _, metric := range report.Metrics {
+			if _, ok := metric["metricID"]; ok {
+				mID, is := metric["metricID"].(string)
+				if !is {
+					continue
+				}
+				mpl := strings.NewReplacer("{{metricID}}", mID)
+
+				for _, tag := range svc.locale.Tags() {
+					out = append(out, &locale.ResourceTranslation{
+						Resource: res.ResourceTranslation(),
+						Lang:     tag.String(),
+						Key:      mpl.Replace(metricLabelK.Path),
+						Msg:      svc.locale.TResourceFor(tag, res.ResourceTranslation(), mpl.Replace(metricLabelK.Path)),
+					})
+				}
+			}
+		}
+	}
+
+	return
+}
+
 func (svc resourceTranslationsManager) pageExtendedAutomationBlock(tag language.Tag, res *types.Page, block types.PageBlock, blockID uint64) (locale.ResourceTranslationSet, error) {
 	var (
 		k     = types.LocaleKeyPagePageBlockBlockIDButtonButtonIDLabel
--- a/pkg/provision/migrations_202203_resource_translations.go
+++ b/pkg/provision/migrations_202203_resource_translations.go
@@ -1,0 +1,116 @@
+package provision
+
+import (
+	"context"
+	"fmt"
+	cmpTypes "github.com/cortezaproject/corteza-server/compose/types"
+	"github.com/cortezaproject/corteza-server/pkg/filter"
+	"github.com/cortezaproject/corteza-server/pkg/id"
+	"github.com/cortezaproject/corteza-server/store"
+	sysTypes "github.com/cortezaproject/corteza-server/system/types"
+	"go.uber.org/zap"
+	"strings"
+)
+
+// Migrates resource translations from the resource
+// struct to the dedicated store (table)
+//
+// While doing this, we also modify some resource substructure:
+//  - chart reports (assign report IDs)
+//
+// Note: we will migrate all translations to current default language
+// If you do not like that, shut down Corteza after migrations and fix this directly in the store
+func migratePost202203ResourceTranslations(ctx context.Context, log *zap.Logger, s store.Storer) (err error) {
+	log.Info("migrating post 202203 resource translations")
+
+	var (
+		migrated = make(map[string]bool)
+	)
+
+	set, _, err := store.SearchResourceTranslations(ctx, s, sysTypes.ResourceTranslationFilter{})
+	if err != nil {
+		return
+	}
+	err = set.Walk(func(r *sysTypes.ResourceTranslation) error {
+		var pos = strings.Index(r.Resource, "/")
+		if pos < 0 {
+			return nil
+		}
+
+		migrated[r.Resource[0:pos]] = true
+		return nil
+	})
+	if err != nil {
+		return
+	}
+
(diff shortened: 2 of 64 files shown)


darh:
will those commit messages get some love? :)

pkg/rbac/roles.go
```diff
@@ -55,7 +55,7 @@ func (k roleKind) String() string {
 	case AnonymousRole:
 		return "anonymous"
 	default:
-		panic("unknown role kind")
```
darh: what, why?

pkg/wfexec/session.go
```diff
@@ -612,7 +612,8 @@ func (s *Session) exec(ctx context.Context, log *zap.Logger, st *State) (nxt []*
 	st.created = *now()
 
 	defer func() {
-		reason := recover()
+		var reason interface{}
```
darh: what does this have to do with what this PR is about?

store/adapters/rdbms/upgrade_tables.go
```diff
@@ -696,3 +697,22 @@ func tableResourceActivityLog() *Table {
...
+		ID,
+		ColumnDef("name", ColumnTypeIdentifier),
+		ColumnDef("request_type", ColumnTypeText, ColumnTypeLength(handleLength)),
+		ColumnDef("status", ColumnTypeInteger),
+		ColumnDef("requested_at", ColumnTypeTimestamp),
+		ColumnDef("requested_by", ColumnTypeTimestamp),
+		ColumnDef("completed_at", ColumnTypeTimestamp),
+		ColumnDef("completed_by", ColumnTypeTimestamp),
+
+		AddIndex("request_type", IColumn("request_type")),
```
darh: index overload? do we really need all of them?

darh (changes requested):

store/adapters/rdbms/upgrade_tables.go
vicpatel: No

pkg/rbac/roles.go
vicpatel: Just IDE, fixing problems for me :D

pkg/wfexec/session.go
vicpatel: MB, It is not relevant to this PR.

compose/service/locale.go
```diff
@@ -272,6 +272,43 @@ func (svc resourceTranslationsManager) pageExtended(ctx context.Context, res *ty
...
+				Lang:     tag.String(),
+				Key:      yAxisLabelK.Path,
+				Msg:      svc.locale.TResourceFor(tag, res.ResourceTranslation(), yAxisLabelK.Path),
+			})
+		}
+
+		for _, metric := range report.Metrics {
+			if _, ok := metric["metricID"]; ok {
+				mpl := strings.NewReplacer(
+					"{{metricID}}", metric["metricID"].(string),
```
tjerman: This feels a bit dangerous to me; if it's not ok it will panic
I see you do some validation before this, but I would still prefer you do the *more safe* type assertion.
Alternatively, you can add a bunch of tests to assure this will be fine.

tjerman: Apply same comment for the rest of such cases

pkg/provision/migrations_202203_resource_translations.go
```diff
@@ -0,0 +1,110 @@
...
+// Migrates resource translations from the resource
+// struct to the dedicated store (table)
+//
+// While doing this, we also modify some resource substructure:
+//  - chart reports (assign report IDs)
+//
+// Note: we will migrate all translations to current default language
+// If you do not like that, shut down Corteza after migrations and fix this directly in the store
+func migratePost202203ResourceTranslations(ctx context.Context, log *zap.Logger, s store.Storer) (err error) {
+	log.Info("migrating post 202203 resource locales")
```
tjerman: `migrating post 202203 resource translations`

pkg/provision/migrations_202203_resource_translations.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var (
+		migrated = make(map[string]bool)
+	)
+
+	set, _, err := store.SearchResourceTranslations(ctx, s, sysTypes.ResourceTranslationFilter{})
```
tjerman, line 30: You missed the error check

pkg/wfexec/session.go
```diff
@@ -612,7 +612,8 @@ func (s *Session) exec(ctx context.Context, log *zap.Logger, st *State) (nxt []*
...
+		reason = recover()
 		if reason == nil {
```
tjerman: Why split it up?

system/data_privacy_request.cue
```diff
@@ -0,0 +1,65 @@
...
+
+	struct: {
+		id: schema.IdField
+		kind: { goType: "types.RequestKind", sortable: true }
+		status: { goType: "types.RequestStatus", sortable: true }
+
+		requested_at: { schema.SortableTimestampField, ident: "requestedAt", storeIdent: "requested_at" }
+		requested_by: { goType: "uint64", ident: "requestedBy", storeIdent: "requested_by" }
+		completed_at: { schema.SortableTimestampNilField, ident: "completedAt", storeIdent: "completed_at" }
+		completed_by: { goType: "uint64", ident: "completedBy", storeIdent: "completed_by" }
```
tjerman: I think you can omit all of the `ident` and `storeIdent` here

system/rest.yaml
```diff
@@ -2198,3 +2198,56 @@ endpoints:
...
+        post:
+          - { name: kind,         type: "string",   title: "Request Kind", required: true }
+    - name: update status
+      method: PUT
+      title: Update data privacy request status
+      path: "/{requestID}/status/{status}"
+      parameters:
+        path:
+          - { name: requestID,     type: "uint64",   title: "ID",             required: true }
+          - { name: status,        type: "string",   title: "Request Status", required: true }
```
tjerman: I think that `PATCH` would be more appropriate here since we're updating just the status

system/rest.yaml
```diff
@@ -2198,3 +2198,56 @@ endpoints:
...
+      parameters:
+        path:
+          - { name: requestID, type: "string",     title: "Request ID", required: true }
+    - name: create response
+      method: GET
+      title: Create response for a specific request
+      path: "/{requestID}/responses"
+      parameters:
+        path:
+          - { name: requestID, type: "string",     title: "Request ID", required: true }
```
tjerman: `POST` with a payload?

system/types/data_privacy.go
```diff
@@ -0,0 +1,92 @@
...
+	RequestStatusRejected RequestStatus = "rejected"
+)
+
+func (k RequestKind) IsValid() error {
+	switch k {
+	case RequestKindCorrect, RequestKindDelete, RequestKindExport:
+		return nil
+	}
+	return errors.New("invalid request kind")
+}
```
tjerman: If you only want to allow these request kinds, what I would consider here is not exposing the `RequestKind` type and defining some conversion function (either here, in the `types` or where you're constructing the privacy struct).

If you hide the type you don't need to worry about an invalid kind coming in so the `IsValid` is no longer needed.

If you decide to change it for the kind, also do so for the status.

tjerman (commented):
Before you merge this, consider cleaning up your commits a bit.
I sometimes `git reset --soft` all of my commits and re-do them to make them cleaner.

I also spotted you did some chart translation commits which probably shouldn't be in this PR; can you move those out?

pkg/wfexec/session.go
vicpatel: IDE warnings; this change doesn't relate to this PR.

system/rest.yaml
vicpatel: These routes were removed and no longer needed as we have new routes for request comments.

vicpatel:
> I also spotted you did some chart translation commits which probably shouldn't be in this PR; can you move those out?

Yeah, it was a bit of a blocker, but I kept a separate commit for it in this branch and squashed other data privacy requests related commits into one commit.

PR #363, opened 2022-05-25
Federation node url generate fix

petergrlica:
Fixed federation node generation url by using the `HTTP Server API base url` as a prefix when using webapps.


darh (approved):
lgtm

PR #366, opened 2022-05-31
Refactored migration of rbac resource rules

vicpatel:
Please take a look.

### DIFF
--- a/pkg/provision/migrations_202203_rbac_resource_fix.go
+++ b/pkg/provision/migrations_202203_rbac_resource_fix.go
@@ -3,7 +3,7 @@ package provision
 import (
 	"context"
 	"fmt"
-	"strconv"
+	"github.com/spf13/cast"
 	"strings"
 
 	composeTypes "github.com/cortezaproject/corteza-server/compose/types"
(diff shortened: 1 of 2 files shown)


pkg/provision/migrations_202203_rbac_resource_fix.go
```diff
@@ -118,149 +206,170 @@ func migratePost202203RbacRule(r *rbac.Rule, rx *resIndex) (op int) {
...
+func preloadRbacResourceIndex(ctx context.Context, s store.Storer, res fetchResIndex) (*resIndex, error) {
 	rx := &resIndex{}
 
 	rx.records = make(map[uint64]*composeTypes.Record)
 	var getRecords func(*composeTypes.Module, *filter.PagingCursor) (composeTypes.RecordSet, composeTypes.RecordFilter, error)
 	getRecords = func(m *composeTypes.Module, cursor *filter.PagingCursor) (rr composeTypes.RecordSet, f composeTypes.RecordFilter, err error) {
 		rr, f, err = store.SearchComposeRecords(ctx, s, m, composeTypes.RecordFilter{
 			Paging: filter.Paging{
-				Limit:      10000,
+				Limit:      30000,
```
tjerman: Why make this larger? Even the original 10k seems a bit extreme to me.

pkg/provision/migrations_202203_rbac_resource_fix.go
```diff
@@ -118,149 +206,170 @@ func migratePost202203RbacRule(r *rbac.Rule, rx *resIndex) (op int) {
...
 
-			r.Resource = composeTypes.ModuleFieldRbacResource(p1, p2, ID)
+	partsLen := len(parts)
+	ID := uint64(0)
+	p1 := uint64(0)
+	p2 := uint64(0)
+	parseUint := func(s string) uint64 {
+		id, _ := strconv.ParseUint(s, 10, 64)
+		return id
+	}
```
tjerman: You can use the `cast.ToUint64()` (the one without the error) to get rid of this custom bit

tjerman (commented):
To me, it seems fine; go through the few comments I've left and consider them.

The only requests I'd have is if you could write a nice long comment above `migratePost202203RbacRules` what exactly this does and why it exists, you replacing the `-1`, `0`, `1` *magic numbers* denoting operations with either a constant (so eg. `actionRemove`, `actionNone`, `actionUpdate`) or a new unexported type.

pkg/provision/migrations_202203_rbac_resource_fix.go
vicpatel: True, I will make it 2k just in case.

PR #374, opened 2022-06-23
Review and add missing col sorting

tjerman:
These are the columns that I thought would make sense to have sortable; thoughts?

### DIFF
--- a/compose/module-field.cue
+++ b/compose/module-field.cue
@@ -13,9 +13,9 @@ moduleField: schema.#Resource & {
 	struct: {
 		id: schema.IdField
 		namespace_id: { ident: "namespaceID", goType: "uint64", store: false }
-		module_id: { ident: "moduleID", goType: "uint64", storeIdent: "rel_module" }
-		place: { goType: "int" }
-		kind: { goType: "string" }
+		module_id: { sortable: true, ident: "moduleID", goType: "uint64", storeIdent: "rel_module" }
+		place: { sortable: true, goType: "int" }
+		kind: { sortable: true, goType: "string" }
 		name: {sortable: true}
 		label: {sortable: true}
 		options: { goType: "types.ModuleFieldOptions" }
--- a/federation/module-exposed.cue
+++ b/federation/module-exposed.cue
@@ -16,8 +16,8 @@ exposedModule: schema.#Resource & {
 	struct: {
 		id:          schema.IdField
 		handle:      schema.HandleField
-		name: {}
-		node_id: { ident: "nodeID", goType: "uint64", storeIdent: "rel_node" }
+		name: { sortable: true }
+		node_id: { sortable: true, ident: "nodeID", goType: "uint64", storeIdent: "rel_node" }
 		compose_module_id: { ident: "composeModuleID", goType: "uint64", storeIdent: "rel_compose_module" }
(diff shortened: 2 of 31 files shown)


compose/module-field.cue
```diff
@@ -13,8 +13,8 @@ moduleField: schema.#Resource & {
 	struct: {
 		id: schema.IdField
 		namespace_id: { ident: "namespaceID", goType: "uint64", store: false }
-		module_id: { ident: "moduleID", goType: "uint64", storeIdent: "rel_module" }
-		place: { goType: "int" }
+		module_id: { sortable: true, ident: "moduleID", goType: "uint64", storeIdent: "rel_module" }
+		place: { sortable: true, goType: "int" }
 		kind: { goType: "string" }
```
vicpatel: Maybe kind too?

vicpatel (approved):

federation/module-exposed.cue
```diff
@@ -16,7 +16,7 @@ exposedModule: schema.#Resource & {
 	struct: {
 		id:          schema.IdField
 		handle:      schema.HandleField
-		name: {}
+		name: { sortable: true }
 		node_id: { ident: "nodeID", goType: "uint64", storeIdent: "rel_node" }
 		compose_module_id: { ident: "composeModuleID", goType: "uint64", storeIdent: "rel_compose_module" }
```
petergrlica, line 21: why in `module-mapping.cue` and not here? or the other way around, why add sorting to the fields that are non-user facing and not used internally in sort? I am referring to `node_id`

federation/node.cue
```diff
@@ -13,11 +13,11 @@ node: schema.#Resource & {
 
 	struct: {
 		id:          schema.IdField
-		name: {}
-		shared_node_id: { ident: "sharedNodeID", goType: "uint64" }
+		name: {sortable: true}
+		shared_node_id: { sortable: true, ident: "sharedNodeID", goType: "uint64" }
 		base_url: { goType: "string", ident: "baseURL" }
```
petergrlica: add

system/dal_connection.cue
```diff
@@ -7,11 +7,11 @@ import (
...
 		id:     schema.IdField
-		name: { goType: "string" }
+		name: { sortable: true, goType: "string" }
 		handle: schema.HandleField
 
 		// omitting isPrimary and replacing with a special type
-		type: { goType: "string" }
+		type: { sortable: true }
 
 		location: { goType: "geolocation.Full" }
```
petergrlica, line 16: not sure how this would sort but maybe worth to take a look?

system/user.cue
```diff
@@ -8,10 +8,10 @@ user: schema.#Resource & {
...
 		id:     schema.IdField
 		handle: schema.HandleField
-		email: {unique: true, ignoreCase: true}
+		email: {sortable: true, unique: true, ignoreCase: true}
 		email_confirmed: {goType: "bool"}
-		username: {unique: true, ignoreCase: true}
-		name: {}
+		username: {sortable: true, unique: true, ignoreCase: true}
+		name: {sortable: true}
 		kind: {goType: "types.UserKind"}
```
petergrlica: add?

petergrlica (commented):
lgtm, maybe revise the points I expressed or reply

PR #376, opened 2022-06-24
Add validation for password while updating system settings

vicpatel:
Admin > System settings for Password constraints,
- We need to validate that the provided min-length (when changing it) option is not less than 8; if it is it should error out
- Also, we should not allow negative numbers for any password constraints otherwise return an error.

Please take a look.

### DIFF
--- a/tests/system/settings_test.go
+++ b/tests/system/settings_test.go
@@ -43,7 +43,7 @@ func TestSettingsList_noPermissions(t *testing.T) {
 		Header("Accept", "application/json").
 		Expect(t).
 		Status(http.StatusOK).
-		Assert(helpers.AssertError("not allowed to read settings")).
+		Assert(helpers.AssertError("settings.errors.notAllowedToRead")).
 		End()
 }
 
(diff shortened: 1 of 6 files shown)


system/types/settings.go
```diff
@@ -38,6 +39,7 @@ type (
 )
 
 const (
+	passwordMinLength        = 8
```
darh: What is the reason that we're enforcing this?

tests/system/settings_test.go
```diff
@@ -88,6 +88,21 @@ func TestSettingsUpdate_noPermissions(t *testing.T) {
 		End()
 }
 
+func TestSettingsUpdate_validation(t *testing.T) {
+	h := newHelper(t)
+	helpers.AllowMe(h, types.ComponentRbacResource(), "settings.manage")
+	helpers.AllowMe(h, types.ComponentRbacResource(), "settings.read")
+
+	h.apiInit().
```
darh, line 97: are there other cases and input variants we could test here?

darh (changes requested):

tests/system/settings_test.go
vicpatel: Yeah, we can I will update it.

system/types/settings.go
vicpatel: We have a default min-length for password is 8.

PR #377, opened 2022-07-01
Forbid password reuse

petergrlica:
When:
 - change password in admin
 - change password on auth
 - reset password via email

### DIFF
--- a/system/service/auth.go
+++ b/system/service/auth.go
@@ -2,27 +2,19 @@ package service
 
 import (
 	"context"
-	"fmt"
-	"math"
-	rand2 "math/rand"
 	"regexp"
-	"strconv"
 	"time"
 
 	"github.com/cortezaproject/corteza-server/pkg/actionlog"
 	internalAuth "github.com/cortezaproject/corteza-server/pkg/auth"
 	"github.com/cortezaproject/corteza-server/pkg/errors"
 	"github.com/cortezaproject/corteza-server/pkg/eventbus"
-	"github.com/cortezaproject/corteza-server/pkg/filter"
 	"github.com/cortezaproject/corteza-server/pkg/handle"
 	"github.com/cortezaproject/corteza-server/pkg/payload"
-	"github.com/cortezaproject/corteza-server/pkg/rand"
 	"github.com/cortezaproject/corteza-server/store"
 	"github.com/cortezaproject/corteza-server/system/service/event"
 	"github.com/cortezaproject/corteza-server/system/types"
-	"github.com/dgryski/dgoogauth"
 	"github.com/markbates/goth"
-	"golang.org/x/crypto/bcrypt"
 )
 
 type (
@@ -42,6 +34,11 @@ type (
 
 	AuthOptions struct {
 		LimitUsers int
+
+		// how fresh can a password be before we consider it
+		// reused?
+		// @todo make this configurable someday
+		PasswordReuseTimeWindow time.Duration
 	}
 
 	authAccessController interface {
@@ -50,31 +47,8 @@ type (
 	}
 )
 
-const (
-	credentialsTypePassword                    = "password"
-	credentialsTypePersistentSession           = "persistent-session"
-	credentialsTypeEmailAuthToken              = "email-authentication-token"
-	credentialsTypeResetPasswordToken          = "password-reset-token"
-	credentialsTypeResetPasswordTokenExchanged = "password-reset-token-exchanged"
-	credentialsTypeCreatePasswordToken         = "password-create-token"
-	credentialsTypeMfaTotpSecret               = "mfa-totp-secret"
-	credentialsTypeMFAEmailOTP                 = "mfa-email-otp"
-
-	credentialsTokenLength = 32
-
-	passwordMinLength = 8
-	passwordMaxLength = 256
-
-	tokenReqMaxCount  = 5
-	tokenReqMaxWindow = time.Minute * 15
-)
-
 var (
 	reEmail = regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
-
-	oneTokenPerUser = map[string]bool{
-		credentialsTypeResetPasswordToken: true,
-	}
 )
 
 func defaultProviderValidator(provider string) error {
@@ -120,7 +94,7 @@ func Auth(opt AuthOptions) *auth {
 //
 // External login/signup does not:
 //  - validate provider on profile, only uses it for matching credentials
-func (svc auth) External(ctx context.Context, profile types.ExternalAuthUser) (u *types.User, err error) {
+func (svc *auth) External(ctx context.Context, profile types.ExternalAuthUser) (u *types.User, err error) {
 	var (
 		authProvider = &types.AuthProvider{Provider: profile.Provider}
 
@@ -312,7 +286,7 @@ func (svc auth) External(ctx context.Context, profile types.ExternalAuthUser) (u
 // Forgiving but strict: valid existing users get notified
 //
 // We're accepting the whole user object here and copy all we need to the new user
-func (svc auth) InternalSignUp(ctx context.Context, input *types.User, password string) (u *types.User, err error) {
+func (svc *auth) InternalSignUp(ctx context.Context, input *types.User, password string) (u *types.User, err error) {
 	var (
 		authProvider = &types.AuthProvider{Provider: credentialsTypePassword}
 
@@ -352,7 +326,8 @@ func (svc auth) InternalSignUp(ctx context.Context, input *types.User, password 
 				return err
 			}
 
-			if c = cc.CompareHashAndPassword(password, true); c == nil {
+			// does password match any of the valid credentials?
+			if !isValidPassword(cc, password) {
 				return AuthErrInvalidCredentials(aam)
 			}
 
@@ -458,7 +433,7 @@ func (svc auth) InternalSignUp(ctx context.Context, input *types.User, password 
 // InternalLogin verifies username/password combination in the internal credentials table
 //
 // Expects plain text password as an input
-func (svc auth) InternalLogin(ctx context.Context, email string, password string) (u *types.User, err error) {
+func (svc *auth) InternalLogin(ctx context.Context, email string, password string) (u *types.User, err error) {
 	var (
 		authProvider = &types.AuthProvider{Provider: credentialsTypePassword}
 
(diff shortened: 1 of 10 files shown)


system/service/auth.go
```diff
@@ -549,6 +550,23 @@ func (svc auth) SetPassword(ctx context.Context, userID uint64, password string)
...
+		cc, _, err = store.SearchCredentials(ctx, svc.store, types.CredentialsFilter{
+			Kind:    credentialsTypePassword,
+			OwnerID: userID,
+			Deleted: filter.StateInclusive})
+
+		if err != nil {
+			return err
+		}
+
+		if svc.CheckPassword(password, true, cc) {
```
darh: can you explain here why we do 2 calls

system/types/credentials.go
```diff
@@ -36,10 +36,9 @@ func (u *Credentials) Valid() bool {
 }
 
 // CompareHashAndPassword returns first valid credentials with matching hash
```
darh: with introduction of the validOnly the description of the function is no longer correct

system/service/user.go
```diff
@@ -715,6 +718,19 @@ func (svc user) SetPassword(ctx context.Context, userID uint64, newPassword stri
 			return svc.auth.RemovePasswordCredentials(ctx, userID)
 		}
 
+		cc, _, err = store.SearchCredentials(ctx, svc.store, types.CredentialsFilter{
```
darh: this block of code is repeated 3 times, it should be moved into a utility func

darh (changes requested):

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+		credentialsTypeResetPasswordToken: true,
+	}
+)
+
+// ValidateEmailConfirmationToken issues a validation token that can be used for
+func (svc *auth) ValidateEmailConfirmationToken(ctx context.Context, token string) (user *types.User, err error) {
+	return svc.loadFromTokenAndConfirmEmail(ctx, token, credentialsTypeEmailAuthToken)
+}
+
+// loadFromTokenAndConfirmEmail loads token, confirms user's
```
tjerman, line 57: ... confirms user's email

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+		}
+
+		out = append(out, c)
+		if limit == len(out) {
+			break
+		}
+	}
+
+	return
+}
```
tjerman, line 1157: Worth considering defining a `type CredentialsSet []*types.Credentials` and having this there

system/service/auth.go
```diff
@@ -600,1082 +514,128 @@ func (svc auth) Impersonate(ctx context.Context, userID uint64) (u *types.User,
...
 		}
 	}
 
-	// - special characters
-	count = svc.settings.Auth.Internal.PasswordConstraints.MinSpecialCount
-	if count > 0 {
-		rr := regexp.MustCompile("[^0-9a-zA-Z]")
-		if uint(len(rr.FindAllStringIndex(password, -1))) < count {
-			return false
+	if c != nil {
```
tjerman, line 573: can `c` ever be `nil`? I'd imagine proc login should never happen if there is no `c`

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+const (
+	credentialsTypePassword                    = "password"
+	credentialsTypePersistentSession           = "persistent-session"
+	credentialsTypeEmailAuthToken              = "email-authentication-token"
+	credentialsTypeResetPasswordToken          = "password-reset-token"
+	credentialsTypeResetPasswordTokenExchanged = "password-reset-token-exchanged"
+	credentialsTypeCreatePasswordToken         = "password-create-token"
+	credentialsTypeMfaTotpSecret               = "mfa-totp-secret"
+	credentialsTypeMFAEmailOTP                 = "mfa-email-otp"
```
tjerman, line 35: Should we enum these?

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+		if !u.EmailConfirmed {
+			// User's email is not confirmed but going through password reset flow
+			// we can confirm it
+			u.EmailConfirmed = true
+			u.UpdatedAt = now()
+			if err = store.UpdateUser(ctx, svc.store, u); err != nil {
+				return err
+			}
+		}
```
tjerman, line 87: Worth considering having another event here; `on userEmailConfirmed` for things like welcome or success email notifications, lead -> customer conversion, ...

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+	}
+)
+
+// ValidateEmailConfirmationToken issues a validation token that can be used for
+func (svc *auth) ValidateEmailConfirmationToken(ctx context.Context, token string) (user *types.User, err error) {
+	return svc.loadFromTokenAndConfirmEmail(ctx, token, credentialsTypeEmailAuthToken)
+}
+
+// loadFromTokenAndConfirmEmail loads token, confirms user's
+func (svc *auth) loadFromTokenAndConfirmEmail(ctx context.Context, token, tokenType string) (u *types.User, err error) {
```
tjerman, line 58: Should the action log logging happen in the parent function instead of here?

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+// Loads user from token and removes that token right after
+func (svc *auth) loadUserFromToken(ctx context.Context, token, kind string) (u *types.User, _ error) {
+	var (
+		aam = &authActionProps{
+			credentials: &types.Credentials{Kind: kind},
+		}
+	)
+
+	return u, svc.store.Tx(ctx, func(ctx context.Context, s store.Storer) (err error) {
```
tjerman, line 128: I think we usually use `store.Tx` instead of the direct receiver fnc

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+		})
+
+		if err != nil {
+			return err
+		}
+
+		// gt/eq since this current request is not yet stored
+		if err = svc.checkTokenRate(cc, tokenReqMaxWindow, tokenReqMaxCount); err != nil {
+			return
+		}
```
tjerman, line 203: I'd move this to some function to make the current one a bit shorter

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+		// removes expired and soft-deleted tokens
+		// and enforces one-token-per-user rule
```
tjerman, line 206: This should be the comment for the function definition

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+		}
+
+		switch kind {
+		case credentialsTypeMFAEmailOTP:
+			// do not alter the final token
+		default:
+			// suffixing tokens with credentials ID
+			// this will help us with token lookups
+			token = fmt.Sprintf("%s%d", token, c.ID)
+		}
```
tjerman, line 260: I'd use `if kind != ...` just to make it shorter

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+	}
+
+	if count > max {
+		return AuthErrRateLimitExceeded()
+	}
+
+	return nil
+}
+
+func (svc *auth) cleanupCredentials(ctx context.Context, s store.Credentials, cc types.CredentialsSet) (err error) {
```
tjerman, line 299: Won't this delete old user passwords making the password reuse check not possible?

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+	)
+
+	err = func() error {
+		if !svc.settings.Auth.Internal.Enabled {
+			return AuthErrInternalLoginDisabledByConfig(aam)
+		}
+
+		if len(oldPassword) == 0 {
+			return AuthErrPasswordNotSecure(aam)
+		}
```
tjerman, line 481: This doesn't need to happen; if the password was previously ok and the code below matches to it, this bit doesn't matter

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+	}()
+
+	return svc.recordAction(ctx, aam, AuthActionChangePassword, err)
+}
+
+func (svc *auth) hashPassword(password string) (hash []byte, err error) {
+	return bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+}
+
+func (svc *auth) CheckPasswordStrength(password string) bool {
```
tjerman, line 530: Should we move the `len(password) == 0` check to here and remove it elsewhere?

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+		Credentials: string(hash),
+	}
+
+	return store.CreateCredentials(ctx, svc.store, c)
+}
+
+//CheckPassword verifies if password matches any of the valid credentials
+//func (svc *auth) CheckPassword(cc types.CredentialsSet, password string) bool {
+//	return findMatchingCredentials(cc, password, true) != nil
+//}
```
tjerman, line 600: Delete

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+	cc, _, err := store.SearchCredentials(ctx, svc.store, types.CredentialsFilter{
+		OwnerID: u.ID,
+		Kind:    credentialsTypePassword,
+	})
+	if err != nil {
+		return
+	}
+
+	if len(cc) > 0 && svc.settings.Auth.Internal.Enabled {
```
tjerman, line 685: you can move the `svc.settings.Auth.Internal.Enabled` bit to the top and avoid needless computation in case it is not enabled

system/service/auth_credentials.go
```diff
@@ -0,0 +1,1157 @@
...
+
+	// Do a partial update and soft-delete all
+	return store.UpdateCredentials(ctx, svc.store, cc...)
+}
+
+// checks if password is reused by looking all valid and invalid credentials
+// and taking reuse time window into consideration
+func (svc *auth) checkPasswordReuse(cc []*types.Credentials, password string) bool {
+	return false
+}
```
tjerman, line 723: remove

system/service/auth_credentials.go
darh: we have it, I chose not to use it.... no spec. reason.

system/service/auth.go
darh: You are correct but I remember some nil pointer errors that this might have addressed. 
I've moved the checks (+ errors) at the top of the fn

system/service/auth_credentials.go
darh: yap.

system/service/auth_credentials.go
darh: nope, might break old pre-pwd-strength-policy passwords.

system/service/auth_credentials.go
darh: how could empty password ever be ok?

system/service/auth_credentials.go
darh: not for now.

PR #380, opened 2022-07-06
Add filters to permissions route for role

vicpatel:
It allows filtering for specific rules and also the rules which are applied to the resource, and not to a specific resource.

Introduces generic methods for RuleSet and FindRules method to access_control generation template.

Please take a look.

### DIFF
--- a/automation/rest/request/permissions.go
+++ b/automation/rest/request/permissions.go
@@ -11,6 +11,7 @@ package request
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/cortezaproject/corteza-server/pkg/filter"
 	"github.com/cortezaproject/corteza-server/pkg/payload"
 	"github.com/cortezaproject/corteza-server/pkg/rbac"
 	"github.com/go-chi/chi/v5"
@@ -66,6 +67,16 @@ type (
 		//
 		// Role ID
 		RoleID uint64 `json:",string"`
+
+		// Specific GET parameter
+		//
+		// Exclude (0, default), include (1) or return only (2) specific rules
+		Specific filter.State
+
+		// Resource GET parameter
+		//
+		// Show only rules for a specific resource
+		Resource []string
 	}
 
 	PermissionsDelete struct {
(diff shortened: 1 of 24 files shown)


pkg/rbac/rule.go
```diff
@@ -71,6 +71,15 @@ func (set RuleSet) FilterResource(res string) (out RuleSet) {
 	return
 }
 
+func (set RuleSet) FilterGenericRules() (out RuleSet) {
```
darh: can we use [copy](https://pkg.go.dev/builtin#copy)() here?

system/rest/request/permissions.go
```diff
@@ -66,6 +66,11 @@ type (
 		//
 		// Role ID
 		RoleID uint64 `json:",string"`
+
+		// GenericOnly GET parameter
+		//
+		// Show only generic rules for resources
```
darh: Can you explain what "generic" rules are? We did not have this concept before.

vicpatel: Generic rules as in more common rule which applies to the resource in general, and not specific to a particular resource.

pkg/rbac/rule.go
vicpatel: There is more inside that func now, after a new commit :)

codegen/assets/templates/gocode/rbac/$component_access_control.go.tpl
```diff
@@ -204,6 +205,52 @@ func (svc accessControl) logGrants(ctx context.Context, rr []*rbac.Rule) {
 	}
 }
 
+// FindRules find all rules based on filters
+//
+// This function is auto-generated
+func (svc accessControl) FindRules(ctx context.Context, roleID uint64, specific uint, rr ...string) (out rbac.RuleSet, err error) {
```
darh: why not `filter.State` instead of `uint`?

automation/rest/request/permissions.go
```diff
@@ -66,6 +66,16 @@ type (
 		//
 		// Role ID
 		RoleID uint64 `json:",string"`
+
+		// Specific GET parameter
+		//
+		// Exclude (0, default), include (1) or return only (2) specific rules
+		Specific uint
```
darh: `filter.State`?

darh (changes requested):

automation/rest/request/permissions.go
vicpatel: Sure

PR #381, opened 2022-07-06
Support translation of chart dimension steps 

darh:
This also introduces cleaner way to handle nested (chart) config and applies that to metric labels as well.

### DIFF
--- a/compose/types/chart.go
+++ b/compose/types/chart.go
@@ -4,6 +4,9 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"github.com/cortezaproject/corteza-server/pkg/locale"
+	"github.com/spf13/cast"
+	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


compose/types/chart.go
```diff
@@ -151,3 +189,70 @@ func (cc *ChartConfig) Scan(value interface{}) error {
...
+		return v.FieldByName(idx)
+	}
+
+	v := reflect.ValueOf(d)
+	for _, s := range path {
+		v = index(v, s)
+	}
+
+	v.Set(reflect.ValueOf(value))
+}
```
tjerman, line 264: Unused?

compose/types/chart_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"testing"
+)
+
+func TestChart_decodeTranslations(t *testing.T) {
+	cc := []struct {
+		name string
+		base *ChartConfigReport
+		ccr  *ChartConfigReport
+		tt   locale.ResourceTranslationIndex
+	}{
```
tjerman, line 16: Should we also use the stringified JSON payload as you've used below? Would make sense to me

compose/types/chart.go
```diff
@@ -102,21 +129,32 @@ func (c Chart) encodeTranslations() (out locale.ResourceTranslationSet) {
...
+		// collect labels from chart config: metrics
+		report.walkMetrics(func(metricID string, m map[string]interface{}) {
+			mpl := strings.NewReplacer(
+				"{{metricID}}", metricID,
+			)
+
+			out = append(out, &locale.ResourceTranslation{
+				Resource: c.ResourceTranslation(),
+				Key:      mpl.Replace(LocaleKeyChartMetricsMetricIDLabel.Path),
+				Msg:      m["label"].(string),
```
tjerman: Is the label guaranteed to be present? We do a check above but we don't prevent this code from happening

tjerman (commented):
A few concerns noted; else this looks ok to me

PR #382, opened 2022-07-06
Refactor RBAC rule cloning route for role

vicpatel:
It moves role permission cloning under /system/roles from /system/permissions, since cloning action copies all rules, and it can not be limited per-component and also shifted all respective service methods to role service and removed Rbac rule cloning methods from access-control template.

Please take a look.

### DIFF
--- a/system/service/access_control.gen.go
+++ b/system/service/access_control.gen.go
@@ -26,7 +26,6 @@ type (
 		Evaluate(rbac.Session, string, rbac.Resource) rbac.Evaluated
 		Grant(context.Context, ...*rbac.Rule) error
 		FindRulesByRoleID(roleID uint64) (rr rbac.RuleSet)
-		CloneRulesByRoleID(ctx context.Context, fromRoleID uint64, toRoleID ...uint64) error
 	}
 
 	accessControl struct {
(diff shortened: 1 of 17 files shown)


system/service/role.go
```diff
@@ -604,6 +611,10 @@ func (svc role) Unarchive(ctx context.Context, roleID uint64) (err error) {
 	return svc.recordAction(ctx, raProps, RoleActionUnarchive, err)
 }
 
+func (svc role) CloneRules(ctx context.Context, roleID uint64, cloneToRoleID ...uint64) (err error) {
+	return svc.rbac.CloneRulesByRoleID(ctx, roleID, cloneToRoleID...)
+}
```
tjerman: Do we not need any access control here?
I'd imagine we should check for the can grant perm

system/service/access_control.gen.go
```diff
@@ -578,17 +577,6 @@ func (svc accessControl) FindRulesByRoleID(ctx context.Context, roleID uint64) (
 	return svc.rbac.FindRulesByRoleID(roleID), nil
 }
 
-// CloneRulesByRoleID clone all rules of a Role S to a specific Role T
-//
-// This function is auto-generated
-func (svc accessControl) CloneRulesByRoleID(ctx context.Context, fromRoleID uint64, toRoleID ...uint64) error {
-	if !svc.CanGrant(ctx) {
-		return AccessControlErrNotAllowedToSetPermissions()
-	}
```
tjerman, line 587: Here we check for `CanGrant` but not in the refactored version

tjerman (changes requested):

PR #383, opened 2022-07-12
2022.9.x feature new jwt algos

darh:


vicpatel (approved):
LGTM, Love the new auth step on boot level :)

PR #385, opened 2022-07-14
Tweak rv sanitizer->value expression interaction

tjerman:
When record value expression returns null the value gets omitted.
Tweak value sanitizer logic for more consistency.


darh (approved):
lgtm.

PR #389, opened 2022-07-21
Fix sensitive level check for data privacy modules

vicpatel:
A connection, module, and module field, all (can) have a sensitivity level. If the sensitivity level is set, then it will consider private.

Please take a look.

### DIFF
--- a/compose/service/data_privacy.go
+++ b/compose/service/data_privacy.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"github.com/cortezaproject/corteza-server/compose/types"
+	"github.com/cortezaproject/corteza-server/pkg/locale"
 	sysService "github.com/cortezaproject/corteza-server/system/service"
 	sysTypes "github.com/cortezaproject/corteza-server/system/types"
 )
@@ -12,6 +13,7 @@ type (
 		ns      NamespaceService
 		m       ModuleService
 		dalConn dalConnectionService
+		locale  ResourceTranslationsManagerService
 	}
 
 	dalConnectionService interface {
@@ -28,6 +30,7 @@ func DataPrivacy() *dataPrivacy {
 		ns:      DefaultNamespace,
 		m:       DefaultModule,
 		dalConn: sysService.DefaultDalConnection,
+		locale:  DefaultResourceTranslation,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


compose/service/data_privacy.go
```diff
@@ -76,10 +76,11 @@ func (svc dataPrivacy) FindModules(ctx context.Context, filter types.PrivacyModu
 			}
 
 			out = append(out, &types.PrivacyModule{
-				ID:         m.ID,
-				Name:       m.Name,
-				Handle:     m.Handle,
-				Connection: c,
+				ID:          m.ID,
+				Name:        m.Name,
```
tjerman: Is the name here pulled from resource translations? Should it be?

compose/service/module.go
```diff
@@ -443,19 +444,46 @@ func (svc module) FindSensitive(ctx context.Context, filter types.PrivacyModuleF
...
 					Name:         m.Name, // @todo get this as per translation
 					Handle:       m.Handle,
+					NamespaceID:  m.NamespaceID,
 					ConnectionID: m.ModelConfig.ConnectionID,
-				})
+				}
+
+				// Set default connection ID
+				if pm.ConnectionID == 0 {
+					pm.ConnectionID, err = svc.dal.GetDefaultConnectionID(ctx)
```
tjerman: Here, I would add a new `ConnectionID` field to the `ConnectionMeta` struct and use `cMeta`

compose/service/module.go
```diff
@@ -443,19 +444,46 @@ func (svc module) FindSensitive(ctx context.Context, filter types.PrivacyModuleF
...
 			if isPrivate && m != nil {
-				set = append(set, types.PrivacyModule{
+				pm := types.PrivacyModule{
 					ID:           m.ID,
 					Name:         m.Name, // @todo get this as per translation
 					Handle:       m.Handle,
+					NamespaceID:  m.NamespaceID,
 					ConnectionID: m.ModelConfig.ConnectionID,
-				})
+				}
```
tjerman: What about module fields; do we not want those here?

Speaking of, the `isPrivate` is no longer correct as is since now all modules of a connection would be included (this bit `isPrivate = cMeta.SensitivityLevel > 0`).
Shouldn't we just look at the field's privacy level?

btw, you can replace all those `if`s  with simple bool operations and remove the branching

pkg/dal/service.go
```diff
@@ -325,6 +325,19 @@ func (svc *service) ReplaceConnection(ctx context.Context, cw *ConnectionWrap, i
...
+
+	cw := svc.getConnectionByID(ID)
+	if cw == nil {
+		err = errConnectionNotFound(ID)
+		return
+	}
+
+	return
+}
+
```
tjerman: If you take my suggestion on the connection meta struct, this can be removed

compose/service/module.go
vicpatel: > What about module fields?

No, but it could be beneficial for the future, I will add it. :)

`isPrivate` was misleading and I didn't rename it after sensitive level changes, fixed it now.

Initially, I did only check the field's privacy level, but we are listing sensitive modules, so along with the module's field, If module and connection have sensitive level then shouldn't we consider those too?

PR #394, opened 2022-07-27
2022.9.x feature record revisions

darh:
Please take a look.

### DIFF
--- a/compose/rest/record.go
+++ b/compose/rest/record.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/cortezaproject/corteza-server/pkg/revisions"
 	"io"
 	"math"
 	"net/http"
@@ -37,7 +38,9 @@ type (
 		CanUpdateRecord        bool `json:"canUpdateRecord"`
 		CanReadRecord          bool `json:"canReadRecord"`
 		CanDeleteRecord        bool `json:"canDeleteRecord"`
-		CanGrant               bool `json:"canGrant"`
+		CanSearchRevisions     bool `json:"canSearchRevisions"`
+
+		CanGrant bool `json:"canGrant"`
 	}
 
 	recordSetPayload struct {
@@ -61,6 +64,7 @@ type (
 		CanReadRecord(context.Context, *types.Record) bool
 		CanDeleteRecord(context.Context, *types.Record) bool
 		CanManageOwnerOnRecord(context.Context, *types.Record) bool
+		CanSearchRevisionsOnRecord(context.Context, *types.Record) bool
 	}
 )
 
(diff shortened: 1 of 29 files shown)


compose/rest/record.go
```diff
@@ -567,6 +571,34 @@ func (ctrl *Record) TriggerScriptOnList(ctx context.Context, r *request.RecordTr
...
+		err = dal.IteratorEncodeJSON(ctx, w, iter, makeRev)
+		if err != nil {
+			return
+		}
+
+		if _, err = w.Write([]byte(`]}`)); err != nil {
+			return
+		}
+
+		return
```
tjerman, line 598: I feel like this bit could be moved to some rest util. function (especially the opening and closing bytes

compose/types/module.go
```diff
@@ -26,6 +26,16 @@ type (
...
 
+		Revisions struct {
+			Enabled bool `json:"enabled"`
+
+			DAL struct {
+				// where will revisions be stored
+				// can contain variables like {{module}}, {{namespace}}
+				Ident string `json:"ident"`
+			} `json:"dal"`
+		} `json:"revisions"`
```
tjerman: Should this be moved to its own type definition?

compose/types/module.go
```diff
@@ -26,6 +26,16 @@ type (
...
+
 		SystemFieldEncoding SystemFieldEncoding `json:"systemFieldEncoding"`
```
tjerman: In here, you'll also need to register the new revision system feld

tjerman: In here as in the struct def.

compose/service/module.go
```diff
@@ -1223,12 +1219,49 @@ func moduleToModel(ctx context.Context, dmm dalModelManager, ns *types.Namespace
...
+	if !ok {
+		err = fmt.Errorf("invalid model ident generated: %s", ident)
+		return
+	}
+
+	return ident, nil
+}
+
 // moduleFieldsToAttributes converts all user-defined module fields to attributes
 func moduleFieldsToAttributes(ctx context.Context, cm dal.ConnectionMeta, ns *types.Namespace, mod *types.Module) (out dal.AttributeSet, err error) {
```
tjerman, line 1282: There is another fnc you'll need to update along with the struct change mentioned above (the added revision sys field comment) -- `partitionedModuleSystemFieldsToAttributes`

compose/service/record.go
```diff
@@ -435,6 +442,43 @@ func (svc record) findSensitive(ctx context.Context, userID uint64, namespace *t
...
+		aProps.setNamespace(ns)
+		aProps.setRecord(rec)
+
+		if !mod.ModelConfig.Revisions.Enabled {
+			return RecordErrRevisionsDisabledOnModule()
+		}
+
+		if !svc.ac.CanSearchRevisionsOnRecord(ctx, rec) {
+			return RecordErrNotAllowedToSearchRevisions()
+		}
```
tjerman, line 473: should we also have a `read` perm. for specific revisions?
I'd say no, but to consider

compose/service/record.go
```diff
@@ -617,6 +661,13 @@ func (svc record) create(ctx context.Context, new *types.Record) (rec *types.Rec
 		return
 	}
 
+	// store revision
+	if m.ModelConfig.Revisions.Enabled {
+		if err = svc.revisions.created(ctx, new); err != nil {
+			return
+		}
+	}
```
tjerman: We discussed doing revisions as a delta where the old value is logged and the record outlines the latest state.
What made it change?

provision/000_base/compose_access_control.yaml
```diff
@@ -60,6 +60,8 @@ allow:
       - read
       - update
       - delete
+      - revisions.search
+
```
tjerman, line 64: rm extra new line

compose/service/record.go
darh: no

compose/service/record.go
darh: feels a bit less magical.

compose/rest/record.go
darh: A agree. As soon as this proves stable an we have a 2nd place were we can use it

PR #396, opened 2022-08-01
Fix Action logs placeholders for resources

vicpatel:
Mainly, It was broken for creation of most resources due to misplace for action-log prop.

Please take a look.


darh (approved):
very nice.

PR #397, opened 2022-08-01
Implement base valuestore for env variables

tjerman:
* Base facility to store generic values,
* integrate with templates,
* integrate with workflows.


app/options/HTTPClient.cue
```diff
@@ -17,6 +17,10 @@ HTTPClient: schema.#optionsGroup & {
 	]
 
 	options: {
+		domain: {
```
darh: I have a feeling you mixed up 2 things here

http-client is something else -- an actual HTTP Client we use to execute outbound requests

app/options/HTTPServer.cue
```diff
@@ -13,6 +13,11 @@ HTTPServer: schema.#optionsGroup & {
 	]
 
 	options: {
+		domain: {
```
darh: it's a bit more complicated -- see `guessHostname` fn.

darh (changes requested):

PR #398, opened 2022-08-10
2022.9.x feature record meta

darh:
Please take a look.

### DIFF
--- a/compose/rest.yaml
+++ b/compose/rest.yaml
@@ -643,7 +643,6 @@ endpoints:
   path: "/namespace/{namespaceID}/module/{moduleID}/record"
   authentication: []
   imports:
-    - github.com/cortezaproject/corteza-server/pkg/label
     - github.com/cortezaproject/corteza-server/compose/types
   parameters:
     path:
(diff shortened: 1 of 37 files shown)


compose/rest/record.go
```diff
@@ -174,7 +174,7 @@ func (ctrl *Record) Create(ctx context.Context, r *request.RecordCreate) (interf
 			NamespaceID: r.NamespaceID,
 			ModuleID:    r.ModuleID,
 			Values:      r.Values,
-			Labels:      r.Labels,
+			Meta:        r.Labels,
```
tjerman: This should be `Meta: r.Meta`, no?
Since we still take in both the `labels` and `meta` parameters, you'll need to check both here.

compose/rest.yaml
```diff
@@ -684,10 +683,14 @@ endpoints:
...
+      - type: map[string]any
+        name: meta
+        title: Record meta data
+        parser: payload.ParseMeta
+      - type: map[string]any
         name: labels
-        title: Labels
-        parser: label.ParseStrings
+        title: Labels, deprecated, use meta instead
+        parser: payload.ParseMeta
```
tjerman: Should we just remove it instead of deprecating it? I would vote to just remove it

compose/types/module.go
```diff
@@ -54,6 +54,7 @@ type (
 		NamespaceID *EncodingStrategy `json:"namespaceID"`
 
 		Revision *EncodingStrategy `json:"revision"`
+		Meta     *EncodingStrategy `json:"revision"`
```
tjerman: Lapsus on JSON tag; should be `meta`

tjerman: Should you give it an `omitempty` as well (you've done it on types/record)?

compose/types/record.go
```diff
@@ -338,7 +333,7 @@ func (r Record) Dict() map[string]interface{} {
 		"recordID":    r.ID,
 		"moduleID":    r.ModuleID,
 		"revision":    r.Revision,
-		"labels":      r.Labels,
+		"labels":      r.Meta,
```
tjerman: Should this be `"meta": r.Meta`?

tjerman: We can have both (`meta` and `labels`) if we want to make it compatible

discovery/rest/internal/documents/compose.go
```diff
@@ -362,7 +362,7 @@ func (d composeResources) Records(ctx context.Context, namespaceID, moduleID uin
 				Namespace:    nsPartial,
 				Module:       modPartial,
 				Url:          d.getUrlToResource(pageDetail{moduleID: moduleID, recordID: recordID}),
-				Labels:       rec.Labels,
+				Labels:       rec.Meta,
```
tjerman: Should this also rename?

PR #405, opened 2022-08-25
Duplicate detection 

vicpatel:
Please take a look.

### DIFF
--- a/compose/service/record.go
+++ b/compose/service/record.go
@@ -51,9 +51,10 @@ type (
 
 		revisions *recordRevisions
 
-		formatter recordValuesFormatter
-		sanitizer recordValuesSanitizer
-		validator recordValuesValidator
+		formatter   recordValuesFormatter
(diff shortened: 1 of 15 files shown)


pkg/detection/detection.go
```diff
@@ -0,0 +1,220 @@
+package detection
```
darh: what else, besides duplicates, will this pkg detect?

compose/service/record.go
```diff
@@ -629,6 +634,39 @@ func (svc record) create(ctx context.Context, new *types.Record) (rec *types.Rec
...
+			md.DuplicationRules,
+			new.Values.CastToDetectionSet(),
+			records.CastToValueDetectionSet(),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		// Error out of duplicate record exist
+		if md.Strict || len(set) > 0 {
```
darh: wouldn't this be per-constraint(field), not per module?

PR #406, opened 2022-08-29
Add support for exec-workflow step type

darh:
Please take a look.

### DIFF
--- a/automation/service/workflow.go
+++ b/automation/service/workflow.go
@@ -564,12 +564,21 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
 		// Find the trigger.
 		// @todo can we cache this as well?
 		t, err = func() (*types.Trigger, error) {
+			if p.CallerWorkflowID > 0 {
+				// skip triggers checking when executed as sub-workflow
+				// @todo be more strict and allow this ONLY when workflow is flagged as a sub-workflow
+			}
+
 			var tt types.TriggerSet
 			// Load triggers directly from the store. At this point we do not care
 			// about trigger search or read permissions
 			tt, err = loadWorkflowTriggers(ctx, svc.store, workflowID)
 			if err != nil {
 				return nil, err
+			}
+
+			if len(tt) == 0 {
+				return nil, nil
 			}
 
 			if p.StepID == 0 && len(tt) > 0 {
@@ -582,32 +591,32 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
 				}
 			}
 
+			if !p.Trace {
+				if t == nil {
+					return nil, WorkflowErrUnknownWorkflowStep()
+				} else if !t.Enabled {
+					return nil, WorkflowErrDisabled()
+				}
+			}
+
+			if t != nil {
+				wap.setTrigger(t)
+				p.StepID = t.StepID
+				p.EventType = t.EventType
+				p.ResourceType = t.ResourceType
+
+				// merge with input from trigger
+				// with trigger input vars are overwritten by input vars
+				p.Input = t.Input.MustMerge(p.Input)
+			} else {
+				p.EventType = "onTrace"
+			}
+
 			return nil, nil
 		}()
 
 		if err != nil {
 			return
-		}
-
-		if !p.Trace {
-			if t == nil {
-				return WorkflowErrUnknownWorkflowStep()
-			} else if !t.Enabled {
-				return WorkflowErrDisabled()
-			}
-		}
-
-		if t != nil {
-			wap.setTrigger(t)
-			p.StepID = t.StepID
-			p.EventType = t.EventType
-			p.ResourceType = t.ResourceType
-
-			// merge with input from trigger
-			// with trigger input vars are overwritten by input vars
-			p.Input = t.Input.MustMerge(p.Input)
-		} else {
-			p.EventType = "onTrace"
 		}
 
 		wait, sessionID, err = svc.exec(ctx, wf, p)
--- a/automation/service/workflow_converter.go
+++ b/automation/service/workflow_converter.go
@@ -15,15 +15,18 @@ import (
 )
 
 type (
+	wfExecutor interface {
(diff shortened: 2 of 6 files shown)


automation/service/workflow.go
```diff
@@ -564,6 +564,11 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
 		// Find the trigger.
 		// @todo can we cache this as well?
 		t, err = func() (*types.Trigger, error) {
+			if p.CallerWorkflowID > 0 {
+				// skip triggers checking when executed as sub-workflow
+				// @todo be more strict and allow this ONLY when workflow is flagged as a sub-workflow
+			}
```
tjerman, line 570: dead code

automation/service/workflow.go
```diff
@@ -572,6 +577,10 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
 				return nil, err
 			}
 
+			if len(tt) == 0 {
+				return nil, nil
+			}
```
tjerman, line 582: perhaps move the above comment about skipping triggers to here and elaborate a bit more on why

automation/service/workflow_converter.go
```diff
@@ -506,6 +512,92 @@ func (svc workflowConverter) convContinueStep() (wfexec.Step, error) {
...
+
+			// optional stepID, follows login in Exec when 0
+			ap.uint64(argNameStepID, &p.StepID)
+
+			// scope for the sub-workflow
+			ap.vars(argNameScope, p.Input)
+		}
+
+		if workflowID > 0 && workflowHandle != "" {
+			return nil, fmt.Errorf("cannot provide both workflowID and workflowHandle")
```
tjerman, line 576: Why not? I'd either prioritize the ID or allow it if the handle points to the same ID (this one is just me being annoying :1st_place_medal: )

automation/service/workflow_converter.go
```diff
@@ -506,6 +512,92 @@ func (svc workflowConverter) convContinueStep() (wfexec.Step, error) {
...
+		}
+
+		if workflowID == 0 {
+			if workflowHandle == "" {
+				return nil, fmt.Errorf("workflowID or workflowHandle must be provided")
+			}
+
+			workflowID = svc.graphs.handleToID(workflowHandle)
+		}
```
tjerman, line 585: This might be a bit easier on the eyes (I prefer this form)

```go
		if workflowID == 0 {
			workflowID = svc.graphs.handleToID(workflowHandle)
               }
		if workflowID == 0 {
                        return nil, fmt.Errorf("workflowID or workflowHandle must be provided")
		}
```

automation/service/workflow_step_args.go
```diff
@@ -0,0 +1,71 @@
...
+	//
+	// when initialised it evaluates expressions using given scope
+	// it provides methods to extract values from the result
+	argsProc struct {
+		expr   types.ExprSet
+		result *expr.Vars
+	}
+)
+
+func processArguments(ctx context.Context, expr []*types.Expr, scope *expr.Vars) (p *argsProc, err error) {
```
tjerman, line 20: Perhaps this would be cleanner

```go
func processArguments(ctx context.Context, scope *expr.Vars, expr ...*types.Expr)
```

tjerman: Primarily if no exprs are provided; is this even possible?

automation/service/workflow_step_args.go
```diff
@@ -0,0 +1,71 @@
...
+
+type (
+	// argsProc is a helper to process arguments
+	//
+	// when initialised it evaluates expressions using given scope
+	// it provides methods to extract values from the result
+	argsProc struct {
+		expr   types.ExprSet
+		result *expr.Vars
+	}
```
tjerman, line 17: would it make sense to add some tests for this?

automation/service/workflow.go
```diff
@@ -582,32 +591,32 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
...
 
-			return nil, nil
-		}()
+			if !p.Trace {
+				if t == nil {
+					return nil, WorkflowErrUnknownWorkflowStep()
+				} else if !t.Enabled {
+					return nil, WorkflowErrDisabled()
+				}
+			}
```
tjerman, line 600: I'm not sure I understand why this bit exists?
Is this correlated to the exec workflow call where the `p.Trace` is set to `false`? If so, I'm not a fan

tjerman: Should we have another parameter to indicate? We can't use the stack (yet).
Perhaps just add some notes why this is here and we'll solve it later

automation/service/workflow.go
```diff
@@ -582,32 +591,32 @@ func (svc *workflow) Exec(ctx context.Context, workflowID uint64, p types.Workfl
...
-		if !p.Trace {
-			if t == nil {
-				return WorkflowErrUnknownWorkflowStep()
-			} else if !t.Enabled {
-				return WorkflowErrDisabled()
+				// merge with input from trigger
+				// with trigger input vars are overwritten by input vars
+				p.Input = t.Input.MustMerge(p.Input)
+			} else {
+				p.EventType = "onTrace"
```
tjerman, line 612: Can we move to a constant? Magic numbers but as a string :eyes:

tjerman (changes requested):
Less code, more comments :+1:

automation/service/workflow_step_args.go
darh: a) not possible b) expressions are handled first (and are more important) and scope after that; and this dictates order of fn. args for me.

automation/service/workflow_step_args.go
darh: It would make a lot of sense yes. 
And since this is a utility funct/struct it does not make sense (for me) to write tests so early

automation/service/workflow.go
darh: it would, but this is an old code and not relevant to the PR.

PR #409, opened 2022-09-16
Resource pagination

vicpatel:
Add IncTotal support for all resource

Updates rdbms.go.tpl to extend store search method to include total in response

Please take a look.

### DIFF
--- a/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
+++ b/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
@@ -162,6 +162,26 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
 	if err != nil {
 		return nil, f, err
 	}
+
+	if f.IncTotal {
+		// Calc total from the number of items fetched
+		// even if we do build the page navigation
+		f.Total = uint(len(set))
+
+		if f.Limit > 0 && uint(len(set)) == f.Limit {
+			// there are fewer items fetched then requested limit
+			limit := f.Limit
+			f.Limit = 0
+			var navSet {{ .goSetType }}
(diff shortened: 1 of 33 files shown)


codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
```diff
@@ -162,6 +162,25 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
 	if err != nil {
 		return nil, f, err
 	}
+
+    // @fixme condition per resource
```
darh: can you elaborate a bit (in the comment). Just so we do not forget.

codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
```diff
@@ -162,6 +162,25 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
...
+	if f.IncTotal {
```
darh, line 166: fix ident please

codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
```diff
@@ -162,6 +162,25 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
...
+        // Calc total from the number of items fetched
+        // even if we do build the page navigation
+        f.Total = uint(len(set))
+    }
+    if f.Limit > 0 && uint(len(set)) == f.Limit {
```
darh: change this as agreed (move inside the block above and send in proper limit value).

darh (changes requested):

darh (approved):

PR #411, opened 2022-09-26
Fix total count for record list

vicpatel:
Please take a look.

### DIFF
--- a/compose/dalutils/records.go
+++ b/compose/dalutils/records.go
@@ -2,8 +2,6 @@ package dalutils
 
 import (
 	"context"
-	"math"
-
 	"github.com/cortezaproject/corteza-server/compose/types"
 	"github.com/cortezaproject/corteza-server/pkg/dal"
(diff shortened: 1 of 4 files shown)


compose/dalutils/records.go
```diff
@@ -198,6 +201,72 @@ func drainIterator(ctx context.Context, iter dal.Iterator, mod *types.Module, f
...
+				if total == howMuchMore {
+					// fetched += total
+					total = 0
+					if err = iter.More(howMuchMore, r); err != nil {
+						return
+					}
+				}
+			}
+
+			// total = 0
```
darh: we probably don't need this commented-out section?

darh (changes requested):

tjerman (commented):
lgtm; there are some commented out variables in case you forgot about those

The other ticket (not seeing the full paging cursor list) will also need work on similar places, so worth doing both in one go?

PR #420, opened 2022-10-10
Reload only existing route on save

vicpatel:
Adds reloadByEndpoint to refresh only saved/update route as needed and for the routes no longer exist we attach the 404 handler.

Please take a look.

### DIFF
--- a/pkg/apigw/service.go
+++ b/pkg/apigw/service.go
@@ -134,6 +134,52 @@ func (s *apigw) Reload(ctx context.Context) (err error) {
 	return nil
 }
 
+// ReloadEndpoint reload a route and its filters
+//
+// The procedure use existing chi mux
+func (s *apigw) ReloadEndpoint(ctx context.Context, endpoint string) (err error) {
+	var (
+		routes []*route
+	)
+
+	routes, err = s.loadRoutes(ctx, endpoint)
+
+	if err != nil {
+		s.log.Error("could not reload Integration Gateway routes", zap.Error(err))
+		return
+	}
+
+	// rr := append(s.routes, routes...)
+	s.PrepRoutes(ctx, routes...)
+
+	if s.mx == nil {
+		// Rebuild the mux
+		s.mx = chi.NewMux()
+	}
+
+	for _, r := range routes {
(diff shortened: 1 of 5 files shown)


system/service/apigw_filter.go
```diff
@@ -18,15 +17,17 @@ type (
 		store     store.Storer
 		ac        routeAccessController
 		route     *apigwRoute
+		ags       apiGwService
 	}
 )
 
-func Filter() *apigwFilter {
+func Filter(ags apiGwService) *apigwFilter {
```
petergrlica: Why use the service as a new fields, since the apigw service is a singleton and can be used as before (`.Service()`)?

system/service/service.go
```diff
@@ -46,6 +50,17 @@ type (
 		WaitFor(ctx context.Context, ev eventbus.Event) (err error)
 		Dispatch(ctx context.Context, ev eventbus.Event)
 	}
+
+	apiGwService interface {
```
petergrlica: We are not using this in unit tests, so why not keep the global service as it was?

pkg/apigw/service.go
```diff
@@ -134,6 +134,52 @@ func (s *apigw) Reload(ctx context.Context) (err error) {
 	return nil
 }
 
+// ReloadByEndpoint reload a route and its filters
+//
+// The procedure use existing chi mux
+func (s *apigw) ReloadByEndpoint(ctx context.Context, endpoint string) (err error) {
```
petergrlica: Maybe `ReloadEndpoint`?

system/service/apigw_filter.go
vicpatel: Yeah, I agree I can revert the changes.
The reason which leads me to this is when I had to use service multiple times in same method then either I will have to create a local instance of service in the method or make service available to the struct but we also have 3 `apigw_*` structs hence I created an interface for apigw service and aligned it with rest of the implementations.

system/service/service.go
vicpatel: I hope this will be resolved with the above reply :)

PR #421, opened 2022-10-11
fix(pkg/expr/func_json.go): fix marshal expr types to empty object bug

ppaanngggg:
`toJSON` func in expr marshal expr types, I think it shoud marshal value of expr types.

### DIFF
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -32,7 +32,7 @@ jobs:
         channel: ${{ secrets.MATRIX_ROOM_ID }}
         message: |
           ### [ ${{ github.event.repository.name }} ]
-          An issue has been **${{ github.event.action }}** by ${{ github.event.issue.user.login }}\
+          An issue has been **${{ github.event.action }}** by ${{ github.triggering_actor }}\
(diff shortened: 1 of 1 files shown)


darh:
Thanks for the patch.

I'm having trouble understanding what problem are you solving?
Can you add a test case or two (there is a test function already prepared for this) that covers this fix?

pkg/expr/func_json_test.go
```diff
@@ -17,3 +17,21 @@ func Example_toJSON() {
...
+					"k2": "v2",
+				}},
+			)),
+		}
+	)
+
+	eval(`toJSON(kv)`, p)
+
+	// output:
+	// {"k1":"v1","k2":"v2"}
```
ppaanngggg: before my fix, the output will be `{}`

darh:
Ok, confirming this. Thank you @ppaanngggg!

Can I ask you to a) squash into on commit and b) reword commit message to follow guidelines in the [CONTRIBUTIONS.md](https://github.com/cortezaproject/corteza-server/blob/2022.9.x/CONTRIBUTING.md).

I'll rebase & merge after that.

Issue #422, opened 2022-10-11
Can not export compose records with CLI


Yangruipis:
**Describe the bug**

When exporting the records to yaml using command line tool, it exited with "not found" error, and no records is exported

**To Reproduce**

1. create one module named `test` and insert some records
2. run command line tool:

```bash
./bin export -o export_dir compose:namespace:test compose:module: compose:record:
```

**Expected behavior**

cli exit zero and records is expored.

**Screenshots**

![image](https://user-images.githubusercontent.com/16623724/195019881-ddb99739-a817-4f55-bd21-92591b19f660.png)

**Additional context**

- branch: 2022.9.x
- commit: 57721291b270d4ed9b9e625d7a17fca2cc6c27eb

**My debug results**

1. `f.ModuleID` here is zero, which raises "not found"

https://github.com/cortezaproject/corteza-server/blob/57721291b270d4ed9b9e625d7a17fca2cc6c27eb/pkg/envoy/store/compose.go#L235-L237

2. after I force set the `f.ModuleID`, I got error:

```
yaml encoder encode: build doc corteza  compose record [305723673694711809]: cannot search data entry: model 305723673694711809 does not exist
```

this is because dal service.models is not loaded before the export process

3. also, rbac is not initialized here and null pointer error will be raised

https://github.com/cortezaproject/corteza-server/blob/57721291b270d4ed9b9e625d7a17fca2cc6c27eb/pkg/envoy/store/compose.go#L256

Yangruipis:
My pr is here: #423 , please take a look @tjerman thanks

Yangruipis:
fixed

PR #423, opened 2022-10-11
fix: export record error && dal.models is not initialized

Yangruipis:
## Desc

related to #422 

### fixes:

1. get right module id when export records
2. init dal.models when running import/export


pkg/envoy/store/compose.go
```diff
@@ -187,7 +187,17 @@ func (d *composeDecoder) decodeComposeRecord(ctx context.Context, s store.Storer
...
-				ffNs = append(ffNs, &fNs)
+
+				nn, _, err := s.SearchComposeModules(ctx, types.ModuleFilter{
+					NamespaceID: nsID,
+				})
+				if err != nil {
+					continue
+				}
+				for _, n := range nn {
+					fNs.ModuleID = n.ID
```
tjerman, line 197: Won't this only include the last module in the slice? Shouldn't we copy the `fNs`?

tjerman (changes requested):

pkg/envoy/store/compose.go
Yangruipis: yes you are right, it is fixed now

Yangruipis:
@tjerman this pr is updated, plz take a look, thanks

tjerman:
Lgtm

Issue #427, opened 2022-10-13
Required Field Attribute On Module Not Respected For Physical Tables


skamensky:
**Describe the bug**
When setting a field to "not required", it still received a `not null` constraint in postgres

**To Reproduce**
1. Create a field that's not required
2. Try to create a record without populating the 'not required' field
3. Get DB error

**Expected behavior**
`not required` translated to no `not null` constraint being applied to the column

**Screenshots**
Definition of the module (I also tried checking and unchecking before the initial saving of the module, to no effect)
![image](https://user-images.githubusercontent.com/19151369/195666274-62303fac-7ef8-40c1-b1f5-de2bd5303d5f.png)

**Additional context**
Generated definition (retreived from goland IDE)
```
-- auto-generated definition
create table "Parent"
(
    "Name"        text         not null,
    id            bigint       not null,
    rel_module    bigint       not null,
    deleted_by    bigint,
    rel_namespace bigint       not null,
    revision      bigint       not null,
    meta          jsonb        not null,
    owned_by      bigint       not null,
    created_at    timestamp(0) not null,
    created_by    bigint       not null,
    updated_at    timestamp(0),
    updated_by    bigint,
    deleted_at    timestamp(0)
);```

skamensky:
I created a fix for this here

https://github.com/cortezaproject/corteza-server/pull/431

@darh , can you please take a look?

skamensky:
Solved with https://github.com/cortezaproject/corteza-server/pull/431

darh:
I'll keep this one open for now. We'll close it when we merge the #443

darh:
Nevermind 🙈

PR #431, opened 2022-10-19
Fix issue #427 (respected required fields)

skamensky:
since ModuleField's attribute "Required" was not explicitly passed as a negated value to Attribute.Type.Nullable, Attribute.Type.Nullable's default value always persisted (false). This made all columns required.

This change makes it so the ModuleField's attribute "Required" properly maps to Attribute.Type.Nullable

Please take a look.

### DIFF
--- a/compose/service/module.go
+++ b/compose/service/module.go
@@ -1479,31 +1479,44 @@ func moduleFieldToAttribute(f *types.ModuleField) (out *dal.Attribute, err error
 
 	switch strings.ToLower(f.Kind) {
 	case "bool", "boolean":
-		at := &dal.TypeBoolean{}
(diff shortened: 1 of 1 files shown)


darh:
Require flag is not directly connected to the not-null flag on attribute. But you could say that required fields should have not-null flag and vice-versa.

@tjerman you have any comments on this

skamensky:
@darh , @tjerman ,

Anything holding this up? I see you released another candidate since this PR was opened. Was hoping to get this out so we can have non required fields...

tjerman:
I do agree with what @darh said; I think this is ok for now. We'll change it if need be

tjerman (approved):

PR #432, opened 2022-10-20
fix: report paging

Yangruipis:
## fixes

1. error of report runs is not captured
2. typo: `___DLMTR___` should be `___DLTR___`, and function `wrapNestedGvalIdent` should be used here


tjerman (approved):
LGTM; thanks for patching it!

PR #433, opened 2022-10-21
Fix marshal expr.Type into empty object bug

ppaanngggg:
When using to_json(...) marshal expr.Type into string will get empty object {} if this Type doesn't implement marshal method. 

So UntypedValue obj before json.Marshal if this obj isn't json.Marshaler.


ppaanngggg:
@darh updated

PR #434, opened 2022-10-21
Fix locale error meta key

vicpatel:
This will convert the locale camelCase error meta key to kebab-case to fix translation reference for converted resources.

Please take a look.


darh:
How does this affect use of locales?

vicpatel:
Resource name will be generated in `kebab-case` as [per the guideline](https://github.com/cortezaproject/corteza-locale/blob/2022.9.x/DEV.md#translation-keys) so we need to make sure we create locale accordingly.

For example, for system/dalConnection, locale directory should be system/dal-connection hence the error reference string should be `dal-connection.errors.{errorName}` instead of `dalConnection.errors.{errorName}`.

PR #435, opened 2022-10-24
fix(automation): session is not created

Yangruipis:
## Problem

Automation session is created when status changes from activate to completed through `store.UpsertAutomationSession(...)` api. It is not created when starting. So we **can not** get this session util it completes or fails.

## Fix

Call `store.CreateAutomationSession(...)` when session starts.

### DIFF
--- a/automation/service/session.go
+++ b/automation/service/session.go
@@ -529,7 +529,7 @@ func (svc *session) stateChangeHandler(ctx context.Context) wfexec.StateChangeHa
 
 		default:
 			// force update on every F new frames (F=sessionStateFlushFrequency) but only when stacktrace is not nil
-			update = ses.RuntimeStacktrace != nil && len(ses.RuntimeStacktrace)%sessionStateFlushFrequency == 0
+			update = ses.RuntimeStacktrace != nil && len(ses.RuntimeStacktrace)%sessionStateFlushFrequency == 1
 		}
 
 		if !update {


darh:
Can you explain the rationale behind this? 
Session is already stored (and updated) in the `stateChangeHandler`

Yangruipis:
> Can you explain the rationale behind this? Session is already stored (and updated) in the `stateChangeHandler`

hi, as i saw from the debug result, 

image

when `stateChangeHandler` was called the first time,  `len(ses.RuntimeStacktrace)=1` and `update=false` so during the top `sessionStateFlushFrequency`(=1000) frames, the session will not be created. 

And it leads to the `session not found` error when trigger the workflow manually：

image

image

Yangruipis:
According to my explaination, this might be a better solution here?

```diff
diff --git a/automation/service/session.go b/automation/service/session.go
index 4363b22a4..056cb2904 100644
--- a/automation/service/session.go
+++ b/automation/service/session.go
@@ -535,7 +535,7 @@ func (svc *session) stateChangeHandler(ctx context.Context) wfexec.StateChangeHa

                default:
                        // force update on every F new frames (F=sessionStateFlushFrequency) but only when stacktrace is not nil
-                       update = ses.RuntimeStacktrace != nil && len(ses.RuntimeStacktrace)%sessionStateFlushFrequency == 0
+                       update = ses.RuntimeStacktrace != nil && len(ses.RuntimeStacktrace)%sessionStateFlushFrequency == 1
                }

                if !update {
```

darh:
Yap. I think that putting it in the handler fn is more consistent than the previous solution. I'm not 100% sure right now if the proposed change is the most optimal one but can review this in a day or two. If you find anything that would work better (inside a hander), feel free to propose. Thank you.

PR #436, opened 2022-10-26
WIP: Fix deletedAt is not initialized correctly 

Yangruipis:
## Fix

1. `DeletedAt` here is not set to nil, which makes member function `IsDeleted` return wrong result

https://github.com/cortezaproject/corteza-server/blob/c710aa079d9b54f46e9a1aa6ac51079030500aec/compose/types/record_value.go#L278-L278

https://github.com/cortezaproject/corteza-server/blob/c710aa079d9b54f46e9a1aa6ac51079030500aec/compose/types/record_value.go#L42-L44

### DIFF
--- a/compose/service/record_test.go
+++ b/compose/service/record_test.go
@@ -482,7 +482,7 @@ func TestRecord_defValueFieldPermissionIssue(t *testing.T) {
 
 		recPartial, _, err = svc.Update(ctx, recPartial)
 		verifyRecErrSet(t, err)
-		req.Equal("<w2><NULL>", valueExtractor(recPartial, "writable", "readable"))
+		req.Equal("<w2><r>", valueExtractor(recPartial, "writable", "readable"))
 	}
 
 	{
--- a/compose/types/record_value.go
+++ b/compose/types/record_value.go
@@ -275,7 +275,7 @@ func (set RecordValueSet) merge(new RecordValueSet) (out RecordValueSet) {
 			Value:     set[s].Value,
 			Ref:       set[s].Ref,
 			Place:     set[s].Place,
-			DeletedAt: &time.Time{},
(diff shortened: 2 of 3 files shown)


darh:
Interesting. Can you take a look at the tests?

There are (surprisingly) two functions: `TestRecordValueSet_Merge` and `TestRecordValueSet_merge`. 

Can you add a case that covers your scenario?
Also -- if you have the courage to join the two functions you'll have my eternal gratitude.

darh:
Another thing -- can I kindly ask you to follow CONTRIBUTIONS.md when it comes to git commit message format.

Yangruipis:
Hi @darh , i'm confused about the update action. As I known, update will just update values if given and changed, and keep the unchanged ones(as known as merge)?   But from the test cases, the expected answers are not results of merge (overwrite maybe). 

So are these cases writen by desgin, or just some mistakes?

- the last answer should be `<w2><r>` ?
    - https://github.com/cortezaproject/corteza-server/blob/55702ec2568f8a34a00fcfaffeaef0341f77ad2c/compose/service/record_test.go#L476-L485
- length of `$.response.values` should be 2 (including field `name` and `email` )
    - https://github.com/cortezaproject/corteza-server/blob/55702ec2568f8a34a00fcfaffeaef0341f77ad2c/tests/compose/record_test.go#L776-L791

darh:
I think you mixed ext and internal operations.

Externally, due to the "laws of REST", we anticipate that client (browser) will always send the full payload and all values (with some exception when it comes to values that user do not have access to, but that is another story).

Internally we use the mentioned value-merge function but not directly and don't only that function.

Hope this makes sense.

Yangruipis:
Wow, thanks for your nice explanation!

PR #439, opened 2022-10-28
Upgrade metrics dependence and middleware

ppaanngggg:
Use my fork of chi-prometheus which fix PatternMiddleware bug and upgrade chi to v5.
Use PatternMiddleware and make password optional.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/SentimensRG/ctx v0.0.0-20180729130232-0bfd988c655d
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/brianvoe/gofakeit/v6 v6.16.0
-	github.com/cortezaproject/corteza-locale v0.0.0-20221026122439-c3b0d9b15a4f
+	github.com/cortezaproject/corteza-locale v0.0.0-20221028080428-f7a520f4dd45
 	github.com/crewjam/saml v0.4.6
 	github.com/crusttech/go-oidc v0.0.0-20180918092017-982855dad3e1
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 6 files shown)


darh:
What would be the reason for not submitting a PR to `766b/chi-prometheus`? I would prefer if we keep that one for now. If there is no response from the repo author and this is not merged in a reasonable time (let's say 1mo after PR is submitted). 

Thoughts?

pkg/api/server/metrics.go
```diff
@@ -1,24 +1,27 @@
...
-	return chiprometheus.NewMiddleware(name)
+	return chiprometheus.NewPatternMiddleware(name)
 }
 
 func metricsMount(r chi.Router, username, password string) {
 	r.Route("/metrics", func(r chi.Router) {
-		r.Use(basicauth.New("Metrics", map[string][]string{
-			username: {password},
-		}))
+		if username != "" && password != "" {
```
darh: You said in the PR description that you are making password optional?
How would this work if you want to use username but password is empty?

ppaanngggg: I want to enable basicauth only when `username` and `password` set

ppaanngggg:
> What would be the reason for not submitting a PR to `766b/chi-prometheus`? I would prefer if we keep that one for now. If there is no response from the repo author and this is not merged in a reasonable time (let's say 1mo after PR is submitted).
> 
> Thoughts?

Good idea, I create a PR to `chi-prometheus` first.

pkg/api/server/metrics.go
darh: TBH I have no knowledge about Prometheus' auth constraints, what is required or possible. 
But your comment about making password optional and the code are incompatible :)

PR #440, opened 2022-10-28
Fix fetch all records when no dedup rules

ppaanngggg:
Current dedup config of module is always enable (as front-end comment).

So the better way to control dedup is rules length.


vicpatel:
Thank you for that :)

**Note:**  We have plan to remove `Enabled` from the configuration struct(ModuleConfigRecordDeDup) in next deDup update.

vicpatel (approved):

PR #441, opened 2022-10-28
Support to list records filter by ref module fields by left join

ppaanngggg:
Left join with ref modules as filed name, then query language can use "field"."ref-model field" to filter data.

### DIFF
--- a/store/adapters/rdbms/dal/connection.go
+++ b/store/adapters/rdbms/dal/connection.go
@@ -171,7 +171,7 @@ func (c *connection) CreateModel(ctx context.Context, mm ...*dal.Model) (err err
 		}
 
 		// cache the model
-		c.models[cacheKey(m)] = Model(m, c.db, c.dialect)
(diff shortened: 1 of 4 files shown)


store/adapters/rdbms/dal/connection.go
```diff
@@ -230,7 +230,7 @@ func (c *connection) UpdateModelAttribute(ctx context.Context, sch *dal.Model, o
 }
 
 func cacheKey(m *dal.Model) (key string) {
-	key = m.ResourceType + "|" + m.Resource + "|" + m.Ident
+	key = fmt.Sprintf("%s|%v", m.ResourceType, m.ResourceID)
```
tjerman, line 233: Why should this change?

tjerman (changes requested):
Can you please add some test coverage for this?

ppaanngggg:
There are some conflicts after your refactors. I will solve them, add tests and reopen this. @tjerman

PR #442, opened 2022-10-30
Add beta column modification

skamensky:
Is is a mistake that 
https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/store/adapters/rdbms/dal/connection.go#L154-L178
calls `UpdateModel`? Why would a creation call an update? As part of this PR I've removed the call to UpdateModel and pulled up potential errors into the CreateModel function. This is not necessary for the PR.

Why does `rdbms/ddl/dal_model.go:UpdateModel` iterate over model attributes instead of

https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/store/adapters/rdbms/ddl/dal_model.go#L47-L55

`rdbms.dal/connection.go:UpdateModelAttribute`?

pkg/dal/driver.go:UpdateModel note says not to update attributes, only metadata.
https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/pkg/dal/driver.go#L80

 However it looks like `store/adapters/rdbms/ddl/dal_model.go:UpdateModel` is preparing to do just that.
Is this just an unfortunate naming coincidence? I see that the dal_model:UpdateModel is not part of the implementation of the aforementioned UpdateModel's interface.

Also `store/adapters/rdbms/dal/connection.go:UpdateModel` which *is* part of the interface that says not to use UpdateModel to update attributes has a todo saying "check if column exists and if it can be removed".
https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/store/adapters/rdbms/dal/connection.go#L223-L230
 Why would it need to do that if it's not supposed to update attributes?
 
Why is there dal specific code in compose/service/module.go ?
What does `Asserted` mean in pkg/dal/diff.go? Was that meant to be "Inserted"?
https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/pkg/dal/diff.go#L12

store/adapters/rdbms/dal/connection.go:UpdateModelAttribute says it's not raising unsupported errors since it doesn't want to break "DAL service model adding procedure". What is it referring to by the adding procedure?

What tests would you like?
How do you do sqlite testing? It doesn't save the data inside the file I give it

I've tested manually on mysql,postgres and sqlite.

Tests I've completed:
1. Add column
2. Delete column
3. Attempt to add required column with records and get error
4. Attempt to modify existing physical column and get error
5. Attempt to change schema of compose_record and get error
6. Attempt to modify json type column and succeed  
4. Attempt to make any physical changes when `DB_ALLOW_DESTRUCTIVE_CHANGES` is not set and get a not-supported error

### DIFF
--- a/app/boot_dal.go
+++ b/app/boot_dal.go
@@ -41,7 +41,9 @@ func (app *CortezaApp) initDAL(ctx context.Context, log *zap.Logger) (err error)
 	}
 
 	// Init DAL and prepare default connection
-	dal.SetGlobal(dal.New(log.Named("dal"), app.Opt.Environment.IsDevelopment()))
+    isDev:= app.Opt.Environment.IsDevelopment()
+    allowDestructiveChanges:= app.Opt.DB.ALLOWDESTRUCTIVECHANGES
+	dal.SetGlobal(dal.New(log.Named("dal"),isDev, allowDestructiveChanges))
 	if err = dal.Service().ReplaceConnection(ctx, cw, true); err != nil {
 		return fmt.Errorf("could not set primary connection: %w", err)
 	}
--- a/app/boot_dal.go
+++ b/app/boot_dal.go
@@ -41,9 +41,7 @@ func (app *CortezaApp) initDAL(ctx context.Context, log *zap.Logger) (err error)
 	}
 
 	// Init DAL and prepare default connection
-	isDev := app.Opt.Environment.IsDevelopment()
-	allowDestructiveSchemaChanges := app.Opt.DB.AllowDestructiveSchemaChanges
-	dal.SetGlobal(dal.New(log.Named("dal"), isDev, allowDestructiveSchemaChanges))
+	dal.SetGlobal(dal.New(log.Named("dal"), app.Opt.Environment.IsDevelopment()))
 	if err = dal.Service().ReplaceConnection(ctx, cw, true); err != nil {
 		return fmt.Errorf("could not set primary connection: %w", err)
 	}
--- a/app/boot_dal.go
+++ b/app/boot_dal.go
@@ -41,9 +41,9 @@ func (app *CortezaApp) initDAL(ctx context.Context, log *zap.Logger) (err error)
 	}
 
 	// Init DAL and prepare default connection
-    isDev:= app.Opt.Environment.IsDevelopment()
-    allowDestructiveChanges:= app.Opt.DB.ALLOWDESTRUCTIVECHANGES
-	dal.SetGlobal(dal.New(log.Named("dal"),isDev, allowDestructiveChanges))
+	isDev := app.Opt.Environment.IsDevelopment()
+	allowDestructiveSchemaChanges := app.Opt.DB.AllowDestructiveSchemaChanges
+	dal.SetGlobal(dal.New(log.Named("dal"), isDev, allowDestructiveSchemaChanges))
 	if err = dal.Service().ReplaceConnection(ctx, cw, true); err != nil {
 		return fmt.Errorf("could not set primary connection: %w", err)
 	}
--- a/app/options/db.cue
+++ b/app/options/db.cue
@@ -12,7 +12,7 @@ DB: schema.#optionsGroup & {
 			defaultValue: "sqlite3://file::memory:?cache=shared&mode=memory"
 			description:  "Database connection string."
 		}
-		ALLOW_DESTRUCTIVE_CHANGES: {
+		allow_destructive_schema_changes: {
 			type:          "bool"
 			defaultGoExpr: "false"
 			description:   "Allow for irreversible changes to the database schema such as dropping columns and tables."
--- a/app/options/db.cue
+++ b/app/options/db.cue
@@ -12,6 +12,11 @@ DB: schema.#optionsGroup & {
 			defaultValue: "sqlite3://file::memory:?cache=shared&mode=memory"
 			description:  "Database connection string."
 		}
+		ALLOW_DESTRUCTIVE_CHANGES: {
+			type:          "bool"
+			defaultGoExpr: "false"
+			description:   "Allow for irreversible changes to the database schema such as dropping columns and tables."
+		}
 	}
 	title: "Connection to data store backend"
 }
--- a/compose/service/dal_interfaces.go
+++ b/compose/service/dal_interfaces.go
@@ -12,7 +12,7 @@ type (
 		SearchModels(ctx context.Context) (out dal.ModelSet, err error)
 		ReplaceModel(ctx context.Context, model *dal.Model) (err error)
 		RemoveModel(ctx context.Context, connectionID, ID uint64) (err error)
-		ReplaceModelAttribute(ctx context.Context, model *dal.Model, diff *dal.ModelDiff,hasRecords bool, trans ...dal.TransformationFunction) (err error)
+		ReplaceModelAttribute(ctx context.Context, model *dal.Model, diff *dal.ModelDiff, hasRecords bool, trans ...dal.TransformationFunction) (err error)
 
 		GetConnectionByID(uint64) *dal.ConnectionWrap
 
--- a/compose/service/dal_interfaces.go
+++ b/compose/service/dal_interfaces.go
@@ -12,7 +12,7 @@ type (
 		SearchModels(ctx context.Context) (out dal.ModelSet, err error)
 		ReplaceModel(ctx context.Context, model *dal.Model) (err error)
 		RemoveModel(ctx context.Context, connectionID, ID uint64) (err error)
-		ReplaceModelAttribute(ctx context.Context, model *dal.Model, old, new *dal.Attribute, trans ...dal.TransformationFunction) (err error)
+		ReplaceModelAttribute(ctx context.Context, model *dal.Model, diff *dal.ModelDiff,hasRecords bool, trans ...dal.TransformationFunction) (err error)
 
 		GetConnectionByID(uint64) *dal.ConnectionWrap
 
--- a/pkg/dal/diff.go
+++ b/pkg/dal/diff.go
@@ -59,6 +59,10 @@ func (a *Model) Diff(b *Model) (out ModelDiffSet) {
 	// Deleted and update ones
 	for _, _attrA := range a.Attributes {
 		attrA := _attrA
+		// store is an interface to something that could be a pointer.
+		// we need to copy it to make sure we don't get a nil pointer
+		// make sure not to modify this since it would modify the original
+		attrA.Store = _attrA.Store
 
 		// Missmatches
 		attrBAux, ok := bIndex[attrA.Ident]
@@ -91,7 +95,6 @@ func (a *Model) Diff(b *Model) (out ModelDiffSet) {
 				Inserted:     attrBAux.attr,
 			})
 		}
-
 		if attrA.Store.Type() != attrBAux.attr.Store.Type() {
 			out = append(out, &ModelDiff{
 				Type:         AttributeCodecMismatch,
--- a/pkg/dal/diff.go
+++ b/pkg/dal/diff.go
@@ -1,24 +1,31 @@
 package dal
 
 type (
-	modelDiffType string
-
+	modelDiffType     string
+    ModelModification string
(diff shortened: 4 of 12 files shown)


skamensky:
I see I have some tests to modify. I'd be interested in getting if I'm going in the right direction on this PR before fixing those.

skamensky:
Hey @darh ,

I'm travelling over the next 1.5 days so I'll have time to put fixes in if it gets reviewed today.

darh:
Ack, will try to review all of them in next couple of hours. Travel safe:)

app/options/db.cue
```diff
@@ -12,6 +12,11 @@ DB: schema.#optionsGroup & {
 			defaultValue: "sqlite3://file::memory:?cache=shared&mode=memory"
 			description:  "Database connection string."
 		}
+		ALLOW_DESTRUCTIVE_CHANGES: {
```
darh: Use lower case here (allow_destructive_changes), this will properly generate golang variables in camel case and keep the ENV var in all-caps snake case.

And I think it would be more appropriate if we are explicit about what kind of changes are we talking about
`allow_destructive_schema_changes`?

darh:
> Is is a mistake that
> 
> https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/store/adapters/rdbms/dal/connection.go#L154-L178
> 
> 
> calls `UpdateModel`? Why would a creation call an update? As part of this PR I've removed the call to UpdateModel and pulled up potential errors into the CreateModel function. This is not necessary for the PR.

This looks like a mistake, yes.




> Why does `rdbms/ddl/dal_model.go:UpdateModel` iterate over model attributes instead of
> 
> https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/store/adapters/rdbms/ddl/dal_model.go#L47-L55
> 
> `rdbms.dal/connection.go:UpdateModelAttribute`?

The reason for the current state of RDBMS schema manipulation is that I only wanted to satisfy the very basic requirements needed by initial DB provisioning. 

In the end, RDBMS implementation should follow the rules. The current implementation in the rdbms/dal can be changed.

This comment also applies the rest of comments and questions below.


> Why would it need to do that if it's not supposed to update attributes?
> Why is there dal specific code in compose/service/module.go ? What does `Asserted` mean in pkg/dal/diff.go? Was that meant to be "Inserted"?
> 
> https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/pkg/dal/diff.go#L12
> 
> store/adapters/rdbms/dal/connection.go:UpdateModelAttribute says it's not raising unsupported errors since it doesn't want to break "DAL service model adding procedure". What is it referring to by the adding procedure?

Another one of the not-really-implemented RDBMS/dal things. It was just a waste of time to do it since we only needed it for the initial provisioning.

With "not-implemented", I actually meant "not-existing" 🤷 

You'll see that we are reading the models as soon as conn. is established. (`registerModelToConnection`). 

> 
> What tests would you like? How do you do sqlite testing? It doesn't save the data inside the file I give it
> 
> I've tested manually on mysql,postgres and sqlite

Proper DAL testing is somehow incomplete ATM.

I want to have a test suite under tests/dal that can run generic tests on a specific dal connection (provided to tests via DB_DSN). 


> 
> Tests I've completed:
> 
> 1. Add column
> 2. Delete column
> 3. Attempt to add required column with records and get error
> 4. Attempt to modify existing physical column and get error
> 5. Attempt to change schema of compose_record and get error
> 6. Attempt to modify json type column and succeed
> 7. Attempt to make any physical changes when `DB_ALLOW_DESTRUCTIVE_CHANGES` is not set and get a not-supported error

👍 

If provisioning of an empty DB still works than that's it.

----

I'll include a couple of comments to the PR file changes directly.

app/boot_dal.go
```diff
@@ -41,7 +41,9 @@ func (app *CortezaApp) initDAL(ctx context.Context, log *zap.Logger) (err error)
 	}
 
 	// Init DAL and prepare default connection
-	dal.SetGlobal(dal.New(log.Named("dal"), app.Opt.Environment.IsDevelopment()))
+    isDev:= app.Opt.Environment.IsDevelopment()
```
darh: Apply `fmt` to changed files. I'm pretty sure your code editor can be configured to automatically run it on save.

pkg/dal/diff.go
```diff
@@ -93,13 +112,30 @@ func (a *Model) Diff(b *Model) (out ModelDiffSet) {
...
 			continue
 		}
 	}
 
 	return
 }
+
+// DiffAsMap calculates the diff between models a and b where a is used as base.
+// The output is a map that represents the differences as ModelDiffSet's
+func (a *Model) DiffAsMap(b *Model) (out ModelDiffMap) {
```
darh: Do we really need a map for this?

How about `DiffType(ModelModification)` that only returns compatible changes? Or extending `Diff` fn arguments with `...ModelModification`?

pkg/dal/service.go
```diff
@@ -833,12 +835,14 @@ func (svc *service) removeModelFromRegistry(model *Model) {
...
 //
 // We rely on the user to provide stable and valid attribute definitions.
-func (svc *service) ReplaceModelAttribute(ctx context.Context, model *Model, old, new *Attribute, trans ...TransformationFunction) (err error) {
+func (svc *service) ReplaceModelAttribute(ctx context.Context, model *Model, diff *ModelDiff, hasRecords bool,trans ...TransformationFunction) (err error) {
 	svc.logger.Debug("updating model attribute", zap.Uint64("model", model.ResourceID))
 
 	var (
 		conn   *ConnectionWrap
 		issues = newIssueHelper().addModel(model.ResourceID)
+        _new   = diff.Asserted
```
darh: Find a better name that does not require a `_` prefix. 
TBH I don't mind using the full name (diff.Asserted and diff.Original).

compose/service/dal_interfaces.go
```diff
@@ -12,7 +12,7 @@ type (
 		SearchModels(ctx context.Context) (out dal.ModelSet, err error)
 		ReplaceModel(ctx context.Context, model *dal.Model) (err error)
 		RemoveModel(ctx context.Context, connectionID, ID uint64) (err error)
-		ReplaceModelAttribute(ctx context.Context, model *dal.Model, old, new *dal.Attribute, trans ...dal.TransformationFunction) (err error)
+		ReplaceModelAttribute(ctx context.Context, model *dal.Model, diff *dal.ModelDiff,hasRecords bool, trans ...dal.TransformationFunction) (err error)
```
darh: Could this `hasRecords` be a part of `ModelDiff`? 
Could we imply that records exist by inspecting the length of the trans slice?
Could trans be part of ModelDiff?

I'm pretty sure @tjerman will have a comment on this as well

store/adapters/rdbms/dal/connection.go
```diff
@@ -220,12 +227,66 @@ func (c *connection) UpdateModel(ctx context.Context, old *dal.Model, new *dal.M
 }
 
 // UpdateModelAttribute alters column on a db table and runs data transformations
-func (c *connection) UpdateModelAttribute(ctx context.Context, sch *dal.Model, old, new *dal.Attribute, trans ...dal.TransformationFunction) error {
-	// not raising not-supported error
-	// because we do not want to break
-	// DAL service model adding procedure
-
-	// @todo implement model column altering
+func (c *connection) UpdateModelAttribute(ctx context.Context, sch *dal.Model,diff *dal.ModelDiff,allowDestructiveChanges bool,hasRecords bool, trans ...dal.TransformationFunction) error {
```
darh: Is this fn sign compatible with UpdateModelAttribute on the interface?
Allowing destructive changes is an RDBMS "feature", Do we want DAL to be burdened with it?

skamensky: > Is this fn sign compatible with UpdateModelAttribute on the interface?

It is now because I changed the interface.

This makes sense. I will remove `allowDestructiveChanges` from this signature and its interface and I'll add a private helper in rdbmsdal/connection.go

compose/service/dal_interfaces.go
skamensky: Semantically I would say neither "having records" nor "a transformation function" are part of a model diff.
We do need those things to run the "replace model attribute" action. OTOH it does look to me like the ReplaceModelAttribute is getting crowded. I'll wait for @tjerman's feedback before changing anything here.

app/options/db.cue
skamensky: I'll make this change

app/boot_dal.go
skamensky: I'll apply `fmt` on all changed files. I added it to my code editor for future PR's.

pkg/dal/diff.go
skamensky: You're right. It's not needed. I'll remove it.

pkg/dal/service.go
skamensky: I'll use the full name everwhere

store/adapters/rdbms/dal/connection.go
skamensky: @darh 
Actually, I just realized that we already put `allowDestructiveChanges` on the dal level when in boot_dal we set the flag on the service. It looks like there's no way currently to set per storage-type flags since dal.New doesn't have any generic way of setting flags.

Should I add a `map[string]bool` type on the dal service so we can have different options per storage type?

skamensky: Or would it make more sense to set that on the connection/connectionWrap?

skamensky: In the end I kept this in the service and did the appropriate checks there

darh: I'm ok if you call `options.DB()` directly from the driver. When we discussed this whole thing the other, we said that this is a temporary (🤞 ) solution anyway, and when we address it properly (with the user having control via UI) this will all be deprecated.

Just add a line or two of comment when you access this option directly + maybe a note on the option description itself.

Thoughts?

skamensky:
I've made the changes requested. 
I also tested that provisioning works on empty DB's for  mysql,postgres, and sqlite.
I'll fix the broken tests later today or tomorrow.

Some unanswered questions from my original PR comment:

Can I change from Assert to Insert?[^1]
How do you check test sqlite data? It's in memory and I can't see the data. Are you using shared memory or something like that to test?

[^1]: https://github.com/cortezaproject/corteza-server/blob/865ee431c15f4f360dd7f1f87770aec2b479e75a/pkg/dal/diff.go#L12

darh:
> Some unanswered questions from my original PR comment:
> 
> Can I change from Assert to Insert?[1](#user-content-fn-1-7f200c915df95cea0dcea2a6dd5e56e7) 

Sure.

> How do you check test sqlite data? It's in memory and I can't see the data. Are you using shared memory or something like that to test?

DB_DSN=sqlite3://path-to-db.db

store/adapters/rdbms/dal/connection.go
skamensky: Yeah, that makes it a lot cleaner. Changed to do that.

skamensky:
@darh ,

Sorry for the long hiatus. I was travelling for a week and a half and had less time than I'd imagined. I'm back to it full force for the next month though.

I see that you migrated to a mono-repo structure. This is great! 

Besides for the conflicts to resolve, are there any other changes I should make given the new repo architecture?

darh:
Glad to have you back 👋 

> Besides for the conflicts to resolve, are there any other changes I should make given the new repo architecture?

Not really. There is a bash script on the root that should help you migrate your branch if you care about your commits.

We've just migrated, and there might be things we overlooked. Patience, please :)

skamensky:
Patience you will have! I was thinking of making some cross repo PR's since one of the changes I had in mind was frontend but I was too overwhelmed. This will ease the process of making full stack changes by a landslide. Great job guys!

PR #443, opened 2022-10-30
Add support for dropdowns in physical tables

skamensky:
Add support in mysql and postgres type converts.
Sqlite doesn't need anything. It uses text for everything.

Please take a look.


skamensky:
While I was at it I checked to make sure no other types were missing. Only `dal.TypeEnum` was missing a conversion case in each dialect's `AttributeToColumn` function

darh:
Looks ok, SQLite probably needs the same change, right?

skamensky:
Yes, I missed that. Fixed. Is this good to go or is there something else that should happen first?

PR #444, opened 2022-11-01
Upgrade metrics dependence and middleware

ppaanngggg:
Use my fork of chi-prometheus which fix PatternMiddleware bug, upgrade chi to v5 and prometheus to latest. Use NewPatternMiddleware(...) instead of NewMiddleware(...). Disable auth when password is empty.

I open PR to original repo https://github.com/766b/chi-prometheus/pull/11 now.
And edit code and commit message as `web-console`.


darh:
Great. Let's wait then until they accept the PR.

ppaanngggg:
> Great. Let's wait then until they accept the PR.

It seems like that project has no maintainer now.

darh:
What can we do here?

I'm not in favour of changing the package like this. Is there another, more active fork?

PR #445, opened 2022-11-01
fix(rdbms): check number regex

Yangruipis:
Regex for checking number is (`'^[0-9]+(\.[0-9])*$'`), which makes floats with 2+ precision like 0.01 was not treat as a number.


darh:
Nice catch -- but then the `*` needs to be changed to `?`, right?

Yangruipis:
yes nice catch as well 😆 and the pr is updated

darh:
Rebased and merged to `2022.9.x` and `2023.3.x`

PR #454, opened 2022-11-07
2022.9.x fix dal multi value field

darh:


pkg/ql/ast_nodes.go
```diff
@@ -15,6 +15,11 @@ type (
 	ASTNode struct {
 		pMeta *parserMeta
 
+		// Meta holds additional node data
+		// that could be passed from time of p
+		//
+		Meta map[any]any
```
tjerman: consider using `map[string]any` since they're all strings currently

store/adapters/rdbms/dal/main_test.go
```diff
@@ -18,13 +18,19 @@ import (
 )
 
 type (
-	kv map[string]any
+	// @todo refactor tests that use kv and migrate to kvv
+	//       no need to keep and maintain both.
+	kv  map[string]any
+	kvv map[string][]any
```
tjerman, line 24: You could consider using `Row` from `pkg/dal/utils.go` as well

store/adapters/rdbms/dal/main_test.go
```diff
@@ -37,6 +43,7 @@ func TestMain(m *testing.M) {
 	if len(dsn) == 0 {
 		// a temporary solution to make sure all tests are ran inside sqlite
 		dsn = "sqlite3+debug://file::memory:?cache=shared&mode=memory"
+		//dsn = "sqlite3+debug://model.db"
```
tjerman: should this be deleted?

store/adapters/rdbms/drivers/dialect.go
```diff
@@ -42,3 +65,121 @@ type (
...
+func OpHandlerIn(d Dialect, n *ql.ASTNode, args ...exp.Expression) (expr exp.Expression, err error) {
+	return opHandlerIn(d, n, false, args...)
+}
+
+func OpHandlerNotIn(d Dialect, n *ql.ASTNode, args ...exp.Expression) (expr exp.Expression, err error) {
+	return opHandlerIn(d, n, true, args...)
+}
+
+func opHandlerIn(d Dialect, n *ql.ASTNode, negate bool, args ...exp.Expression) (expr exp.Expression, err error) {
+	if n.Args[1] != nil && n.Args[1].Meta["dal.Attribute"] != nil && n.Args[1].Meta["dal.Attribute"].(*dal.Attribute).MultiValue {
```
tjerman: can n.Args ever be empty? This would panic then

store/adapters/rdbms/drivers/mysql/dialect.go
```diff
@@ -25,21 +23,64 @@ var (
 	dialect            = &mysqlDialect{}
 	goquDialectWrapper = goqu.Dialect("mysql")
 	quoteIdent         = string(mysql.DialectOptions().QuoteRune)
+
+	nuances = drivers.Nuances{
+		HavingClauseMustUseAlias: false,
```
tjerman, line 28: In the comment above you write `// MySQL requires use of aliases inside HAVING clause` did you mean some other DB?

store/adapters/rdbms/drivers/dialect.go
darh: I thought I can rely on `ql` package :)

store/adapters/rdbms/drivers/mysql/dialect.go
darh: Correct, it is other way around.

PR #506, opened 2022-11-15
Legacy corteza server add beta column modification

skamensky:
Unfortunately when doing some cleanup the PR's I created were deleted (probably because I deleted my fork and created a new one).

As I left it standing I needed to clean up some tests that were failing. I'll get working on that now.

### DIFF
--- a/server/pkg/dal/diff.go
+++ b/server/pkg/dal/diff.go
@@ -59,6 +59,10 @@ func (a *Model) Diff(b *Model) (out ModelDiffSet) {
 	// Deleted and update ones
 	for _, _attrA := range a.Attributes {
 		attrA := _attrA
+		// store is an interface to something that could be a pointer.
+		// we need to copy it to make sure we don't get a nil pointer
+		// make sure not to modify this since it would modify the original
+		attrA.Store = _attrA.Store
 
 		// Missmatches
 		attrBAux, ok := bIndex[attrA.Ident]
@@ -91,7 +95,6 @@ func (a *Model) Diff(b *Model) (out ModelDiffSet) {
 				Inserted:     attrBAux.attr,
 			})
 		}
-
 		if attrA.Store.Type() != attrBAux.attr.Store.Type() {
 			out = append(out, &ModelDiff{
 				Type:         AttributeCodecMismatch,
--- a/server/pkg/dal/diff.go
+++ b/server/pkg/dal/diff.go
@@ -1,24 +1,31 @@
 package dal
 
 type (
-	modelDiffType string
-
+	modelDiffType     string
+    ModelModification string
 	// ModelDiff defines one identified missmatch between two models
 	ModelDiff struct {
-		Type modelDiffType
-		// Original will be nil when a new attribute is being added
+		Type         modelDiffType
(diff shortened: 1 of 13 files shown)


skamensky:
Hey @darh ,

I'm ready for a final review when you guys have the time. I also caught a small bug and fixed it ( see 561a7761a9deed5f018facb679b7195428292442 ) so that a test would not fail.
Some tests do fail, but these were failiing when I started working on this:


`TestApigwRouteUpdate`
`TestAuthExternalSAMLSuccess`
`Test0001_basics_detect_datarace_issues`

darh:
Reviewing this now. It will not go to 2022.9.3 (this week) because I want it to live in a version branch for a couple of days.

skamensky:
What is a version branch? `2022.9.3` is not a version branch?

darh:
version branch would be `2022.9.x`

More info: https://docs.cortezaproject.org/corteza-docs/2022.9/developer-guide/release-cycle/index.html

server/pkg/dal/diff.go
```diff
@@ -1,24 +1,29 @@
...
 )
 
 const (
-	AttributeMissing              modelDiffType = "attributeMissing"
-	AttributeTypeMissmatch        modelDiffType = "typeMissmatch"
-	AttributeSensitivityMissmatch modelDiffType = "sensitivityMissmatch"
+	AttributeMissing             modelDiffType     = "attributeMissing"
+	AttributeTypeMissmatch       modelDiffType     = "typeMissmatch"
+	AttributeSensitivityMismatch modelDiffType     = "sensitivityMismatch"
+	AttributeCodecMismatch       modelDiffType     = "sensitivityMismatch"
```
darh: value probably needs to be "codecMismatch"

server/store/adapters/rdbms/dal/connection.go
```diff
@@ -161,14 +162,21 @@ func (c *connection) CreateModel(ctx context.Context, mm ...*dal.Model) (err err
 	c.mux.Lock()
 	defer c.mux.Unlock()
 	for _, m := range mm {
-		err = ddl.UpdateModel(ctx, c.dataDefiner, m)
+
+		if err != nil {
+			return
+		}
+		_, err = c.dataDefiner.TableLookup(ctx, m.Ident)
```
darh, line 165: This does not make sense. Probably error check after TableLookup call?

darh (changes requested):
Tried it. Works 🚀 . Two minor things -- see coments.

server/store/adapters/rdbms/dal/connection.go
skamensky: Yeah you're right. There already is an error check after the TableLookup call. The check you noticed was an unnecessary no-op. Fixed.

server/pkg/dal/diff.go
skamensky: Good catch. Fixed.

skamensky:
@darh ,

I've address your comments

darh:
👍 I'll squash and merge in the next couple of days.

PR #513, opened 2022-11-16
Add auto-refresh to page blocks

### LINKED ISSUE #466: Add support for auto-refresh of certain block-types
# Problem
Compose page blocks do not support auto-content refreshing and this forces users to refresh the browser or navigate away and back to the current page.

# Extending page block configurators
All page block configurators need to be extended to support auto-refresh settings, explicitly enable or disable manual refresh of a block.

Refresh interval can be expressed in seconds and must have lower limit at 5 seconds.

# Page refresh implementation
Since automatic refresh can interfere with user's workflow and can cause data loss, extra protection need to be implemented to prevent that.  

Image

- update 2022.3 Plan #4 @darh

EmmyMay:
Please take a look.

### DIFF
--- a/client/web/compose/src/components/PageBlocks/ChartBase.vue
+++ b/client/web/compose/src/components/PageBlocks/ChartBase.vue
@@ -2,9 +2,11 @@
   <wrap
     v-bind="$props"
     v-on="$listeners"
+    @refreshBlock="refresh"
   >
     <chart-component
       v-if="chart"
+      :key="key"
       :chart="chart"
       :record="record"
       :reporter="reporter"
@@ -35,22 +37,28 @@ export default {
   },
 
   mounted () {
-    const { chartID } = this.options
-
-    if (chartID === NoID) {
-      return
-    }
-
-    const { namespaceID } = this.namespace
-    this.findChartByID({ chartID, namespaceID }).then((chart) => {
-      this.chart = chart
-    }).catch(this.toastErrorHandler(this.$t('chart.loadFailed')))
+    this.fetchChart()
+    this.refreshBlock(this.refresh, true)
   },
 
   methods: {
     ...mapActions({
       findChartByID: 'chart/findByID',
     }),
+
+    fetchChart (params = {}) {
+      const { chartID } = this.options
+
+      if (chartID === NoID) {
+        return
+      }
+
+      const { namespaceID } = this.namespace
+
+      this.findChartByID({ chartID, namespaceID, ...params }).then((chart) => {
+        this.chart = chart
+      }).catch(this.toastErrorHandler(this.$t('chart.loadFailed')))
+    },
 
     reporter (r) {
       const nr = { ...r }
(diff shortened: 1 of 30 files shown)


client/web/compose/src/components/PageBlocks/ChartBase.vue
```diff
@@ -72,6 +80,12 @@ export default {
       const { namespaceID } = this.namespace
       return this.$ComposeAPI.recordReport({ namespaceID, ...nr })
     },
+
+    refresh () {
+      this.fetchChart({ force: true })
+    },
+
```
Fajfa: Remove empty line

client/web/compose/src/components/PageBlocks/CommentBase.vue
```diff
@@ -227,7 +232,8 @@ export default {
       this.$store.dispatch('user/fetchUsers', userListID)
     },
 
-    reloadRecords () {
+    // reloadRecords
```
Fajfa: No need for this comment (doesn't really provide any info)

client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -80,14 +81,38 @@
...
+          v-if="block.options.refreshRate !== undefined"
+          :label="$t('general.refresh.label')"
+          :description="$t('general.refresh.description')"
         >
-          {{ $t('general.wrap') }}
-        </b-form-checkbox>
+          <b-col
+            cols="12"
+            sm="3"
+            class="pl-0"
```
Fajfa, line 103: i think p-0 is the play here (not 100%)

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -4,30 +4,18 @@
     v-bind="$props"
     :scrollable-body="false"
     v-on="$listeners"
+    @refreshBlock="refresh"
   >
     <template
-      v-if="showHeader"
-      #header
```
Fajfa, line 10: Please triple check if no other blocks used this slot, we need to handle that if they did

client/web/compose/src/components/PageBlocks/RecordOrganizerBase.vue
```diff
@@ -421,6 +410,27 @@ export default {
       return this.$ComposeAPI.recordList({ namespaceID, moduleID, query, sort })
         .then(({ set }) => set.map(r => Object.freeze(new compose.Record(module, r))))
     },
+
+    // pull records
```
Fajfa: Remove this comment as well

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/Configurator.vue
EmmyMay: Not really. pl is okay. It gets too wide with p-0

PR #524, opened 2022-11-17
Implement opening record page in a modal

yonkov:
## Screenshots of the implementation

Screenshot from 2022-11-11 13-52-31

Screenshot from 2022-11-11 13-48-19

## Changelog

**What was added?**

When adding a "record list" page block in the Compose page builder, a dropdown field has been added with three options:

- open record in the same tab
- open record in a new tab
- open record in a modal

The default option is to open in the same tab.

_Note: The first two options already existed and are handled gracefully, e.g. respected for existing record lists_

The new thing here is the third option - "open record in a modal". When this option is selected and the user goes to a record list page or a page that displays record lists, he can click on a specific record and it will open as a modal. From the modal, the user can do whatever he wants with the existing record (edit, delete, clone) or even add a new record.

The opened modal remains open even after the user refreshes the page for some reason. This also allow users to share links with the modal opened or closed.

**How it was added?**

In the Compose app, a root event was emitted from the RecordList component that emits an event with all the necessary information to a new modal component, which has been called from the public page. The modal is then able to display the record and its information. In order to do this, instead of using routes, the ViewRecord component is added as a child component to the modal and it receives its props from the modal itself. The  modal state (opened or closed) is persisted  after page reload by adding query params with the record data in the url. We will reuse this functionality in future tasks, since it is benefitial for UI.

### DIFF
--- a/client/web/compose/src/components/Common/RecordToolbar.vue
+++ b/client/web/compose/src/components/Common/RecordToolbar.vue
@@ -1,7 +1,8 @@
 <template>
   <b-container
     fluid
-    class="bg-white shadow border-top p-3"
+    :class="{'shadow border-top': !showRecordModal}"
+    class="bg-white p-3"
   >
     <b-row
       no-gutters
@@ -19,10 +20,10 @@
           @click.prevent="$emit('back')"
         >
           <font-awesome-icon
-            :icon="['fas', 'chevron-left']"
+            :icon="['fas', showRecordModal ? 'times' : 'chevron-left']"
             class="back-icon"
           />
-          {{ labels.back || $t('label.back') }}
+          {{ showRecordModal ? $t('label.close') : labels.back || $t('label.back') }}
         </b-button>
       </div>
 
@@ -172,6 +173,11 @@ export default {
     isDeleted: {
       type: Boolean,
       default: true,
+    },
+
+    showRecordModal: {
+      type: Boolean,
+      required: false,
     },
   },
 
--- a/client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
@@ -250,12 +250,20 @@
         <b-form-checkbox v-model="options.selectable">
           {{ $t('recordList.selectable') }}
         </b-form-checkbox>
-        <b-form-checkbox
-          v-model="options.openInNewTab"
-        >
-          {{ $t('recordList.record.openInNewTab') }}
-        </b-form-checkbox>
       </b-form-group>
+
+      <b-form-group
+        :label="$t('recordList.record.recordDisplayOptions')"
+        :label-cols="3"
+        breakpoint="md"
+        horizontal
+      >
+        <b-form-select
+          v-model="options.recordDisplayOption"
+          :options="recordDisplayOptions"
+        />
+      </b-form-group>
+
       <b-form-group
         horizontal
         :label-cols="3"
@@ -323,6 +331,14 @@ export default {
       modules: 'module/set',
       pages: 'page/set',
     }),
+
+    recordDisplayOptions () {
+      return [
+        { value: 'sameTab', text: this.$t('recordList.record.openInSameTab') },
+        { value: 'newTab', text: this.$t('recordList.record.openInNewTab') },
+        { value: 'modal', text: this.$t('recordList.record.openInModal') },
+      ]
+    },
 
     moduleOptions () {
       return [
--- a/client/web/compose/src/components/Public/Record/Modal.vue
+++ b/client/web/compose/src/components/Public/Record/Modal.vue
@@ -1,0 +1,123 @@
+<template>
+  <b-modal
+    id="record-modal"
+    v-model="showRecordModal"
+    scrollable
+    body-class="p-0"
+    footer-class="p-0"
+    size="xl"
+    @hidden="hideModal"
+  >
+    <template #modal-title>
+      <portal-target
+        name="record-modal-header"
+      />
+    </template>
+
+    <view-record
+      :namespace="namespace"
+      :page="page"
+      :module="module"
+      :record-i-d="recordID"
+      :show-record-modal="showRecordModal"
+    />
+
+    <template #modal-footer>
+      <portal-target
+        name="record-modal-footer"
+        class="w-100 m-0"
+      />
+    </template>
+  </b-modal>
+</template>
+
+<script>
+import record from 'corteza-webapp-compose/src/mixins/record'
+import { compose } from '@cortezaproject/corteza-js'
+import ViewRecord from 'corteza-webapp-compose/src/views/Public/Pages/Records/View'
+import { mapGetters } from 'vuex'
+
+export default {
+  i18nOptions: {
+    namespaces: 'block',
+  },
+
+  name: 'RecordModal',
+
+  components: {
+    ViewRecord,
+  },
+
+  mixins: [
+    record,
+  ],
+
+  props: {
+    namespace: {
+      type: compose.Namespace,
+      required: true,
+    },
+  },
+
+  data () {
+    return {
+      showRecordModal: false,
+      recordID: null,
+      module: null,
+      page: null,
+    }
+  },
+
+  computed: {
+    ...mapGetters({
+      getModuleByID: 'module/getByID',
+      getPageByID: 'page/getByID',
+    }),
+  },
+
+  created () {
+    this.$root.$on('show-record-modal', this.showModal)
+    this.showModal(this.$route.query)
+  },
+
+  beforeDestroy () {
+    this.$root.$off('show-record-modal')
+  },
+
+  methods: {
+    showModal ({ recordID, moduleID, recordPageID }) {
+      if (recordID && moduleID && recordPageID) {
+        this.recordID = recordID
+        this.module = this.getModuleByID(moduleID)
+        this.page = this.getPageByID(recordPageID)
+
+        this.showRecordModal = true
+
+        // persist the modal in the url
+        this.$router.replace({
+          query: {
+            recordID,
+            moduleID,
+            recordPageID: this.page.pageID,
+          },
+        })
+      }
+    },
+
+    hideModal () {
+      this.$router.replace({ query: { } })
+      this.recordID = null
+      this.module = null
+      this.page = null
+    },
+  },
+}
+
+</script>
+
+<style>
+#record-modal .modal-dialog {
+  height: 100%;
+  max-width: 90vw;
+}
+</style>
--- a/client/web/compose/src/views/Public/Pages/Records/View.vue
+++ b/client/web/compose/src/views/Public/Pages/Records/View.vue
@@ -1,5 +1,8 @@
 <template>
-  <div class="d-flex flex-grow-1 w-100">
+  <div
+    :class="{ 'flex-column': showRecordModal }"
+    class="d-flex flex-grow-1 w-100 h-100"
+  >
     <b-alert
       v-if="isDeleted"
       show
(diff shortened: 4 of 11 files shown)


client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -19,10 +20,16 @@
...
         >
           <font-awesome-icon
+            v-if="!showRecordModal"
             :icon="['fas', 'chevron-left']"
             class="back-icon"
           />
-          {{ labels.back || $t('label.back') }}
+          <font-awesome-icon
+            v-else
+            :icon="['fa', 'times']"
```
Fajfa: Instead of repeating the whole font-awesome element, just have a computed for the icon name, since otherwise the two font awesome elements are the same

`['fas', backIcon]`

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -19,10 +20,16 @@
...
+            class="back-icon"
+          />
```
Fajfa, line 25: A new line between elements that share same parent (like they do here)

client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
```diff
@@ -250,12 +250,21 @@
...
           {{ $t('recordList.selectable') }}
         </b-form-checkbox>
-        <b-form-checkbox
-          v-model="options.openInNewTab"
-        >
-          {{ $t('recordList.record.openInNewTab') }}
-        </b-form-checkbox>
       </b-form-group>
+
+      <!--Select options how to open a record-->
```
Fajfa: No need for this comment

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+  methods: {
+    showModal (e) {
+      const { recordID, moduleID, recordPageID } = e
+      this.showRecordModal = true
+      this.recordID = recordID
+      this.page = this.pages.find(p => p.pageID === recordPageID)
+      this.module = this.getModuleByID(moduleID)
+      // persist the modal in the url
+      this.$router.replace({ query: { 
+        modalRecordID: recordID,
```
Fajfa: Can we just call them recordID, moduleID, pageID without the modal prefix

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -119,7 +139,15 @@ export default {
 
     title () {
       const { name, handle } = this.module
-      return this.$t('page:public.record.view.title', { name: name || handle, interpolation: { escapeValue: false } })
+      let titlePreffix
```
Fajfa: You can set default to 'view' and then you don't need the else.

Plus typo on prefix! (I suggest you install the spellchecker extension)

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -119,7 +139,15 @@ export default {
...
+      if (this.inCreating && this.inEditing) {
```
Fajfa: Think of how you can write this without this many lines please

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+      }
+    },
+  },
+
+  created () {
+    this.$root.$on('show-record-modal', this.showModal)
+
+    const { modalRecordID, modalModuleID, modalPageID } = this.$route.query
+    if (modalRecordID) {
+      this.showRecordModal = true
```
Fajfa: Why this ? if you do it in show modal anyway?

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+        this.$router.replace({ query: { } })
+      }
+    },
+  },
+
+  created () {
+    this.$root.$on('show-record-modal', this.showModal)
+
+    const { modalRecordID, modalModuleID, modalPageID } = this.$route.query
+    if (modalRecordID) {
```
Fajfa: Shouldn't you check for this in the showModal, and just pass this in no matter what?

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+    ...mapGetters({
+      getModuleByID: 'module/getByID',
+      pages: 'page/set',
+    }),
+  },
+
+  watch: {
+    showRecordModal: function(){
+      if(!this.showRecordModal){
+        this.$router.replace({ query: { } })
```
Fajfa: Why this? You need to replace it when the modal is closed, and you can do this in a better way rather than the watcher

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+      this.recordID = recordID
+      this.page = this.pages.find(p => p.pageID === recordPageID)
+      this.module = this.getModuleByID(moduleID)
+      // persist the modal in the url
+      this.$router.replace({ query: { 
+        modalRecordID: recordID, 
+        modalModuleID: moduleID, 
+        modalPageID: this.page.pageID 
+      } })
+      .catch(()=>{})
```
Fajfa: Whats up with the catch? This isn't a promise

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,132 @@
...
+  },
+
+  methods: {
+    showModal (e) {
+      const { recordID, moduleID, recordPageID } = e
+      this.showRecordModal = true
+      this.recordID = recordID
+      this.page = this.pages.find(p => p.pageID === recordPageID)
+      this.module = this.getModuleByID(moduleID)
+      // persist the modal in the url
```
Fajfa: Make an empty line between info gathering and persisting modal please

Fajfa (changes requested):

client/web/compose/src/components/Public/Record/Modal.vue
yonkov: I need this check here, otherwise the modal iinstantiates on the record list page on every page load, even when no modal has been opened and there are no query params in the url. Instead, I need to show the modal only if the data exists in the query params (the modal has been opened before the page reload).

client/web/compose/src/components/Public/Record/Modal.vue
yonkov: If I remove the error handling here, I got an error "Uncaught (in promise) NavigationDuplicated: Avoided redundant navigation to current location". I found some info on the matter in this thread: https://stackoverflow.com/questions/62462276/how-to-solve-avoided-redundant-navigation-to-current-location-error-in-vue

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -19,10 +20,10 @@
           @click.prevent="$emit('back')"
         >
           <font-awesome-icon
-            :icon="['fas', 'chevron-left']"
+            :icon="showRecordModal ? ['fa', 'times'] : ['fas', 'chevron-left']"
```
Fajfa: Close, `['fas', showRecordModal ? 'times' : 'chevron-left']`

client/web/compose/src/components/Public/Record/Modal.vue
Fajfa: Exactly, thats why you can do this check, plus moduleID and pageID in the showModal function.

client/web/compose/src/components/Public/Record/Modal.vue
Fajfa: Yes, that is an error we dont catch, since its a dev only error. So remove it please

client/web/compose/src/components/Public/Record/Modal.vue
Fajfa: This still needs to be addressed

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -97,6 +110,13 @@ export default {
   },
 
   computed: {
+    portalTopbarTitle () {
+      return this.showRecordModal ? 'record-modal-header' : 'topbar-title'
+    },
```
Fajfa, line 115: Empty line between computeds

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -119,7 +139,9 @@ export default {
 
     title () {
       const { name, handle } = this.module
-      return this.$t('page:public.record.view.title', { name: name || handle, interpolation: { escapeValue: false } })
+      const titlePrefix = this.inCreating && this.inEditing ? 'create' : !this.inCreating && this.inEditing ? 'edit' : 'view'
```
Fajfa: Not quite, no need to check inEditing, if inCreating is true then you're creating, otherwise if inEditing is true then you're editing, otherwise viewing

Fajfa (changes requested):

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,130 @@
...
+      const { recordID, moduleID, recordPageID } = e
+      this.showRecordModal = true
+      this.recordID = recordID
+      this.page = this.pages.find(p => p.pageID === recordPageID)
+      this.module = this.getModuleByID(moduleID)
+
+      // persist the modal in the url
+      this.$router.replace({
+        query: {
+          modalRecordID: recordID,
```
Fajfa: Not sure why you resolved my previous comment regarding this, but no need to call the variables with the modal prefix. Just use recordID and so on

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,129 @@
...
+      recordPageID,
+    })
+  },
+
+  beforeDestroy () {
+    this.$root.$off('show-record-modal')
+  },
+
+  methods: {
+    showModal (e) {
```
Fajfa: Deconstruct the params in the arguments please, i think i mentioned this once before in our calls

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,129 @@
...
+      const { recordID, moduleID, recordPageID } = e
```
Fajfa: After this line is when you should check the params and only after that try to fetch them.

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,129 @@
...
+      this.recordID = recordID
+      this.page = this.pages.find(p => p.pageID === recordPageID)
```
Fajfa: You should get the page the same way you do the module, by using the store getter instead of manually going through the pages

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,129 @@
...
+
+</script>
+
+<style>
+#record-modal .modal-dialog {
+  height: 100%;
+  max-width: 90vw;
+}
+
+#record-modal .modal-content {
```
Fajfa: I think you only need the class above this one, double check

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -0,0 +1,129 @@
...
+          query: {
+            recordID,
+            moduleID,
+            recordPageID: this.page.pageID,
+          },
+        })
+      }
+    },
+    hideModal () {
+      this.$router.replace({ query: { } })
```
Fajfa, line 108: Lets also reset the data back to null (recordID, pageID and moduleID)

Fajfa (changes requested):

PR #539, opened 2022-11-18
Add support for SMTP Configurations test

### LINKED ISSUE #497: SMTP send an email button
It will be really beneficial if we have a test button in SMTP server setting in admin that will automatically trigger an email to be sent by clicking on it. This will come in handy when testing SMTP server settings, MailHog and other features.

KinyaElGrande:
This is a new feature that implements SMTP Configurations test.
It has 2 sections :

1.  Test configuration (connects and ping HELO to the SMTP server)
2.  Send an actual hello from Corteza to provided recipient's email address(es)

V1 validates  the following bits :
- host & port -- are the params correct and can we ping the server
- username & password -- are the params correct and can we authenticate
- connectivity -- can we connect to the server

### DIFF
--- a/client/web/admin/src/components/Settings/Email/CSystemEmailServer.vue
+++ b/client/web/admin/src/components/Settings/Email/CSystemEmailServer.vue
@@ -1,8 +1,8 @@
 <template>
   <b-card
-    class="shadow-sm"
     header-bg-variant="white"
     footer-bg-variant="white"
+    class="shadow-sm"
   >
     <b-form
       @submit.prevent="submit()"
@@ -34,6 +34,7 @@
           </b-input-group-append>
         </b-input-group>
       </b-form-group>
+
       <b-form-group
         :label="$t('user.label')"
         :description="$t('user.description')"
@@ -45,6 +46,7 @@
           autocomplete="off"
         />
       </b-form-group>
+
       <b-form-group
         :label="$t('password.label')"
         :description="$t('password.description')"
@@ -86,6 +88,7 @@
           {{ $t('tlsInsecure.label') }}
         </b-form-checkbox>
       </b-form-group>
+
       <b-form-group
         :label="$t('tlsServerName.label')"
         :description="$t('tlsServerName.description')"
@@ -106,10 +109,21 @@
 
     <template #footer>
       <c-submit-button
-        class="float-right"
         :disabled="disabled"
         :processing="processing"
         :success="success"
+        variant="light"
+        class="float-left"
+        @submit="smtpConnectionCheck()"
+      >
+        {{ $t('testSmtpConfigs.button') }}
+      </c-submit-button>
+
+      <c-submit-button
+        :disabled="disabled"
+        :processing="processing"
+        :success="success"
+        class="float-right"
         @submit="submit()"
       />
     </template>
@@ -171,6 +185,10 @@ export default {
   methods: {
     submit () {
       this.$emit('submit', this.server)
+    },
+
+    smtpConnectionCheck () {
+      this.$emit('smtpConnectionCheck', this.server)
     },
   },
 }
--- a/client/web/admin/src/views/System/Email/Index.vue
+++ b/client/web/admin/src/views/System/Email/Index.vue
@@ -13,6 +13,7 @@
       :success="auth.success"
       :disabled="!canManage"
       @submit="onEmailServerSubmit($event)"
+      @smtpConnectionCheck="onSmtpConnectionCheck($event)"
     />
   </b-container>
 </template>
@@ -113,6 +114,41 @@ export default {
           this.external.processing = false
         })
     },
+
+    onSmtpConnectionCheck (server) {
+      this.external.processing = true
+
+      // Append the list of recepient's email  Addresses
+      const recepients = []
+      recepients.push(server.from)
+
+      this.$SystemAPI.smtpConfigurationCheckerCheck({
+        host: server.host,
+        port: parseInt(server.port),
+        recipients: recepients,
+        username: server.user,
+        password: server.pass,
+        tlsInsecure: server.tlsInsecure,
+        tlsServerName: server.tlsServerName,
+      })
+        .then(response => {
+          if (Object.values(response).every(resp => resp === '')) {
+            this.animateSuccess('external')
+            this.toastSuccess(this.$t('notification:settings.system.smtpCheck.success'))
+          }
+
+          Object.keys(response).forEach(key => {
+            if (response[key]) {
+              this.toastWarning(`${key}: ${response[key]}`)
+            }
+          })
+        })
+        .catch(this.toastErrorHandler(this.$t('notification:settings.system.smtpCheck.error')))
+        .finally(() => {
+          this.external.processing = false
+        })
+    },
   },
+
 }
 </script>
--- a/server/provision/002_templates/healthcheck.yml
+++ b/server/provision/002_templates/healthcheck.yml
@@ -1,0 +1,17 @@
+templates:
+  smtp_configuration_check_subject:
+    type: text/plain
+    meta:
+      short: SMTP configuration check content
+    template: SMTP connection check
+
+  smtp_configuration_check_content:
+    type: text/html
+    meta:
+      short: SMTP configuration check content
+    template: |-
+      {{template "email_general_header" .}}
+        <h2 style="color: #568ba2;text-align: center;">SMTP configurations check</h2>
+        <p>Hello,</p>
+        <p>Your SMTP configuration test passed</p>
+      {{template "email_general_footer" .}}
--- a/server/system/rest.yaml
+++ b/server/system/rest.yaml
@@ -2256,3 +2256,21 @@ endpoints:
             required: false
             title: Exclude (0, default), include (1) or return only (2) deleted connections
             type: filter.State
+
+- title: SMTP Configuration Checker
+  entrypoint: smtpConfigurationChecker
+  path: "/smtp"
+  apis:
+    - name: check
+      method: POST
+      title: Check SMTP server configuration settings
+      path: "/configuration-checker/"
+      parameters:
+        post:
+          - { name: host,           type: string,       title: SMTP server host name,  required: true }
+          - { name: port,           type: uint,         title: SMTP server port, }
+          - { name: recipients,     type: "[]string",   title: List of recipients email addresses that should recieve test email }
+          - { name: username,       type: string,       title: SMTP server authentication username }
+          - { name: password,       type: string,       title: SMTP server authentication password }
+          - { name: tlsInsecure,    type: bool,         title: TLS mode }
+          - { name: tlsServerName,  type: string,       title: TLS server name }
--- a/server/system/service/smtp_configuration_checker.go
+++ b/server/system/service/smtp_configuration_checker.go
@@ -1,0 +1,207 @@
+package service
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	intAuth "github.com/cortezaproject/corteza/server/pkg/auth"
+	"github.com/cortezaproject/corteza/server/pkg/mail"
+	"github.com/cortezaproject/corteza/server/pkg/options"
+	"github.com/cortezaproject/corteza/server/system/types"
+	gomail "gopkg.in/mail.v2"
+	htpl "html/template"
+	"io/ioutil"
+	"strings"
+)
+
+type (
+	smtpConfigurationChecker struct {
+		settings      *types.AppSettings
+		ts            TemplateService
+		opt           options.AuthOpt
+		accessControl smtpCheckAccessController
+	}
+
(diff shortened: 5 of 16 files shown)


server/provision/002_templates/healthcheck.yml
```diff
@@ -0,0 +1,17 @@
...
+      short: SMTP connection check content
+    template: SMTP connection check
+
+  smtp_connection_check_content:
+    type: text/html
+    meta:
+      short: SMTP connection check content
+    template: |-
+      {{template "email_general_header" .}}
+        <h2 style="color: #568ba2;text-align: center;">Reset your password</h2>
```
darh: Leftovers :)

server/system/rest.yaml
```diff
@@ -2256,3 +2256,34 @@ endpoints:
...
+            name: recipients
+            required: false
+            title: List of recipients email addresses that should recieve test email
+          - type: string
+            name: username
+            required: false
+            title: SMTP server authentication username
+          - type: string
+            name: password
+            required: false
```
darh: remove `required: false` (this is default) and if you can, squeeze each param into a single line so the whole thing is a bit shorter (Se "Create client" in the same file for example

server/system/rest.yaml
```diff
@@ -2256,3 +2256,34 @@ endpoints:
...
+      path: "/checker/"
+      parameters:
+        post:
+          - type: string
+            name: host
+            required: true
+            title: SMTP server host name
+          - type: uint
+            name: port
+            required: true
```
darh: could we have this optional with the standard port as default?

server/system/rest.yaml
```diff
@@ -2256,3 +2256,34 @@ endpoints:
...
+
+- title: SMTP Connection Checker
+  entrypoint: smtpConnection
+  path: "/smtp/connection"
+  apis:
+    - name: check
+      method: POST
+      title: Check SMTP server settings
+      path: "/checker/"
+      parameters:
```
darh, line 2268: did you skip other connection options (tls insecure, tls server name) intentionally?

server/system/rest/handlers/smtpConnection.go
```diff
@@ -0,0 +1,57 @@
...
+// the code is regenerated.
+//
+// Definitions file that controls how this file is generated:
+//
+
+import (
+	"context"
+    "github.com/cortezaproject/corteza/server/pkg/api"
+    "github.com/cortezaproject/corteza/server/system/rest/request"
+	"github.com/go-chi/chi/v5"
```
darh: did you format your code with `go fmt`?

server/pkg/mail/healthcheck.go
```diff
@@ -3,6 +3,14 @@ package mail
...
 	"context"
 	"fmt"
+	"net/smtp"
+	"regexp"
+	"strconv"
+)
+
+const (
+	hostCheckRE = "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
+	portCheckRE = "^([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$"
```
darh: why RE for number checking?

server/pkg/mail/healthcheck.go
```diff
@@ -17,3 +25,39 @@ func Healthcheck(ctx context.Context) error {
 
 	return nil
 }
+
+func IsValidHost(host string) bool {
```
darh: are these new functions here really healtchecking functions or just... checking functions? I would rather see them in a different file

server/pkg/mail/healthcheck.go
```diff
@@ -17,3 +25,39 @@ func Healthcheck(ctx context.Context) error {
...
+	hostCheck := regexp.MustCompile(hostCheckRE)
+	return hostCheck.MatchString(host)
+}
+
+func IsValidPort(port uint) bool {
+	portCheck := regexp.MustCompile(portCheckRE)
+	portStr := strconv.Itoa(int(port))
+	return portCheck.MatchString(portStr)
```
darh: that's just silly :)

server/system/service/smtp_connection_checker.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (svc smtpConnectionChecker) smtpSendEmail(ctx context.Context, recipients []string) (err error) {
+	var (
+		tmp []byte
+		ntf = mail.New()
+		hdl string
+		// context with service user
+		// we need this for retrieving & rendering email templates
+		suCtx = intAuth.SetIdentityToContext(ctx, intAuth.ServiceUser())
```
darh: Kudos on this!! First I thought you missed the scenario where user could be w/o permissions to check templates. Kudos!

server/system/service/smtp_connection_checker.go
```diff
@@ -0,0 +1,159 @@
...
+	return &smtpConnectionChecker{
+		actionlog: DefaultActionlog,
+		settings:  s,
+		ts:        ts,
+		opt:       opt,
+	}
+}
+
+// SmtpServerChecker Expected to error out
+func (svc smtpConnectionChecker) SmtpServerChecker(ctx context.Context, smtpConfigs *types.SmtpServer) (err error) {
```
darh: Let us restructure this function a bit.

The desired goal is not to send an email but to check the SMTP configuration.
So, invalid SMTP settings, and inability to send email is not an error but a result of this check (you expect it that it might not work).
But the invalid input parameters or inability to load the template you use to send an check email is an actual error (you are expecting it to be valid and to work).

So, return 2 vars; the first one would be a struct with checks that passed or failed (and reasons for them). Could be a simple:

```go
SmtpCheckResult struct {
   Host error `json:"host"`
   Port error
   ServerCheck error
   SendCheck error
  // ....?
   Valid bool `json:....`
}
```

server/system/service/smtp_connection_checker.go
```diff
@@ -0,0 +1,159 @@
...
+		email string
+	)
+
+	if len(rr) == 0 {
+		return
+	}
+
+	for _, rcpt := range rr {
+
+		email = strings.TrimSpace(rcpt)
```
darh: why you are not reusing th rcpt var here?

server/system/types/smtp_server.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+type (
+	SmtpServer struct {
```
darh: Rename this to something more appropriate. 

_What it "says on a box" is not what it does._ 

Why did you decide to use a struct (and not use the fn. args)?

darh (changes requested):

server/pkg/mail/healthcheck.go
KinyaElGrande: This is for checking the Port Number [making sure it's between the range of [0 and 65535]

server/pkg/mail/healthcheck.go
```diff
@@ -17,3 +19,32 @@ func Healthcheck(ctx context.Context) error {
 
 	return nil
 }
+
+func SMTPConfigurationCheck(host string, port uint, username string, password string, tlsConfig *tls.Config) (expected string) {
```
darh: 1) this is not a healthcheck, move this to `check.go`
2) remove the smtp prefix - just ConfigCheck 
3) `expected` sounds a bit off, do we have a better alternative?

server/system/service/smtp_configuration_checker.go
```diff
@@ -0,0 +1,215 @@
...
+	gomail "gopkg.in/mail.v2"
+	htpl "html/template"
+	"io/ioutil"
+	"regexp"
+	"strings"
+)
+
+type (
+	smtpConfigurationChecker struct {
+		actionlog actionlog.Recorder
```
darh: this is unused, right?

server/system/service/smtp_configuration_checker.go
```diff
@@ -0,0 +1,215 @@
...
+func SmtpConfigurationChecker(s *types.AppSettings, ts TemplateService, opt options.AuthOpt) *smtpConfigurationChecker {
+	return &smtpConfigurationChecker{
+		actionlog: DefaultActionlog,
+		settings:  s,
+		ts:        ts,
+		opt:       opt,
+	}
+}
+
+func (svc smtpConfigurationChecker) SmtpConfigurationCheck(ctx context.Context, smtpConfigs *types.SmtpConfiguration) (checkResults *types.SmtpCheckResult, err error) {
```
darh: - We talked about this - rename it to `Check`
 - What I'm also missing here is any kind of access control; anyone can call this function and we do not want that. Check if the user can manage settings (`CanManageSettings`).

server/system/types/smtp_configuration.go
```diff
@@ -0,0 +1,24 @@
...
+		TLSServerName string
+	}
+
+	SmtpCheckResult struct {
+		Host   CheckResponse `json:"host"`
+		Server CheckResponse `json:"server"`
+		Send   CheckResponse `json:"send"`
+	}
+
+	CheckResponse struct {
```
darh: why bool & string? Isn't empty sting implicit false?

server/system/service/smtp_configuration_checker.go
```diff
@@ -0,0 +1,215 @@
...
+	smtpConfigurationChecker struct {
+		actionlog actionlog.Recorder
+		settings  *types.AppSettings
+		ts        TemplateService
+		opt       options.AuthOpt
+	}
+)
+
+const (
+	hostCheckRE = "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
```
darh: move this to pkg/mail

server/system/service/smtp_configuration_checker.go
```diff
@@ -0,0 +1,215 @@
...
+			return fmt.Errorf("invalid recipient email address %s", email)
+		}
+
+	}
+
+	m.SetHeader(field, recipients...)
+	return nil
+}
+
+func (svc smtpConfigurationChecker) isValidHost(host string) bool {
```
darh: - move this to pkg/mail as a standalone function, export it
 - add some tests (look at `mail.TestMailValidator` for inspiration)

darh (changes requested):
Comments added.

And please, write at least some basic description of the functions.

client/web/admin/src/views/System/Email/Index.vue
```diff
@@ -113,6 +114,45 @@ export default {
...
+        host: server.host,
+        port: parseInt(server.port),
+        recipients: recepients,
+        username: server.user,
+        password: server.pass,
+        tlsInsecure: server.tlsInsecure,
+        tlsServerName: server.tlsServerName,
+      })
+        .then(response => {
+          console.log(response)
```
Fajfa: Console log

client/web/admin/src/views/System/Email/Index.vue
```diff
@@ -113,6 +114,45 @@ export default {
...
+          if (Object.values(response).every(resp => resp === '')) {
+            this.animateSuccess('external')
+            this.toastSuccess(this.$t('notification:settings.system.smtpCheck.success'))
+          }
+
+          let message = ''
+          for (var key in response) {
```
Fajfa: no vars please

client/web/admin/src/views/System/Email/Index.vue
```diff
@@ -113,6 +114,45 @@ export default {
...
+            if (response[key] !== '') {
```
Fajfa: Do we want to catch undefined null here?

client/web/admin/src/components/Settings/Email/CSystemEmailServer.vue
```diff
@@ -172,6 +186,9 @@ export default {
     submit () {
       this.$emit('submit', this.server)
     },
+    smtpConnectionCheck () {
```
Fajfa, line 190: Empty line above

Fajfa (changes requested):

server/system/types/smtp_configuration.go
```diff
@@ -0,0 +1,19 @@
...
+		Host          string
+		Port          uint
+		Recipients    []string
+		Username      string
+		Password      string
+		TLSInsecure   bool
+		TLSServerName string
+	}
+
+	SmtpCheckResult struct {
```
darh, line 16: add a comment, explain a bit how it is used

darh (changes requested):
One small fix and it is good to go.

tjerman (approved):
lgtm :+1:

PR #547, opened 2022-11-22
Supports coalesce sorting

### LINKED ISSUE #468: Sorting resource by "last-change" does not work as expected
As reported by @Bojan-Svirkov 

> I just came a cross this case, so when you are for example in the namespace list in compose and you sort them by last change, the order is mixed up and the namespaces are not sorted in the correct order. Same thing for workflows, when I sort them by `Updated at` you can see from the video that the order is mixed up, we have entries one by one September, October and then August which cannot be, they need to be in the correct order one by one. Currently I just noticed this on Compose and Workflows but it is possible to be present also on other places. I see that some of the formats where changed from `11/09/2022`, to `Wednesday, November 9, 2022 at 7:01 AM`, so this could be also the mix up. Also from the screenshot you can see that the order is mixed up.

Image

---

@Fajfa 

> we sort by updatedAt but show either updatedAt || createdAt
> 
> If we want a proper solution we need to discuss it a bit, so next sprint.

----

- [x] NS list
- [x] Module list
- [x] Charts
- [x] WF list
- [x] Other places?

vicpatel:
Add helper method to support coalesce sorting for multiple columns

Please take a look.

### DIFF
--- a/client/web/compose/src/mixins/listHelpers.js
+++ b/client/web/compose/src/mixins/listHelpers.js
@@ -10,7 +10,7 @@ export default {
       filter: {},
 
       pagination: {
-        limit: 10,
+        limit: 12,
         pageCursor: undefined,
         prevPage: '',
         nextPage: '',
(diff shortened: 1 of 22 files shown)


darh (changes requested):
where are the tests?
use `exp.NewSQLFunctionExpression`

darh (approved):
This still needs a lot of work. There is no modifier validation, and many places are very naive. 
Create a new issue and address these in the following weeks.

tjerman (approved):
lgtm

PR #576, opened 2022-11-25
2023.3.x feature magnify page blocks

### LINKED ISSUE #448: Allow end-users to open records in a modal window [PlanID:2023.3/2/4.3]
Record list can be configured to open a modal window where an entire record page can be opened in view or edit mode. 

Image
Record list configuration 

 - [x] Change record list configurator to support one of three modes (same window, new tab, modal)
 - [x] Ensure old config (new tab) is gracefully handled (respected when loaded and converted on change/save)
 - [x] Implement opening record page in a modal
 - [x] Ensure all functions (delete, cone, edit, add-new) work

### LINKED ISSUE #450: Allow end users to magnify certain Compose page blocks [PlanID:2023.3/18/2.7]
Integrators can configure certain block types to allow magnification, showing the content of the block inside a modal window or in fullscreen mode (not an actual new page, just extending block boundaries to the edge of the window.

Image

 - [x] Add a dropdown element with options to a general tab on the existing block 
   - [x] Options: Disabled, Show in a modal window, Show in a full window
   - [x] Block types where magnification is enabled:
     - [x] calendar
     - [x] chart
     - [x] content
     - [x] comments
     - [x] file
     - [x] iframe
     - [x] record list
     - [x] report
     - [x] record revisions
 - [x] Add support for modal window 
 - [x] Add support for full window mode
 - [x] Show magnifying glass 🔍 icon in the top right corner of the block with enabled magnification
 - [x] Preserve the magnification status of a block in the URL; when the user refreshes the browser, Compose must magnify the block again

yonkov:

Integrators can configure certain block types to allow magnification, showing the content of the block inside a modal window or in fullscreen mode.

image

image

### What was added?
The administrators can now add an option to magnify certain page blocks when viewing a public Compose page. In order to do this, a select dropdown has been added in the general tab of page block configuration modal for the following blocks: calendar, chart, content, comments, file, iframe, record list, report and record revisions. If the administrator has selected the Fullwidth or Modal select option, when the user goes to a public page, the user will see a magnify icon at the top of the block. Upon clicking on that icon, he will be able to magnify the block to a regular modal or fullwidth modal.

### How was added?
Magnify icon has been added to the page block wrap component and a select dropdown has been added to the block configurator. An event is being emitted from the wrap component, cointaining the block id. Upon clicking to the magnifying icon, the event is being passed and the modal is triggered. It gets the block id from the event and the page and namespace as props from the current page. In this way, the correct  block is being displayed inside the modal.

### What was improved?
Adding an option to magnify certain page blocks increases user experience and user satisfaction. This option should help users to find the info they are looking for much faster and easier.


client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -172,6 +182,14 @@ export default {
     isNew () {
       return this.block.blockID === NoID
     },
+
+    magnifyOptions () {
+      return [
+        { value: 'disabled', text: this.$t('general.magnifyOptions.disabled') },
```
Fajfa: Change the value of this to '' and you can have an easier time in the other conditions.

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -105,6 +117,10 @@ export default {
       ]
     },
 
+    hasMagnifyIcon () {
+      return this.block.options.magnifyOption ? !(this.block.options.magnifyOption === 'disabled') : false
```
Fajfa: For example this can be a much nicer condition if your "disabled" value is ''. Think about it a bit

You should also hide this if the block is already magnified

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
+<template>
```
Fajfa: Wouldn't this be located in `/Public/Page/Block` ?
Since its magnifying a block not the page

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+    this.$root.$off('magnify-page-block')
+  },
+
+  methods: {
+    showModal (blockID) {
+      if (blockID) {
+        this.blockID = blockID
+        this.block = this.page.blocks.find(block => block.blockID === blockID)
+        this.showBlockModal = true
+        this.modalClass = this.block.options.magnifyOption
```
Fajfa: modal class should be a computed property, since you already are saving the block

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+  },
+
+  beforeDestroy () {
+    this.$root.$off('magnify-page-block')
+  },
+
+  methods: {
+    showModal (blockID) {
+      if (blockID) {
+        this.blockID = blockID
```
Fajfa: Why save blockID separately? All you need is the block

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <b-modal
+    :id="modalClass === 'fullscreen' ? 'page-block-fullwidth-modal' : 'page-block-modal'"
```
Fajfa: No need for id please. Use the :class to achieve this. Return it with the modalClass computed property

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+    >
+      <page-block
+        :bounding-rect="boundingRect"
+        :block="block"
+        v-bind="$props"
+        v-on="$listeners"
+      />
+    </template>
+
+    <template #modal-footer>
```
Fajfa: Why do we need a footer here?

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+  <b-modal
+    :id="modalClass === 'fullscreen' ? 'page-block-fullwidth-modal' : 'page-block-modal'"
+    v-model="showBlockModal"
+    scrollable
+    body-class="p-0"
+    footer-class="p-0"
+    size="xl"
+    @hidden="hideModal"
+  >
+    <template
```
Fajfa: This template doesn't do anything

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+
+  methods: {
+    showModal (blockID) {
+      if (blockID) {
+        this.blockID = blockID
+        this.block = this.page.blocks.find(block => block.blockID === blockID)
+        this.showBlockModal = true
+        this.modalClass = this.block.options.magnifyOption
+        this.$router.replace({
+          query: { blockID: this.blockID },
```
Fajfa: Why use this.blockID when you can just use blockID

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -5,6 +5,18 @@
       class="h-100 border-0 shadow"
       :class="blockClass"
     >
+      <b-button
+        v-if="hasMagnifyIcon"
+        variant="link"
+        :title="$t('general.label.magnify')"
+        size="md"
+        class="ml-auto"
+        @click="handleModalOpen"
```
Fajfa: More fitting name please (openBlockInModal) for example.
Or something even better

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -5,6 +5,18 @@
...
+      >
+        <font-awesome-icon
+          :icon="['fas', 'search-plus']"
+        />
+      </b-button>
```
Fajfa: How about that empty line

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+    body-class="p-0"
+    footer-class="p-0"
+    size="xl"
+    @hidden="hideModal"
+  >
+    <template
+      slot-scope="{ boundingRect }"
+    >
+      <page-block
+        :bounding-rect="boundingRect"
```
Fajfa: Do you need this?

lib/js/src/compose/types/page-block/base.ts
```diff
@@ -27,7 +27,8 @@ export class PageBlock {
   xywh: number[] = defaultXYWH
   kind = ''
 
-  public options = {}
+  public options = { }
```
Fajfa: no need for extra space between {}

client/web/compose/src/components/Public/Page/Modal.vue
```diff
@@ -0,0 +1,106 @@
...
+}
+
+</script>
+
+  <style>
+  #page-block-fullwidth-modal .modal-dialog {
+    height: 100%;
+    max-width: 100vw;
+  }
+  #page-block-modal .modal-dialog {
```
Fajfa: So both should be height 100% right? Only the max width should change with class

Also fullscreen means fullscreen!

lib/js/src/compose/types/page-block/calendar/page-block.ts
```diff
@@ -36,6 +36,7 @@ class CalendarOptions {
   public header: Partial<CalendarOptionsHeader> = {}
   public locale = 'en-gb'
   public refreshRate = 0
+  public magnifyOption = 'disabled'
```
Fajfa: '' should be the default

Fajfa (changes requested):

client/web/compose/src/components/Public/Page/Modal.vue
yonkov: I do not think you can use `:class` prop on a b-modal. I tried but it did not work.

client/web/compose/src/components/Public/Page/Modal.vue
yonkov: If I remove this, I get `[Vue warn]: Error in render: "Error: block prop missing"`

client/web/compose/src/components/Public/Page/Block/Modal.vue
```diff
@@ -0,0 +1,89 @@
...
+
+  data () {
+    return {
+      showBlockModal: false,
+      block: undefined,
+    }
+  },
+
+  computed: {
+    modalClass () {
```
Fajfa: Instead of having this computed, have computed for the dialog and content class

client/web/compose/src/components/Public/Page/Block/Modal.vue
```diff
@@ -0,0 +1,89 @@
...
+        this.block = this.page.blocks.find(block => block.blockID === blockID)
+        this.showBlockModal = true
+        this.$router.replace({
+          query: { blockID },
+        })
+      }
+    },
+
+    hideModal () {
+      this.$router.replace({ query: { } })
```
Fajfa: The extra space inside the {} is not needed

client/web/compose/src/components/PageBlocks/Wrap/Plain.vue
```diff
@@ -4,6 +4,19 @@
       class="d-flex flex-column position-relative h-100 border-0"
       :class="blockClass"
     >
+      <b-button
```
Fajfa: Look at line 43 in the same file and do it like that please, since its pretty much the same kind of function.

And same in plain

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -184,6 +184,7 @@ export default {
     },
 
     magnifyOptions () {
+      const disabledOption = this.block.options.magnifyOption === NoID ? NoID: '';
```
Fajfa: No need for this no?

client/web/compose/src/components/Public/Page/Block/Modal.vue
```diff
@@ -81,6 +82,10 @@ export default {
           query: { blockID },
         })
       }
+      else {
+        this.showBlockModal = false
```
Fajfa: I think we can safely put this into hideModal

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -105,6 +116,14 @@ export default {
       ]
     },
 
+    hasMagnifyIcon () {
+      return this.block.options.magnifyOption
+    },
+
+    isBlockOpened () {
+      return this.block.id = this.$route.query.blockID
```
Fajfa: isnt it this.block.blockID? Same for Plain

PR #584, opened 2022-12-01
Add support for navigational page block

### LINKED ISSUE #449: Add support for navigational page block [PlanID:2023.3/11/2.7]
Introduce new page block type for showing one or two level of dynamic and static navigation 

Inspired by bootstrap's [nav component](https://bootstrap-vue.org/docs/components/nav).

**List configuration**
 - [x] Add tab right of **general** - "Navigation"
 - [x] options for align, vertical, fill, pills, small, tabs
 - [x] Default properties for all items
   - [x] disabled
   - [x] text variant or colour
   - [x] background variant or colour
 - [x] List items can be sorted (drag & drop)

**Supported navigation item types**
 - [x] **Text**
 - [x] **Delimiter** (only inside dropdown)
 - [x] **URL**
   - [x] new window/tab
 - [x] Compose **page** in the same namespace
   - [x] new window/tab
   - [x] use `to` prop on `<b-nav-item>` to to ensure internal router is used
   - [x] select a page to be used, keep reference (ID)
   - [x] option to show list of child pages in-place (1 level only)
   - [x] option to show referenced page as dropdown with subpages 
 - [x] **Dropdown**
   - [x] Only **text** and **URL**
   - [x] If used in combination with compose page, ensure there is no 2nd level
   - [x] Support drop(left|right|up) prop
 
**Item types properties**
Each item can override the default (see above) item properties

 - [x] Ability to remove explicit property value (and use the default)

kelanik8:
# The following changes are implemented #449
Introduced new page block type for showing one or two level of dynamic and static navigation
Inspired by bootstrap's [nav component](https://bootstrap-vue.org/docs/components/nav).

image
image
image

 - [ ] Code builds
 - [ ] All existing tests pass


client/web/compose/src/components/PageBlocks/base.vue
```diff
@@ -69,9 +69,8 @@ export default {
       get () {
         return this.block.options
       },
-
-      set (options) {
-        this.block.options = options
+      set (value) {
```
Fajfa: Please revert this, it was fine as it was

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+      <b-nav
+        :tabs="options.display.appearance === 'tabs'"
+        :pills="options.display.appearance === 'pills'"
+        :small="options.display.appearance === 'small'"
+        :fill="options.display.fillJustify === 'fill'"
+        :justify="options.display.fillJustify === 'justify'"
+        :align="options.display.alignment"
+        class="border-0 h-100"
+      >
+        <b-nav-item
```
Fajfa: Why not v-for in a template and then display the correct item?

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+          :to="processLink(list)"
+          :href="processHrefLink(list)"
+          :disabled="list.options.disabled"
+          :variant="list.options.bgVariant"
+          :style="{ order: index, color: list.options.textColor }"
+          class="nav-list-item"
+          :link-attrs="{ style: `color: ${list.options.textColor}` }"
+          :target="list.options.targetType"
+        >
+          <template v-if="list.navigationType === 'text-type'">
```
Fajfa: Since all three ifs content is the same, do we really need the ifs?

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
```diff
@@ -0,0 +1,75 @@
...
+      type: Object,
+      required: true,
+    },
+  },
+
+  data () {
+    return {
+      openInType: [
+        { value: '_self', text: this.$t('navigation.sameTab') },
+        { value: '_blank', text: this.$t('navigation.newTab') },
```
Fajfa: Since the two values are the same, if I save new window it thinks I saved newTab. So you'll need something like Atanas did in record-list for the `recordDisplayOption`

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
```diff
@@ -0,0 +1,75 @@
...
+  },
+
+  watch: {
+    column: {
+      immediate: true,
+      handler (column) {
+        this.value = column
+      },
+    },
+    value (value) {
```
Fajfa: Why this watcher? You should add the event to the inputs

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
```diff
@@ -0,0 +1,75 @@
...
+          v-model="value.options.itemOption.url"
+          type="text"
+          size="sm"
+        />
+      </b-form-group>
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.openIn')">
+        <b-form-select
+          v-model="value.options.itemOption.targetType"
```
Fajfa: i think we can call the itemOption just item. And then hold all options inside it, same for dropdown

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Dropdown.vue
```diff
@@ -0,0 +1,138 @@
+<template>
+  <tr>
+    <td />
+
+    <td colspan="4" class="p-0">
+      <div class="d-flex">
+        <th>
+          <b-form-group :label="$t('navigation.fieldLabel')">
+            <b-form-input
+              v-model="value.options.dropdownText"
```
Fajfa: if you move this to options.item.label its much more consistent

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
...
+          v-model="value.options.itemOption.label"
+          type="text"
+          size="sm"
+        />
+      </b-form-group>
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.composePage')">
+        <b-form-select
+          v-model="value.options.itemOption.referenceId"
```
Fajfa: wouldn't this be the pageID?
Also it would probably make sense to replace the b-form-select with vue select.
You can look at namespace sidebar for a good example

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
...
+        { value: '_self', text: this.$t('navigation.sameTab') },
+        { value: '_blank', text: this.$t('navigation.newTab') },
+        { value: '_blank', text: this.$t('navigation.newWindow') },
+      ],
+    }
+  },
+
+  computed: {
+    composePageList () {
+      return this.tree.map(t => ({
```
Fajfa: If you replace select with vue-select you wont need this aswell

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
...
+      required: true,
+    },
+  },
+
+  data () {
+    return {
+      tree: [],
+      openInType: [
+        { value: '_self', text: this.$t('navigation.sameTab') },
+        { value: '_blank', text: this.$t('navigation.newTab') },
```
Fajfa: Since the two values are the same, if I save new window it thinks I saved newTab. So you'll need something like Atanas did in record-list for the `recordDisplayOption`

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
+<template>
```
Fajfa: This type should probably be ComposePage no? Since Compose is the whole app

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
...
+              {{ $t('navigation.none') }}
+            </b-form-select-option>
+          </template>
+        </b-form-select>
+      </b-form-group>
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.openIn')">
+        <b-form-select
+          v-model="value.options.itemOption.targetType"
```
Fajfa: maybe just target instead of targetType

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
```diff
@@ -0,0 +1,141 @@
...
+          :options="openInType"
```
Fajfa: targetTypes?

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+      fillJustifyOptions: [
+        { value: 'fill', text: this.$t('navigation.fill') },
+        { value: 'justify', text: this.$t('navigation.justify') },
+      ],
+
+      bgVariants: [
+        { value: 'primary', text: this.$t('navigation.primary') },
+        { value: 'secondary', text: this.$t('navigation.secondary') },
+        { value: 'success', text: this.$t('navigation.success') },
+        { value: 'warning', text: 'Warning' },
```
Fajfa: Lets translate this aswell

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+export default {
+  i18nOptions: {
+    namespaces: 'block',
+  },
+
+  components: { Draggable, TextType, UrlType, ComposeType, DropdownType },
+
+  extends: base,
+
+  inheritAttrs: false,
```
Fajfa: What dis?

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+import UrlType from './NavTypes/Url.vue'
+import ComposeType from './NavTypes/Compose.vue'
+import DropdownType from './NavTypes/Dropdown.vue'
+
+export default {
+  i18nOptions: {
+    namespaces: 'block',
+  },
+
+  components: { Draggable, TextType, UrlType, ComposeType, DropdownType },
```
Fajfa: Expand this like all other components do

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+                          type="color"
+                          debounce="300"
+                          class="w-100"
+                        />
+                      </b-form-group>
+                    </td>
+                    <td>
+                      <b-form-group>
+                        <b-form-select
+                          v-model="column.options.bgVariant"
```
Fajfa: If you look at metric we have text color and background color, might make sense to do that here aswell to just keep it consistent. Sorry for late change

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+          >
+            <div
+              v-for="(column, index) in block.options.lists"
+              :key="`column-${index}`"
+            >
+              <hr v-if="index" />
+
+              <table
+                class="
+                  table table-sm table-borderless table-responsive-lg
```
Fajfa: if you use b-table-simple i think you can replace some of those classes with props, do check it out

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+
+        <div class="mt-3">
+          <draggable
+            v-model="block.options.lists"
+            group="sort"
+            handle=".grab"
+          >
+            <div
+              v-for="(column, index) in block.options.lists"
+              :key="`column-${index}`"
```
Fajfa: Just index is ok here i think

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+    processHrefLink (list) {
+      if (list.navigationType === 'url') {
+        return list.options.itemOption.url
+      }
+
+      return ''
+    },
+
+    getSubPages (selfID) {
+      return [...this.$store.getters['page/set']].filter(value => value.selfID === selfID)
```
Fajfa: Lets use the mapGetters

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+                    <td class="align-middle">
+                      <font-awesome-icon
+                        :icon="['fas', 'bars']"
+                        class="grab text-grey"
+                      />
+                    </td>
+                    <td>
+                      <b-form-group>
+                        <b-form-select
+                          v-model="column.navigationType"
```
Fajfa: Maybe just type

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+          </template>
+        </b-nav-item>
+
+        <b-nav-item-dropdown
+          v-for="(list, index) in options.lists"
+          v-show="list.navigationType === 'dropdown-type'"
+          :key="`lists-dropdown-${index}`"
+          :text="list.options.dropdownText"
+          :variant="list.options.bgVariant"
+          :right="list.options.dropdownAlignType === 'right'"
```
Fajfa: dropdown.align i think

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+          <template v-if="(list.navigationType === 'compose-type' && !list.options.itemOption.displaySubPages)">
+            <span>
+              {{ list.options.itemOption.label }}
+            </span>
+          </template>
+        </b-nav-item>
+
+        <b-nav-item-dropdown
+          v-for="(list, index) in options.lists"
+          v-show="list.navigationType === 'dropdown-type'"
```
Fajfa: dont suffix with -type, just 'dropdown', 'compose', 'url'

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
+<template>
+  <wrap
+    v-bind="$props"
+    v-on="$listeners"
+  >
+    <div class="h-100 w-100 navigation-page-block-item">
```
Fajfa: replace navigation-page-block-item with actual bootstrap classes. Remember we want to avoid new classes as much as possible

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+        <b-nav-item
+          v-for="(list, index) in options.lists"
+          v-show="list.options.itemOption.label"
+          :key="`lists-${index}`"
+          :to="processLink(list)"
+          :href="processHrefLink(list)"
+          :disabled="list.options.disabled"
+          :variant="list.options.bgVariant"
+          :style="{ order: index, color: list.options.textColor }"
+          class="nav-list-item"
```
Fajfa: If you really need this custom class then try to write is as bootstrap class not by introducing a new one

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,136 @@
...
+      if (list.navigationType === 'compose') {
+        return list.options.itemOption.referenceId
+      }
+
+      return ''
+    },
+
+    processHrefLink (list) {
+      if (list.navigationType === 'url') {
+        return list.options.itemOption.url
```
Fajfa: wouldn't the item.url be undefined anyway so you can just do return item.url.
You can add || '' if you need it to be a string

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+  <div>
+    <b-tab :title="$t('navigation.label')">
+      <div class="mb-3">
+        <h5 class="text-primary">
+          {{ $t("navigation.displayOptions") }}
+        </h5>
+
+        <b-row class="justify-content-between">
+          <b-col
+            cols="6"
```
Fajfa: a bit more responsive please, cols=12 is always a safe bet, then md the final layout. You can start from there

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+            <b-form-group
+              horizontal
+              :label="$t('navigation.fillJustify')"
+            >
+              <b-form-radio-group
+                v-model="options.display.fillJustify"
+                buttons
+                button-variant="outline-secondary"
+                size="sm"
+                name="fill-justify"
```
Fajfa: Do we need name? Same for other `name=` in this PR

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,300 @@
...
+      <div class="mb-3 mt-2">
+        <div class="d-flex align-items-center mb-4">
+          <h5 class="text-primary mb-0">
+            {{ $t("navigation.listItems") }}
+          </h5>
+
+          <b-button
+            variant="link"
+            class="d-flex align-items-center text-decoration-none"
+            @click="handleAddButton"
```
Fajfa: More appropriate name please, like add navigation item

Fajfa (changes requested):
There is other changes necessary but lets start with these

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
kelanik8: I'm using a v-for here

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
kelanik8: the value for new window and new tab is the same for the `target` value, I don't seem to know why. Atanas implementation does not have a new window

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Compose.vue
kelanik8: the value for new window and new tab is the same for the target value, I don't seem to know why. Atanas implementation does not have a new window

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
kelanik8: the reason I made use of the class is because I could not easily edit the styling on the nav list item to take the full height of the block, and I did not want to make the stlyes get affected by something else i the app, `scoped` is not helping because its a component
image

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
Fajfa: True, but that doesnt change the fact that your values shouldn't be _self and _blank, but sameTab, newTab, newWindow

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
Fajfa: Not sure what you mean, you can control the class where you call the b-nav-item with the class property

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
kelanik8: I tried v-for in a template but there is an error returned by vue, I was trying to implement something like the field editor but still coudlnt work

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
kelanik8: I made use this update the value to the parent, instead of adding change methods to the input

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
kelanik8: I have moved the class to the `b-nav` component

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,125 @@
...
+
+  computed: {
+    ...mapGetters({
+      pages: 'page/set',
+    }),
+  },
+
+  methods: {
+    displayNavItemDrop (list) {
+      if (list.options.item.label && (!list.options.item.displaySubPages && list.type !== 'dropdown')) {
```
Fajfa: If you think about it you can improve the conditions here a bit. since two of the conditions share a variable.
list type !== 'dropdown' and list.type === 'dropdown', also the displaySubPages is used in two conditions, so else-if might come in handy. The same for template ifs (line 32-66)

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      if (list.type === 'dropdown') {
+        return 'b-nav-item-dropdown'
+      }
+
+      if (list.type === 'compose' && list.options.item.displaySubPages) {
+        return 'b-nav-item-dropdown'
+      }
+    },
```
Fajfa: Lets respect new lines between methods

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Url.vue
Fajfa: If you abandon the value, and just use column and don't model but emit update instead then it should work the same with less code, same for other types

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,125 @@
...
+          :style="{ order: index, color: list.options.textColor }"
+          :link-attrs="{ style: `color: ${list.options.textColor}` }"
+          :target="selectTargetOption(list.options.item.target)"
+          :text="list.options.item.label"
+          :right="list.options.align === 'right'"
+          :left="list.options.align === 'left'"
+          class="nav-item-dropdown"
+        >
+          <template v-if="!list.options.item.displaySubPages && list.type !== 'dropdown'">
+            <span>
```
Fajfa: I dont think span is needed here

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+                      <font-awesome-icon
+                        :icon="['fas', 'bars']"
+                        class="grab text-grey"
+                      />
+                    </td>
+                    <td class="align-middle">
+                      <b-form-select
+                        v-model="column.type"
+                        :options="navigationItemTypes"
+                        size="sm"
```
Fajfa: To make it look consistent (vue-select problems), just change all inputs and buttons sizes to 'md'. Remember bootstrap things default to md, others might not. Same for the NavTypes

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+                    </td>
+                  </tr>
+
+                  <component
+                    :is="column.type"
+                    :column="column"
+                    :namespace="namespace"
+                    @update="(value) => column = value"
+                  />
+                  <Text />
```
Fajfa: Do we need this text?

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+      </div>
+    </b-tab>
+  </div>
+</template>
+
+<script>
+import base from '../base'
+import Draggable from 'vuedraggable'
+import { compose } from '@cortezaproject/corteza-js'
+import TextSection from './NavTypes/Text.vue'
```
Fajfa: Name consistency with other types, call it Text

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+                  <tr>
+                    <td class="align-middle">
+                      <font-awesome-icon
+                        :icon="['fas', 'bars']"
+                        class="grab text-grey"
+                      />
+                    </td>
+                    <td class="align-middle">
+                      <b-form-select
+                        v-model="column.type"
```
Fajfa: there should be no None value, only valid values, the Compose Page should be selected as default (in the type definition)

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,125 @@
...
+              v-for="(dropdown, dIndex) in list.options.item.dropdown"
+              :key="`dropdown-${dIndex}`"
+              :to="dropdown.url"
+              :disabled="list.options.disabled"
+              :style="{ order: dIndex * 2 }"
+            >
+              {{ dropdown.label }}
+            </b-dropdown-item>
+
+            <b-dropdown-divider
```
Fajfa: how about a template v-for with a b-dropdown-item and b-dropdown-divider inside, again same as above, one v-for instead of two

lib/js/src/compose/types/page-block/navigation/list.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type ItemOptions = {
+    label: string;
+    url: string;
+    target: string;
+    delimiter: boolean;
+    pageID: string;
+    displaySubPages: boolean;
+    dropdown: Dropdown[];
```
Fajfa: wouldnt it make sense to have a `dropdown: { label, items }` kind of structure

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+              />
+            </b-form-group>
+          </b-col>
+        </b-row>
+      </div>
+
+      <div class="mb-3 mt-2">
+        <div class="d-flex align-items-center mb-4">
+          <h5 class="text-primary mb-0">
+            {{ $t("navigation.listItems") }}
```
Fajfa: in my opinion these aren't really "List items", they are more like navigation items, that would be the same for the class and properties in corteza-js. You could just call it Items, and since they are in navigation it makes sense they are navigation items

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,291 @@
...
+        </div>
+
+        <div class="mt-3">
+          <draggable
+            v-model="block.options.lists"
+            group="sort"
+            handle=".grab"
+          >
+            <div
+              v-for="(column, index) in block.options.lists"
```
Fajfa: Same here, is this a column? Or is this an item?

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/ComposePage.vue
```diff
@@ -0,0 +1,130 @@
...
+    <td
+      v-if="selectedPageChildren(value.options.item.pageID)"
+      cols="12"
+      sm="6"
+      class="align-middle text-center"
+    >
+      <b-form-group class="m-0">
+        <b-form-checkbox
+          v-model="value.options.item.displaySubPages"
+          name="enabled-button"
```
Fajfa: No need for names (remove from others aswell), also do we really need text-left class (we don't just remove text-center)?

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/ComposePage.vue
```diff
@@ -0,0 +1,130 @@
...
+          switch
+          class="text-left"
+        >
+          {{ $t('navigation.showReferencedPage') }}
```
Fajfa: Text should just read, "Display sub-pages"

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/ComposePage.vue
```diff
@@ -0,0 +1,130 @@
...
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.composePage')">
+        <vue-select
+          key="pageID"
+          v-model="value.options.item.pageID"
+          :placeholder="$t('navigation.none')"
+          :options="tree"
+          class="bg-white"
+          label="title"
```
Fajfa: append-to-body prop will be needed!

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/ComposePage.vue
```diff
@@ -0,0 +1,130 @@
...
+      </b-form-group>
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.composePage')">
+        <vue-select
+          key="pageID"
+          v-model="value.options.item.pageID"
+          :placeholder="$t('navigation.none')"
+          :options="tree"
+          class="bg-white"
```
Fajfa: class is last prop before events!

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/ComposePage.vue
```diff
@@ -0,0 +1,130 @@
...
+          option-text="title"
+          @input="updateLabelValue"
+        />
+      </b-form-group>
+    </td>
+    <td>
+      <b-form-group :label="$t('navigation.openIn')">
+        <b-form-select
+          v-model="value.options.item.target"
+          :options="openInType"
```
Fajfa: openInType makes me think this holds the value of what is selected. 
Try to keep naming convention of appending options to the end of the variable of the options

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,125 @@
...
+  },
+}
+</script>
+
+<style>
+.nav-item-dropdown .dropdown-menu.show {
+  display: grid;
+}
+
+.navigation-page-block-item .nav-item {
```
Fajfa: Same as the above class, you can have it on the component like you do now. So you just need one class

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
kelanik8: because there seem to be an existing html tag named `<text>` vue throws an error in the console because the component and the html tag has the same name and it won't render

Fajfa: You can still import it as text and just define it with text-section for example

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
kelanik8: I've already remove the none value below, I dont' seem to why it;s not saying `outdated`

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,286 @@
...
+          >
+            <div
+              v-for="(column, index) in block.options.navigationItems"
+              :key="index"
+            >
+              <hr v-if="index">
+
+              <b-table-simple
+                borderless
+                class="nav-display-options-items"
```
Fajfa: Do we still need this?

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
Fajfa: I see you changed them, but like I said above, bootstrap inputs default to md!
Also the radio group should stay sm

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
Fajfa: There is still names on inputs

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,274 @@
...
+                        type="color"
+                        debounce="300"
+                        class="w-100"
+                      />
+                    </td>
+                    <td class="align-middle text-center">
+                      <b-form-checkbox
+                        v-model="item.options.enabled"
+                        switch
+                        size="sm"
```
Fajfa: remove this (size should be md, but that is default so you can just remove it)

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,274 @@
...
+                        class="mb-0"
+                      />
+                    </td>
+                    <td class="align-middle">
+                      <c-input-confirm
+                        size="lg"
```
Fajfa: `size="md"` here, also add `button-class=px-2`

client/web/compose/src/components/PageBlocks/Navigation/Base.vue
```diff
@@ -0,0 +1,158 @@
...
+
+    displayDropdownText (navItem) {
+      if (navItem.type === 'dropdown') {
+        return navItem.options.item.dropdown.label
+      }
+
+      return navItem.options.item.label
+    },
+
+    toggleToLink (navItem) {
```
Fajfa: all good, but can you not call it toggle, since that implies you're toggling the value.. 
It should be more like, getToLink, getHrefLink

client/web/compose/src/components/PageBlocks/Navigation/NavTypes/Dropdown.vue
```diff
@@ -0,0 +1,156 @@
...
+                  v-model="item.delimiter"
+                  switch
+                  size="sm"
+                />
+              </b-form-group>
+            </th>
+
+            <th class="align-middle text-center">
+              <c-input-confirm
+                button-class="px-0"
```
Fajfa: remove this please

client/web/compose/src/components/PageBlocks/Navigation/Configurator.vue
```diff
@@ -0,0 +1,274 @@
...
+        }),
+      )
+    },
+  },
+}
+</script>
+
+<style lang="scss" scoped>
+th {
+  width: 25%;
```
Fajfa: I dont think this is needed

Fajfa (changes requested):

Fajfa (approved):

PR #618, opened 2022-12-08
Integration gateway response support

petergrlica:
Ability to return a meaningful response via integration gateway's `Response` postfilter.
This is the backend implementation that might change a bit, depending on the frontend work.

Ref: https://github.com/cortezaproject/corteza/issues/517


petergrlica:
closing to rename branch

PR #632, opened 2022-12-13
Add support for helper env-variable that create super users

### LINKED ISSUE #527: Add support for helper env-variable that creates a user and logs you in
When a user sets `AUTH_PROVISION_SUPER_USER`, Corteza creates one or more users with the configured values using provided email as a password. It skips existing (email, handle).  All new users are assigned to all bypass roles.

Value format:
```
<users> = <user>[;<user>]
<user> = <email>[,<handle>[,<name>]
<email> valid email string
<handle> valid handle
```

Examples:
```
AUTH_PROVISION_SUPER_USER=name@corteza.tld;name2@corteza.tld;name3@corteza.tld
AUTH_PROVISION_SUPER_USER=name@corteza.tld;name,name2@corteza.tld;name2,Foo Bar,name4@corteza.tld
```

All actions (created, existing-skipped) must be logged as warnings.

If the command is used with `ENVIRONMENT=production` Corteza should stop and report an error.

KinyaElGrande:
When a user sets `AUTH_PROVISION_SUPER_USER`  Corteza creates one or more users with the configured values using provided email as a password,  existing (email, handle) are skipped and all new users are assigned to all bypass roles.
All actions (created, existing-skipped) are logged as warnings.

When `ENVIRONMENT=production`, Corteza stops and reports an error.

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -39,7 +39,7 @@
 
 ###############################################################################
 # Allow insecure (invalid, expired TLS/SSL certificates) connections.
-#
+# 
 # [IMPORTANT]
 # ====
 # We strongly recommend keeping this value set to false except for local development or demos.
@@ -209,7 +209,7 @@
 
 ###############################################################################
 # Password for the web console endpoint. When running in dev environment, password is not required.
-#
+# 
 # Corteza intentionally sets default password to random chars to prevent security incidents.
 # Type:    string
 # Default: <no value>
@@ -296,7 +296,7 @@
 # Email sending
 #
 # Configure your local SMTP server or use one of the available providers.
-#
+# 
 # These values are copied to settings when the server starts and can be managed from the administration console.
 # We recommend you remove these values after they are copied to settings.
 # If server detects difference between these options and settings, it shows a warning in the log on server start.
@@ -436,7 +436,7 @@
 
 ###############################################################################
 # Password security allows you to disable constraints to which passwords must conform to.
-#
+# 
 # [CAUTION]
 # ====
 # Disabling password security can be useful for development environments as it removes the need for complex passwords.
@@ -448,12 +448,12 @@
 
 ###############################################################################
 # Algoritm to be use for JWT signature.
-#
+# 
 # Supported valus:
 #  - HS256, HS384, HS512
 #  - PS256, PS384, PS512,
 #  - RS256, RS384, RS512
-#
+# 
 # Provide shared secret string for HS256, HS384, HS512 and full private key or path to the file PS* and RS* algorithms.
 # Type:    string
 # Default: HS512
@@ -462,7 +462,7 @@
 ###############################################################################
 # Secret used for signing JWT tokens.
 # Value is used only when HS256, HS384 or HS512 algorithm is used.
-#
+# 
 # [IMPORTANT]
 # ====
 # If secret is not set, system auto-generates one from DB_DSN and HOSTNAME environment variables.
@@ -486,7 +486,7 @@
 
 ###############################################################################
 # Lifetime of the refresh token. Should be much longer than lifetime of the access token.
-#
+# 
 # Refresh tokens are used to exchange expired access tokens with new ones.
 # Type:    time.Duration
 # Default: 72h
@@ -494,7 +494,7 @@
 
 ###############################################################################
 # Redirect URL to be sent with OAuth2 authentication request to provider
-#
+# 
 # `provider` placeholder is replaced with the actual value when used.
 # Type:    string
 # Default: <no value>
@@ -502,7 +502,7 @@
 
 ###############################################################################
 # Secret used for securing cookies
-#
+# 
 # [IMPORTANT]
 # ====
 # If secret is not set, system auto-generates one from DB_DSN and HOSTNAME environment variables.
@@ -545,21 +545,21 @@
 
 ###############################################################################
 # Maximum time user is allowed to stay idle when logged in without "remember-me" option and before session is expired.
-#
+# 
 # Recomended value is between an hour and a day.
-#
+# 
 # [IMPORTANT]
 # ====
 # This affects only profile (/auth) pages. Using applications (admin, compose, ...) does not prolong the session.
 # ====
-#
+# 
 # Type:    time.Duration
 # Default: 24h
 # AUTH_SESSION_LIFETIME=24h
 
 ###############################################################################
 # Duration of the session in /auth lasts when user logs-in with "remember-me" option.
-#
+# 
 # If set to 0, "remember-me" option is removed.
 # Type:    time.Duration
 # Default: 8640h
@@ -586,7 +586,7 @@
 
 ###############################################################################
 # Secret used for securing CSRF protection
-#
+# 
 # [IMPORTANT]
 # ====
 # If secret is not set, system auto-generates one from DB_DSN and HOSTNAME environment variables.
@@ -616,19 +616,19 @@
 
 ###############################################################################
 # Handle for OAuth2 client used for automatic redirect from /auth/oauth2/go endpoint.
-#
+# 
 # This simplifies configuration for OAuth2 flow for Corteza Web applications as it removes
 # the need to suply redirection URL and client ID (oauth2/go endpoint does that internally)
-#
+# 
 # Type:    string
 # Default: corteza-webapp
 # AUTH_DEFAULT_CLIENT=corteza-webapp
 
 ###############################################################################
 # Path to js, css, images and template source files
-#
+# 
 # When corteza starts, if path exists it tries to load template files from it.
-#
+# 
 # When empty path is set (default value), embedded files are used.
 # Type:    string
 # Default: <no value>
@@ -637,11 +637,20 @@
 ###############################################################################
 # When enabled, corteza reloads template before every execution.
 # Enable this for debugging or when developing auth templates.
-#
+# 
 # Should be disabled in production where templates do not change between server restarts.
 # Type:    bool
 # Default: <no value>
 # AUTH_DEVELOPMENT_MODE=<no value>
+
+###############################################################################
+# When set, Corteza creates one or more users with the configured values using provided email as a password.
+# It skips existing (email, handle). All new users are assigned to all bypass roles.
+# 
+# When set in production, Corteza stops and reports an error
+# Type:    string
+# Default: <no value>
+# AUTH_PROVISION_SUPER_USER=<no value>
 
 ###############################################################################
 ###############################################################################
@@ -815,17 +824,15 @@
 ###############################################################################
 # List of compa delimited languages (language tags) to enable.
 # In case when an enabled language can not be loaded, error is logged.
-#
+# 
 # When loading language configurations (config.xml) from the configured path(s).
-#
+# 
 # Type:    string
 # Default: en
 # LOCALE_LANGUAGES=en
 
 ###############################################################################
-# 	One or more paths to locale config and translation files, separated by colon
-#
-# 	When with LOCALE_DEVELOPMENT_MODE=true, default value for path is ../../locale
+# One or more paths to locale config and translation files, separated by colon
 # Type:    string
 # Default: <no value>
 # LOCALE_PATH=<no value>
@@ -834,7 +841,7 @@
 # Name of the query string parameter used to pass the language tag (it overrides Accept-Language header).
 # Set it to empty string to disable detection from the query string.
 # This parameter is ignored if only one language is enabled
-#
+# 
 # Type:    string
 # Default: lng
 # LOCALE_QUERY_STRING_PARAM=lng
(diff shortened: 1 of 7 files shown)


server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+
+// addAuthSuperUsers assigns BYPASS roles to users from AUTH_PROVISION_SUPER_USER value
+// When in Production, Corteza should stop and report an error.
+func addAuthSuperUsers(ctx context.Context, log *zap.Logger, s store.Storer, envOpt options.EnvironmentOpt) (err error) {
+	var (
+		val, has = os.LookupEnv("AUTH_PROVISION_SUPER_USER")
+	)
+
+	if envOpt.IsProduction() {
+		panic(errors.New(0, "Cannot provision super users in Production enviroment!"))
```
tjerman: No need to panic here; we can just return the err

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
 }
+
+// addAuthSuperUsers assigns BYPASS roles to users from AUTH_PROVISION_SUPER_USER value
+// When in Production, Corteza should stop and report an error.
+func addAuthSuperUsers(ctx context.Context, log *zap.Logger, s store.Storer, envOpt options.EnvironmentOpt) (err error) {
+	var (
+		val, has = os.LookupEnv("AUTH_PROVISION_SUPER_USER")
+	)
+
+	if envOpt.IsProduction() {
```
tjerman, line 72: This condition will cause Corteza to error out in production regardless if the env value is set or not

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
 		})
 	})
 }
+
+// addAuthSuperUsers assigns BYPASS roles to users from AUTH_PROVISION_SUPER_USER value
+// When in Production, Corteza should stop and report an error.
+func addAuthSuperUsers(ctx context.Context, log *zap.Logger, s store.Storer, envOpt options.EnvironmentOpt) (err error) {
+	var (
+		val, has = os.LookupEnv("AUTH_PROVISION_SUPER_USER")
```
tjerman: Can't we use the env opt struct that was generated for you?
OIDC provisioning did this because it had to be a bit more flexible.

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+	)
+
+	if envOpt.IsProduction() {
+		panic(errors.New(0, "Cannot provision super users in Production enviroment!"))
+	}
+
+	if has {
```
tjerman: Instead of saying

```
if something {
    some complex code
}
```

you can (should!!) do

```
if !something {
  return
}
some complex code
```

"Guard statements" let you flatten out the code which makes it more readable, safer, ...

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+
+		if len(usr) == 3 {
+			u.Username = usr[0]
+			u.Name = usr[1]
+			u.Email = usr[2]
+			return u
+		}
+
+		u.Username = usr[0]
+		u.Email = usr[1]
```
tjerman: Bold assumption that we can expect 2 elements; is that a requirement? If so, enforce it

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+	)
+
+	if envOpt.IsProduction() {
+		panic(errors.New(0, "Cannot provision super users in Production enviroment!"))
+	}
+
+	if has {
+		users := strings.Split(val, ";")
+
+		for _, user := range users {
```
tjerman: Re-think the code in this `for` loop -- I don't agree with it.
Spend an hour trying and figure it out yourself; if you can't, let me know.

Look at the comments in this `for` loop for some hints

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+			// skip existing email
+			_, err = s.LookupUserByEmail(ctx, user)
```
tjerman: This won't work for a lot of cases

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+			if err != store.ErrNotFound {
+				log.Warn(fmt.Sprintf("Email address already %s exists", user))
+				continue
+			}
+
+			u := newUser(user)
+
+			createUserHandle(u)
```
tjerman: Why is this not part of the `newUser` fnc? Should this instead return the handle value?

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,119 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+				return err
+			}
+
+			if err = setPassword(ctx, u.ID, u.Email, s); err != nil {
+				return err
+			}
+
+			log.Warn(fmt.Sprintf("User %#v created", u))
+
+			//assign the user a BYPASS role
```
tjerman: `BYPASS` -> `bypass`

tjerman (changes requested):

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,120 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+
+		log.Warn(fmt.Sprintf("User %#v created", u))
+
+		//assign the user a bypass role
+		for _, r := range internalAuth.BypassRoles() {
+			m := &types.RoleMember{UserID: u.ID, RoleID: r.ID}
+			if err = store.CreateRoleMember(ctx, s, m); err != nil {
+				return err
+			}
+		}
```
tjerman: Run in a transaction; `store.Tx`

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,120 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+		return u
+	}
+
+	if len(usr) > 1 {
+		u.Username = usr[1]
+		u.Handle = createUserHandle(u)
+		return u
+	}
+
+	u.Handle = createUserHandle(u)
```
tjerman: You misunderstood me.

In the lines of:

```go
handle email
if len(usr) > 1 {
  handle username
}
if len(usr) > 2 {
  handle the other bit
}
...

return u
```

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,120 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+		val, has = os.LookupEnv("AUTH_PROVISION_SUPER_USER")
+		envOpt   = options.Environment()
+	)
+
+	if !has {
+		return
+	}
+
+	if envOpt.IsProduction() {
+		return errors.New(0, "Cannot provision super users in Production environment!")
```
tjerman: Please be more instructive in regards to what the user tried to do and what they need to do to fix the issue -- in this case it's the dilemma with the different env variables.

Also, start error strings with a lower case character and don't punctuate it (especially not with `!`)

server/pkg/provision/auth.go
```diff
@@ -43,3 +61,120 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+		u.Username = usr[1]
+		u.Handle = createUserHandle(u)
+		return u
+	}
+
+	u.Handle = createUserHandle(u)
+	return u
+}
+
+func setPassword(ctx context.Context, userID uint64, password string, s store.Storer) (err error) {
```
tjerman: Did you check in the service code if there is any code you could reuse; perhaps a function call for this?

tjerman (changes requested):

server/pkg/provision/auth.go
```diff
@@ -43,3 +50,116 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+		envOpt   = options.Environment()
+	)
+
+	if !has {
+		return
+	}
+
+	if envOpt.IsProduction() {
+		return errors.New(0, "when in production environment (ENVIRONMENT=production) you cannot provision "+
+			"super users, set the environment to dev (ENVIRONMENT=dev) to provision super users")
```
tjerman: `super users, set the environment` -> `super users; set the environment`

server/pkg/provision/auth.go
```diff
@@ -43,3 +50,116 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+		val, has = os.LookupEnv("AUTH_PROVISION_SUPER_USER")
+		envOpt   = options.Environment()
+	)
+
+	if !has {
+		return
+	}
+
+	if envOpt.IsProduction() {
+		return errors.New(0, "when in production environment (ENVIRONMENT=production) you cannot provision "+
```
tjerman: Use the appropriate constant instead of the `0`

server/pkg/provision/auth.go
```diff
@@ -43,3 +50,116 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+
+	usr := strings.Split(user, ",")
+	u.Email = usr[0]
+
+	if len(usr) > 1 {
+		u.Username = usr[1]
+	}
+
+	if len(usr) > 2 {
+		u.Username = usr[1]
```
tjerman: remove

server/pkg/provision/auth.go
```diff
@@ -43,3 +50,110 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+
+// addAuthSuperUsers assigns BYPASS roles to users from AUTH_PROVISION_SUPER_USER value
+// When in Production, Corteza should stop and report an error.
+func addAuthSuperUsers(ctx context.Context, log *zap.Logger, s store.Storer) (err error) {
+	var (
+		val    = options.Auth().ProvisionSuperUser
+		envOpt = options.Environment()
+	)
+
+	if envOpt.IsProduction() {
```
petergrlica, line 72: If I do not set the provisionSuperUser var in production, it still stops.

server/pkg/provision/provision.go
```diff
@@ -42,6 +42,7 @@ func Run(ctx context.Context, log *zap.Logger, s store.Storer, provisionOpt opti
 		func() error { return authAddExternals(ctx, log.Named("auth.externals"), s) },
 		func() error { return oidcAutoDiscovery(ctx, log.Named("auth.oidc-auto-discovery"), s, authOpt) },
 		func() error { return defaultAuthClient(ctx, log.Named("auth.clients"), s, authOpt) },
+		func() error { return addAuthSuperUsers(ctx, log.Named("auth.super-users"), s) },
```
petergrlica: can you use authOpt as the last argument and lose the `val = options.AuthOpt()...` in the body of the function?

server/.env.example
```diff
@@ -39,7 +39,7 @@
 
 ###############################################################################
 # Allow insecure (invalid, expired TLS/SSL certificates) connections.
-#
+#
```
tjerman, line 42: can you get rid of these spaces please

server/pkg/provision/auth.go
```diff
@@ -43,3 +51,121 @@ func emailSettings(ctx context.Context, s store.Storer) error {
...
+	return
+}
+
+// prepareUser creates and fills a new user depending on the number of arguments provided
+func prepareUser(user string) *types.User {
+	var (
+		nextID = func() uint64 {
+			return id.Next()
+		}
+	)
```
tjerman: Isn't there already a `nextID` func in this package?
If not, I think we should have it defined globally instead of here (there isn't any harm in doing that)

server/system/service/auth_credentials.go
```diff
@@ -532,20 +532,25 @@ func (svc *auth) CheckPasswordStrength(password string) bool {
 
 // SetPasswordCredentials (soft) deletes old password entry and creates a new entry with new password on every change
 //
-// This way we can implement more strict password-change policies in the future
+// # This way we can implement more strict password-change policies in the future
```
tjerman, line 535: what's up with this `#`?

server/system/service/auth_credentials.go
```diff
@@ -568,18 +573,17 @@ func (svc *auth) SetPasswordCredentials(ctx context.Context, userID uint64, pass
 
 // SetPassword sets new password for a user
 //
-// This function also records an action
+// # This function also records an action
```
tjerman, line 576: #?

tjerman (approved):
Address these few comments, then 👍

server/system/service/auth_credentials.go
KinyaElGrande: gofmt adds it automatically.

PR #636, opened 2022-12-13
Fix paste bug for all kinds of input

### LINKED ISSUE #629: When copy/pasting a text to content page block a warning is popping up
When you try to paste a text from the clipboard to the content page block configuration a warning message saying `Not a valid page-block` is popping up even though the content is pasted and when you save the text is there and everything looks fine. I guess just the warning message needs to be removed from popping up. 

https://user-images.githubusercontent.com/53232926/206737364-a2133b4f-561b-4c41-960c-9629f6cef803.webm

EmmyMay:
This addresses https://github.com/cortezaproject/corteza/issues/629

### DIFF
--- a/client/web/compose/src/views/Admin/Pages/Builder.vue
+++ b/client/web/compose/src/views/Admin/Pages/Builder.vue
@@ -1,7 +1,9 @@
 <template>
   <div
     v-if="!!page"
+    id="page-builder"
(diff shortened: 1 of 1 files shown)


Fajfa (changes requested):
Instead of expanding the condition for each case where we can paste on page, improve the condition so it only cares if you're pasting an actually JSON object.. And if an error happens then, then you can raise the warning

PR #642, opened 2022-12-13
Fix reporter text quote issue

### LINKED ISSUE #598: Reporter text block paragraph option not working as expected
When you add a text block in reporter and you want to use the paragraph setting, the paragraph option is not applied when the block is saved.

Image

EmmyMay:
Fixes https://github.com/cortezaproject/corteza/issues/598


Fajfa (changes requested):
Try to add the class to the text element itself (lib/vue)

PR #644, opened 2022-12-14
Improve text when copying a page

### LINKED ISSUE #509: Compose page duplication name improvement
When you create couple copies of a page they are named as:

Page
copy of page
copy of a copy of a page

Image

and so on respectively, so if we create 3-4 or more the naming kinda gets confusing and long, so let's try to improve the naming for example if we have test page, instead of saying copy of a test page, and after that copy of a copy of a test page, let's do this like:

Test
Copy of Test
Copy of Test (1)
Copy of Test (2) 

and so on.

yonkov:
# Screenshots
image
Image 1 - Copied pages list in the page admin

image
Image 2 - Save as Copy functionality

When copying a page in the page builder or the page admin, new naming structure has been implemented. To illustrate this, if you copy a page, called "Test", the new page will become "Copy of Test". If you copy "Copy of Test", it will become "Copy of Test (2)" and so on:  
Test
Copy of Test
Copy of Test (2)
Copy of Test (3)

### What was changed

When you go to admin => pages and edit a specific page, you have an option to clone existing page by clicking on "Save as Copy" button. (Another way to clone the page by going to the page builder of the specific page). When you click on "Save as Copy" button, a new page is generated and its title recomputed. If no other copies of the old page exist, the title of the new page is constructed by adding a prefix "copy of" to the page title. If other copies exist, a number is added as a sufix to the page's name. To give an example, if you copy a page, called "Test", the new page will become "Copy of Test". If you copy "Copy of Test", it will become "Copy of Test (2)" and so on:  

Test
Copy of Test
Copy of Test (2)
Copy of Test (3)

In previous versions of Corteza, the copied page was always prefixed with 'copy of'. If we continue with the same example as above, the page dublication process  used to look like that:

Test
Copy of Test
Copy of Copy of Test
Copy of Copy of Test
Copy of Copy of Copy of Test

### How was changed

In Compose, two files have been adjusted, `views/Admin/Pages/Edit.vue` and `views/Admin/Pages/Builder.vue`. The namespace's pages have been programatically filtered to find how many times a certain page with same name already exists by excluding 'copy of' prefix and number sufix. Depending on the number of pages that match the same name, a counter is adjusted. 

### Why was changed

If we copy a page 3-4 or more times, the naming gets confusing and too long. Reshaping this aims to improve user interaction across Corteza.

### DIFF
--- a/client/web/compose/src/mixins/pages.js
+++ b/client/web/compose/src/mixins/pages.js
@@ -1,0 +1,66 @@
+import { NoID } from '@cortezaproject/corteza-js'
+
+export default {
+  methods: {
+    countDuplicateTitles (pages, pattern) {
+      let substrTitle = this.page.title
+
+      // if the old page begins with 'copy of' and it does not end with a number in brackets, remove copy of prefix
+      if (this.page.title.includes(this.$t('copyOf')) && !pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.length)
+      // if the old page begins with 'copy of' and it ends with a number in brackets, remove them
+      } else if (this.page.title.includes(this.$t('copyOf')) && pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.lastIndexOf('(') - 1)
+      }
+
+      return pages.filter(p => p.title.includes(substrTitle)).length
+    },
+
+    handleClone () {
+      const { namespaceID = NoID } = this.namespace
+      const pattern = /\([0-9]+\)/
+
+      let page = this.page.clone()
+
+      this.loadPages({ namespaceID, force: true })
+        .then(pages => {
+          let newPageTitle = this.$t('copyOf', { title: this.page.title })
+          const countDuplicates = this.countDuplicateTitles(pages, pattern)
+
+          // if page with the same name already exists two times, add a number (2) sufix
+          if (countDuplicates === 2) {
+            newPageTitle = this.page.title.includes(this.$t('copyOf'))
+              ? `${this.page.title} (${countDuplicates})`
+              : `${this.$t('copyOf')} ${this.page.title} (${countDuplicates})`
+          } else if (countDuplicates > 2 && this.page.title.includes(this.$t('copyOf'))) {
+            /**
+            * if page with the same name exists more than two times and it has "copy of" prefix, check old page's sufix
+            * If page ends with a number inside parenthesis, replace it with the new number
+            * If page does not have a number, add it
+            */
+            newPageTitle = pattern.test(page.title)
+              ? `${this.page.title.substr(0, this.page.title.lastIndexOf('('))}(${countDuplicates})`
+              : `${this.page.title} (${countDuplicates})`
+          } else if (countDuplicates > 2 && !this.page.title.includes(this.$t('copyOf'))) {
+            // if page with the same name exists more than two times and it does not have "copy of" prefix, append the prefix and the new number
+            newPageTitle = `${this.$t('copyOf')}${this.page.title} (${countDuplicates})`
+          }
+
+          page = {
+            ...page,
+            pageID: NoID,
+            title: newPageTitle,
+            handle: '',
+          }
+
+          return this.createPage({ namespaceID, ...page })
+        })
+        .then(({ pageID }) => {
+          this.$router.push({ name: this.$route.name, params: { pageID } })
+        })
+        .catch(this.toastErrorHandler(this.$t('notification:page.cloneFailed')))
+    },
+
+  },
+
+}
--- a/client/web/compose/src/views/Admin/Pages/Builder.vue
+++ b/client/web/compose/src/views/Admin/Pages/Builder.vue
@@ -223,6 +223,7 @@
 
 <script>
 import { mapGetters, mapActions } from 'vuex'
+import pages from 'corteza-webapp-compose/src/mixins/pages'
 import NewBlockSelector from 'corteza-webapp-compose/src/components/Admin/Page/Builder/Selector'
 import PageTranslator from 'corteza-webapp-compose/src/components/Admin/Page/PageTranslator'
 import Grid from 'corteza-webapp-compose/src/components/Common/Grid'
@@ -245,6 +246,10 @@ export default {
     PageTranslator,
   },
 
+  mixins: [
+    pages,
+  ],
+
   props: {
     namespace: {
       type: compose.Namespace,
@@ -261,7 +266,6 @@ export default {
     return {
       editor: undefined,
       page: undefined,
-
       blocks: [],
       board: null,
     }
@@ -360,7 +364,7 @@ export default {
 
         if (pageID) {
           const { namespaceID, name } = this.namespace
-          this.findPageByID({ namespaceID, pageID: this.pageID, force: true })
+          this.findPageByID({ namespaceID, pageID, force: true })
             .then(page => {
               document.title = [page.title, name, this.$t('general:label.app-name.private')].filter(v => v).join(' | ')
 
@@ -386,6 +390,7 @@ export default {
       deletePage: 'page/delete',
       updatePageSet: 'page/updateSet',
       createPage: 'page/create',
+      loadPages: 'page/load',
     }),
 
     addBlock (block, index = undefined) {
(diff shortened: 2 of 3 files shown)


client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -384,8 +383,34 @@ export default {
       deletePage: 'page/delete',
       updatePageSet: 'page/updateSet',
       createPage: 'page/create',
+      loadPages: 'page/load',
     }),
 
+    // helper to count pages with the same name
+    countDuplicatePages (pageTitles) {
```
Fajfa: Are we counting duplicate pages or duplicate page titles? Also the param is called pageTitles, no need to map it before passing it in, just pass the pages. Since we were talking about faster/slower, well you do a for loop here (map) for no reason so that is another O(n) slower

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -384,8 +383,34 @@ export default {
...
+      let substrTitle
+
+      // find a number inside parentheses
+      const pattern = /\([0-9]+\)/
+
+      if (this.page.title.includes(this.$t('copyOf')) && !pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.length)
+      } else if (this.page.title.includes(this.$t('copyOf')) && pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.lastIndexOf('(') - 1)
+      } else {
```
Fajfa: Do we need an else, why not set a default instead

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -570,20 +595,38 @@ export default {
...
-        pageID: NoID,
-        title: this.$t('copyOf', { title: this.page.title }),
-        handle: '',
-      }
 
-      const { namespaceID = NoID } = this.namespace
-      this.createPage({ namespaceID, ...page })
-        .then(({ pageID }) => {
-          this.$router.push({ name: 'admin.pages.builder', params: { pageID } })
+      this.loadPages({ namespaceID, force: true })
```
Fajfa: So if loadPages fails there will be no error, please handle that since you have nested promises here

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -570,20 +595,38 @@ export default {
...
+        .then(pages => {
+          const pageTitles = pages.map(p => new compose.Page(p).title)
```
Fajfa: Do we really need to map the pages to the class here? Look at the store, they already are

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -570,20 +595,38 @@ export default {
...
+          page = {
+            ...page,
+            pageID: NoID,
+            title: newPageTitle,
+            handle: '',
+          }
+
+          this.createPage({ namespaceID, ...page })
+            .then(page => {
+              this.page = new compose.Page(page)
```
Fajfa: I dont think we need to set this.page, since you get redirected so its just confusing

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Edit.vue
```diff
@@ -313,19 +293,22 @@ export default {
...
 
       this.loadPages({ namespaceID, force: true })
         .then(pages => {
-          const pageTitles = pages.map(p => new compose.Page(p).title)
           let newPageTitle = this.$t('copyOf', { title: this.page.title })
-          const countDuplicates = this.countDuplicatePages(pageTitles)
+          const countDuplicates = this.countDuplicatePages(pages, pattern)
 
           if (countDuplicates === 2) {
             newPageTitle = this.page.title + ' (' + countDuplicates + ')'
```
Fajfa: String interpolation instead of the old approach, everywhere you use it

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
...
+      if (this.page.title.includes(this.$t('copyOf')) && !pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.length)
+      } else if (this.page.title.includes(this.$t('copyOf')) && pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.lastIndexOf('(') - 1)
+      }
+
+      return pages.filter(p => p.title.includes(substrTitle)).length
+    },
+
+    async handleClone (name) {
```
Fajfa: No need to pass the name in you can get it from route

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
+import { compose, NoID } from '@cortezaproject/corteza-js'
+
+export default {
+  methods: {
+    countDuplicatePages (pages, pattern) {
+      let substrTitle = this.page.title
+
+      if (this.page.title.includes(this.$t('copyOf')) && !pattern.test(this.page.title)) {
```
Fajfa, line 9: Might be beneficial to add some comments for the ifs of what they check for

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
+import { compose, NoID } from '@cortezaproject/corteza-js'
+
+export default {
+  methods: {
+    countDuplicatePages (pages, pattern) {
```
Fajfa: Name is still countDuplicatePages but thats not what this method does

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
...
+        const pattern = /\([0-9]+\)/
+
+        const pages = await this.loadPages({ namespaceID, force: true })
+
+        let newPageTitle = this.$t('copyOf', { title: this.page.title })
+        const countDuplicates = this.countDuplicatePages(pages, pattern)
+
+        if (countDuplicates === 2) {
+          newPageTitle = `${this.page.title} (${countDuplicates})`
+        } else if (countDuplicates > 2 && this.page.title.includes(this.$t('copyOf'))) {
```
Fajfa: Add comments here aswell and maybe think about rewriting the condition so that `countDuplicates > 2` doesn't repeat

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
...
+          pageID: NoID,
+          title: newPageTitle,
+          handle: '',
+        }
+
+        const newPage = await this.createPage({ namespaceID, ...page })
+
+        const { pageID = '' } = newPage
+        this.$router.push({ name, params: { pageID } })
+        this.page = new compose.Page(page)
```
Fajfa: This shouldn't be necessary, check why builder works and edit doesn't

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -179,7 +179,7 @@
         @save="handleSave()"
         @delete="handleDeletePage"
         @saveAndClose="handleSave({ closeOnSuccess: true })"
-        @clone="handleClone()"
+        @clone="handleClone('admin.pages.builder')"
```
Fajfa: handleClone shouldnt need any params

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,56 @@
...
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.length)
+      } else if (this.page.title.includes(this.$t('copyOf')) && pattern.test(this.page.title)) {
+        substrTitle = this.page.title.substring(this.$t('copyOf').length, this.page.title.lastIndexOf('(') - 1)
+      }
+
+      return pages.filter(p => p.title.includes(substrTitle)).length
+    },
+
+    async handleClone (name) {
+      try {
```
Fajfa: We don't use try catch for promises in es6

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Edit.vue
```diff
@@ -244,6 +249,22 @@ export default {
...
       return this.hasChildren && this.page.canDeletePage && !this.page.deletedAt
     },
+
+  },
+
+  watch: {
+    pageID: {
+      immediate: true,
+      handler (pageID) {
+        if (pageID) {
```
Fajfa, line 259: Make sure you do it how it was done in created, since currently you're double fetching it. 
And the catch is missing, so yes. Look at created

client/web/compose/src/mixins/pages.js
```diff
@@ -0,0 +1,65 @@
...
+            ...page,
+            pageID: NoID,
+            title: newPageTitle,
+            handle: '',
+          }
+
+          return this.createPage({ namespaceID, ...page })
+        })
+        .then(page => {
+          const { pageID = '' } = page
```
Fajfa: Also lets do this deconstruct in the above line, and the fallback shouldn't be necessary

PR #668, opened 2022-12-15
fix authorized clients empty-list text view

### LINKED ISSUE #580: Auth pages (authorized clients) do not show empty-list text properly
Empty list:

Screenshot 2022-11-29 at 14 42 29

List of auth clients

Screenshot 2022-11-29 at 14 42 10

See the "no authorized clients found". That text should show on the empty list but not on the full one. It works in reverse right now.

Make sure it is fixed on `2022.9.x` and on `2023.3.x`.

KinyaElGrande:
Fix : show  **_no authorized clients found_**  text on an empty authorized clients list instead of a full one.


tjerman (approved):
Add an empty line at the end of the file you modified then lgtm

PR #709, opened 2022-12-21
Fix filtering not available in block magnification

### LINKED ISSUE #635: Filtering not available in block magnification
When you configure a block (filter list) to be magnified the filter is present but it is not working, it is not clickable. It looks like the problem is with the z-index and the popoverTarget being the same as the record list not in modal.

Steps to reproduce:
1. Add a block and configure it the magnification to be enabled
2. Click on magnify icon 
3. Once the list opens in modal for example try clicking on the filter

kelanik8:
When you configure a block (filter list) to be magnified the filter is present but it is not working, it is not clickable. It looks like the problem is with the z-index and the popoverTarget being the same as the record list not in modal.

 - [x] Code builds
 - [x] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -214,7 +214,7 @@
                     <record-list-filter
                       v-if="field.filterable"
                       class="d-print-none"
-                      :target="block.blockID"
+                      :target="uniqueID"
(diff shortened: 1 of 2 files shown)


kelanik8:
image

Fajfa (changes requested):
Filter is not displaying because of the ID and  z-index. So this is not the correct approach

Fajfa (approved):

PR #710, opened 2022-12-21
Remove whitespace from module field select fields

### LINKED ISSUE #643: Whitespaces trimmed on key -> val dropdowns
## Description
Currently, the key values setting for dropdown lists allows for whitespaces. This can be difficult to detect and debug when it occurs because it: 
- prevents record manipulation
- the exact field is not logged in the error log

## Screenshot of issue
Image

## TODO:
- [ ] remove the whitespace after the values

EmmyMay:


EmmyMay:
Fixes https://github.com/cortezaproject/corteza/issues/643

client/web/compose/src/components/ModuleFields/Configurator/Select.vue
```diff
@@ -10,14 +10,14 @@
           class="mb-1"
         >
           <b-form-input
-            v-model="f.options.options[index].value"
+            v-model.trim="f.options.options[index].value"
```
Fajfa, line 13: Only value should be trimmed here, text can have spaces since we just display that

Fajfa (approved):

PR #712, opened 2022-12-21
Add spinner when loading actionlog in admin

### LINKED ISSUE #650: Wrong state shown while loading action log list
## Current behaviour
While we're fetching the action log list, the text `Not found` is displayed

Image

## Expected behaviour
the loading spinner should be shown

EmmyMay:
Fixes https://github.com/cortezaproject/corteza/issues/650

### DIFF
--- a/client/web/admin/src/views/System/Actionlog/Index.vue
+++ b/client/web/admin/src/views/System/Actionlog/Index.vue
@@ -98,6 +98,7 @@
         hover
         class="mb-0 small"
         head-variant="light"
+        :busy="processing"
         :items="items"
         :fields="fields"
         tbody-tr-class="pointer"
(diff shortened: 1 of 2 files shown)


client/web/admin/src/views/System/Actionlog/Index.vue
```diff
@@ -98,6 +98,7 @@
         hover
         class="mb-0 small"
         head-variant="light"
+        :busy="isBusy"
```
Fajfa: Just to keep consistent, can we call isBusy processing instead.
Also, does the spinner look like other resource lists? If not lets try to make it as similar as possible

Fajfa (changes requested):

Fajfa (approved):

PR #714, opened 2022-12-21
Fix reporter page block load elements error

### LINKED ISSUE #597: Having more than 2 elements in a reporter page block is giving an error
When I configure a reporter page block to have two elements, for example metrics and table, when I save the second element I am receiving the error `Failed to run report: expecting at most one definition, got 2`. Even though this error is displayed it looks like everything is functioning well.

https://user-images.githubusercontent.com/53232926/205629572-d9b5b8bf-2c00-4851-b883-48bb4351c23a.webm

KinyaElGrande:

More than one load element can be added to the reporter page block without   getting 
 `expecting at most one definition, got [number of elements]`  error message.

Fixes https://github.com/cortezaproject/corteza/issues/597

### DIFF
--- a/client/web/reporter/src/components/Report/Blocks/index.vue
+++ b/client/web/reporter/src/components/Report/Blocks/index.vue
@@ -188,10 +188,14 @@ export default {
       this.dataframes = {}
       const frames = []
 
-      this.block.elements.forEach((element) => {
+      this.block.elements.forEach((element, key) => {
         element = reporter.DisplayElementMaker(element)
 
         if (element && element.kind !== 'Text') {
+          if (element.elementID === '0') {
+            element.elementID = `${key}`
+          }
+
           const { dataframes = [] } = element.reportDefinitions(this.getScenarioDefinition(element))
 
           frames.push(...dataframes.filter(({ source }) => source))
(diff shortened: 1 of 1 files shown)


client/web/reporter/src/components/Report/Blocks/index.vue
```diff
@@ -188,10 +188,14 @@ export default {
       this.dataframes = {}
       const frames = []
 
-      this.block.elements.forEach((element) => {
+      this.block.elements.forEach((element, key) => {
         element = reporter.DisplayElementMaker(element)
 
         if (element && element.kind !== 'Text') {
+          if (element.elementID === '0') {
```
Fajfa, line 195: If it works its probably by accident, since how does the element.elementID match any names if that's not whats in the block element names

client/web/reporter/src/components/Report/Blocks/index.vue
```diff
@@ -188,10 +188,14 @@ export default {
...
+            element.elementID = key.toString()
```
Fajfa: Can be simplified by just doing `${key}`

client/web/reporter/src/components/Report/Blocks/index.vue
KinyaElGrande: `element.reportDefinitions` will assign the element.elementID to element's name https://github.com/cortezaproject/corteza/blob/79b414ecdbe85c9080b6ae12a5b4d220aeaba23a/client/web/reporter/src/components/Report/Blocks/index.vue#L199

PR #730, opened 2022-12-30
Align text message in inc_toasts.html

### LINKED ISSUE #676: Client authorization removed message notification alignment
When you revoke the access of the authorization clients the message/notification that is displayed below is not aligned properly as you can see from the screenshot. The text is not on the same line as the `X`.

Image

katrinDY:
The text of the toast message was centered according to the y axis

Screenshot 2022-12-30 at 12 56 28


Fajfa:
LGTM

Fajfa (approved):

PR #734, opened 2023-01-02
Preview page attachments in public pages

### LINKED ISSUE #718: Can't zoom in uploaded file in record
**Low priority**

## Steps to reproduce
1. Create a module with a field of type file
2. Click on `All records`
3. Add a new record
4. Upload a file
5. Save the record
6. Hover over the uploaded file and see the magnifying glass appear
7. Click on the file and see that nothing happens
7.1. Also can we add margin/padding on the right side of the file so it looks better?

## Video of issue
Image

## Why doesn't this work
The preview modal is only initialized on public pages so it will only work there

## TODO
- [x] move the attachment modal from public/pages/view, to namespace/view so the view can be used in both places

katrinDY:
Moved attachment modal from Public/Pages/View to Namespace/View so attachments can be previewed in the admin area of compose

### DIFF
--- a/client/web/compose/src/views/Namespace/View.vue
+++ b/client/web/compose/src/views/Namespace/View.vue
@@ -45,11 +45,14 @@
         </div>
       </div>
     </div>
+
+    <attachment-modal />
   </div>
 </template>
 
 <script>
 import { mapGetters } from 'vuex'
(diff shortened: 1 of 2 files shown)


Fajfa (approved):
LGTM

PR #741, opened 2023-01-04
Add a nicer error when trying to delete in use sens. levels

### LINKED ISSUE #651: Error when trying to delete a sensitivity level added to a connection
## Error
Image

## Steps to reproduce
1. Create a sensitivity level
2. Create a connection and add the newly created sensitivity level
3. Try to delete the sensitivity level used in the newly created connection and see the error pop up

tjerman:
This code adds some analysis to find out if a sensitivity level is being in use before trying to delete it.
Before, the user would get a _scary_ looking error but now it's clearer.

### DIFF
--- a/locale/en/corteza-server/system/dal-sensitivity-level.yaml
+++ b/locale/en/corteza-server/system/dal-sensitivity-level.yaml
@@ -6,3 +6,4 @@ errors:
   invalidID: invalid ID
   notAllowedToManage: not allowed to manage sensitivity level
   notFound: sensitivity level not found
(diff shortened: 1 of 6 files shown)


tjerman:
## Changed

### What

error message when trying to delete a sensitivity level that's in use to be more informative on what is wrong.

### How

adding additional validation before the sensitivity level is deleted

### Refs

Monorepo

- 2022.9.x `56e55439d`
- 2023.3.x omit; not relevant since it's a new version

PR #758, opened 2023-01-10
Add Record Revisions Text Tessage When Revisions are Enabled

### LINKED ISSUE #747: Record revision page block doesn't show any data when in page builder mode
## Steps to reproduce
1. Pick a module (doesn't matter which)
2. Edit the module
3. Select the record revision tab
4. Enable record revisioning
5. Go to the record page of the same module
6. Edit a record
7. Add a record revision page block
8. Notice how in page builder mode no data is displayed
9. Switch to view mode and see the data displayed

Image

## Expected behavior
Either the same data should be displayed as in view mode or some kind of text should be shown. Discuss with Jože

yonkov:
image

### DIFF
--- a/client/web/compose/src/components/PageBlocks/RecordRevisionsBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordRevisionsBase.vue
@@ -14,10 +14,12 @@
       >
         {{ $t('errors.disabled-on-module') }}
       </span>
+
       <b-spinner
         v-else-if="noRecord || processing"
         class="my-auto"
       />
+
       <b-button
         v-else-if="!preloadRevisions && !loadedRevisions"
         class="my-auto"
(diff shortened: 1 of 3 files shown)


locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -419,7 +419,8 @@ recordRevisions:
   viewer:
     errors:
       no-record: No record found to display revisions for
-      disabled-on-module: Record revisions are disabled on this module
+      disabled-on-module: Record revisions are disabled on this module.
+      enabled-on-module: Record revisions are enabled on this module. To see the record revisions for a specific record, go to the public page for that record.
```
Fajfa: No need for such long text, just say why the block cannot be displayed. Which is because the is no record

client/web/compose/src/components/PageBlocks/RecordRevisionsBase.vue
```diff
@@ -181,6 +187,14 @@ export default {
     noRecord () {
       return !this.record
     },
+
+    isPageBuilder () {
```
Fajfa: Why introduce the two new computeds if you can just check if `this.record.recordID === NoID`

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordRevisionsBase.vue
```diff
@@ -26,65 +28,86 @@
...
+          </template>
+        </b-table-lite>
+
+        <div
+          v-if="!revisions.length"
+          class="position-absolute text-center mt-5 d-print-none"
+          style="left: 0; right: 0;"
+        >
+          <b-spinner
+            v-if="noRecord || processing"
```
Fajfa: This is a duplicate of the spinner on line 18, I would prefer if you use that one and just remove this one to avoid edge cases

client/web/compose/src/components/PageBlocks/RecordRevisionsBase.vue
```diff
@@ -26,65 +28,86 @@
...
+            class="my-auto"
+          />
+
+          <p
+            v-else
+            class="pt-3"
```
Fajfa: mt-3 instead please, more of a margin job

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -419,7 +419,7 @@ recordRevisions:
   viewer:
     errors:
       no-record: No record found to display revisions for
-      disabled-on-module: Record revisions are disabled on this module
+      disabled-on-module: Record revisions are disabled on this module.
```
Fajfa: Remove the . here, no need

Fajfa (changes requested):

PR #770, opened 2023-01-11
Support Conditional fields on record pageblocks

### LINKED ISSUE #478: Support for conditional fields [PlanID:2023.3/48/2.6]
This is a basic implementation of conditional fields for the **record page block**. It provides a way for users to control fields that should be hidden or shown by checking the current context and values of the current record.

Examples:
 - when the selected age is 18 or more, show the field for gender
 - when the occupation is developer, show text input for a favourite IDE

The condition expression is written in JavaScript and executed inside a browser using **eval**. There might be other expression languages and execution environments provided in the future. 

Configuration for conditional fields is stored under record block options and not on the field!

- [x] extend `lib/js/src/compose/types/page-block/record.ts` to support field conditions
- [x] make sure implementation is future-proof (eg, allow extension to other languages for conditions)
- [x] configurator must warn/prevent users when adding a conditional view to required fields

EmmyMay:
Addresses https://github.com/cortezaproject/corteza/issues/478

### DIFF
--- a/client/web/compose/src/components/PageBlocks/RecordBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordBase.vue
@@ -9,6 +9,7 @@
     >
       <b-spinner />
     </div>
+
     <div
       v-else-if="module"
       class="mt-3"
@@ -16,39 +17,43 @@
       <div
         v-for="(field, index) in fields"
         :key="index"
-        class="d-flex flex-column mb-3 px-3"
+        :class="{ 'd-flex flex-column mb-3 px-3': canDisplay(field)}"
       >
-        <label
-          class="text-primary mb-0"
-          :class="{ 'mb-0': !!(field.options.description || {}).view || false }"
+        <template
+          v-if="canDisplay(field)"
         >
-          {{ field.label || field.name }}
-          <hint
-            :id="field.fieldID"
-            :text="((field.options.hint || {}).view || '')"
-            class="d-inline-block"
-          />
-        </label>
+          <label
+            class="text-primary mb-0"
+            :class="{ 'mb-0': !!(field.options.description || {}).view || false }"
+          >
+            {{ field.label || field.name }}
+            <hint
+              :id="field.fieldID"
+              :text="((field.options.hint || {}).view || '')"
+              class="d-inline-block"
+            />
+          </label>
 
-        <small
-          class="text-muted"
-        >
-          {{ (field.options.description || {}).view }}
-        </small>
-        <div
-          v-if="field.canReadRecordValue"
-          class="value mt-2"
-        >
-          <field-viewer
-            v-bind="{ ...$props, field }"
-          />
-        </div>
-        <i
-          v-else
-          class="text-primary"
-        >
-          {{ $t('field.noPermission') }}
-        </i>
+          <small
+            class="text-muted"
+          >
+            {{ (field.options.description || {}).view }}
+          </small>
+          <div
+            v-if="field.canReadRecordValue"
+            class="value mt-2"
+          >
+            <field-viewer
+              v-bind="{ ...$props, field }"
+            />
+          </div>
+          <i
+            v-else
+            class="text-primary"
+          >
+            {{ $t('field.noPermission') }}
+          </i>
+        </template>
       </div>
     </div>
   </wrap>
@@ -59,6 +64,7 @@ import base from './base'
 import FieldViewer from 'corteza-webapp-compose/src/components/ModuleFields/Viewer'
 import Hint from 'corteza-webapp-compose/src/components/Common/Hint.vue'
 import users from 'corteza-webapp-compose/src/mixins/users'
+import conditionalFields from 'corteza-webapp-compose/src/mixins/conditionalFields'
 
 export default {
   i18nOptions: {
@@ -74,6 +80,7 @@ export default {
 
   mixins: [
     users,
+    conditionalFields,
   ],
 
   computed: {
@@ -96,7 +103,7 @@ export default {
     },
 
     processing () {
-      return !this.record
+      return !this.record || this.evaluating
     },
   },
 
@@ -104,7 +111,16 @@ export default {
     'record.recordID': {
       immediate: true,
       handler (recordID) {
-        if (recordID && recordID !== NoID) {
+        if (!recordID) return
+
+        this.evaluating = true
+
+        this.evaluateExpressions()
+          .finally(() => {
+            this.evaluating = false
+          })
+
+        if (recordID !== NoID) {
           this.fetchUsers(this.fields, [this.record])
         }
       },
--- a/client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
@@ -8,6 +8,7 @@
       <label>
         {{ $t('general.module') }}
       </label>
+
       <input
         v-if="module"
         v-model="module.name"
@@ -27,11 +28,120 @@
         style="max-height: 52vh;"
       />
     </b-form-group>
+
+    <hr>
+
+    <div>
+      <b-form-group
+        label-size="lg"
+      >
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            {{ $t('record.fieldConditions.label') }}
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1 text-decoration-none border-0"
+              :disabled="addRuleDisabled"
+              @click="addRule"
+            >
+              {{ $t('record.fieldConditions.action') }}
+            </b-button>
+
+            <b-button
+              variant="link"
+              :href="`${documentationURL}#value-sanitizers`"
+              target="_blank"
+              class="p-0 ml-auto"
+            >
+              {{ $t('record.fieldConditions.help') }}
+            </b-button>
+          </div>
+        </template>
+      </b-form-group>
+
+      <b-table-simple
+        v-if="block.options.fieldConditions.length"
+        borderless
+        small
+        responsive="lg"
+      >
+        <b-thead>
+          <b-tr>
+            <b-th
+              class="text-primary"
+            >
+              {{ $t('record.fieldConditions.field') }}
+            </b-th>
+
+            <b-th
+              class="text-primary"
+            >
+              {{ $t('record.fieldConditions.condition') }}
+            </b-th>
+
+            <b-th />
+          </b-tr>
+        </b-thead>
+
+        <b-tbody>
+          <b-tr
+            v-for="(condition, i) in block.options.fieldConditions"
+            :key="i"
+          >
+            <b-td
+              class="align-middle"
+              style="width: 33%; min-width: 250px;"
+            >
+              <vue-select
+                v-model="condition.field"
+                :options="block.options.fields"
+                append-to-body
+                :placeholder="$t('record.fieldConditions.selectPlaceholder')"
+                :selectable="option => isSelectable(option)"
+                :get-option-label="getOptionLabel"
+                :reduce="option => option.isSystem ? option.name : option.fieldID"
+                class="field-selector bg-white"
+              />
+            </b-td>
+
+            <b-td
+              class="align-middle"
+              style="min-width: 300px"
+            >
+              <b-input-group>
+                <b-input-group-prepend>
+                  <b-button variant="dark">
+                    ƒ
+                  </b-button>
+                </b-input-group-prepend>
+                <b-form-input
+                  v-model="condition.condition"
+                  :placeholder="$t('record.fieldConditions.placeholder')"
+                />
+              </b-input-group>
+            </b-td>
+
+            <b-td
+              class="text-center align-middle pr-2"
+              style="width: 100px;"
+            >
+              <c-input-confirm
+                @confirmed="deleteRule(i)"
+              />
+            </b-td>
+          </b-tr>
+        </b-tbody>
+      </b-table-simple>
+    </div>
   </b-tab>
 </template>
 <script>
 import base from './base'
 import FieldPicker from 'corteza-webapp-compose/src/components/Common/FieldPicker'
+import { VueSelect } from 'vue-select'
 
 export default {
   i18nOptions: {
@@ -42,31 +152,63 @@ export default {
 
   components: {
     FieldPicker,
+    VueSelect,
   },
 
   extends: base,
+
+  computed: {
+    documentationURL () {
+      // eslint-disable-next-line no-undef
+      const [year, month] = VERSION.split('.')
+      return `https://docs.cortezaproject.org/corteza-docs/${year}.${month}/integrator-guide/compose-configuration/index.html`
+    },
+
+    addRuleDisabled () {
+      return this.block.options.fields.filter(f => !f.isRequired).length === this.block.options.fieldConditions.length
+    },
+  },
+
+  methods: {
+    addRule () {
+      this.options.fieldConditions.push({
+        field: undefined,
+        condition: '',
+      })
+    },
+
+    deleteRule (i) {
+      this.options.fieldConditions.splice(i, 1)
+    },
+
+    isSelectable (option) {
+      return !this.block.options.fieldConditions.find(({ field }) => field === option.fieldID || field === option.name) && !option.isRequired
+    },
+
+    getOptionLabel (option) {
+      return option.label || option.name
+    },
+  },
 }
 </script>
-<style lang="scss" scoped>
-  div.fields {
-    display: flex;
-    flex-flow: row nowrap;
 
-    & > div {
-      flex: 1;
-      margin: 5px;
-
-      button.all {
-        float: right;
-        font-size: 80%;
-      }
-
-      .drag-area {
-        height: 150px;
-        overflow-x: auto;
-        border: 1px solid silver;
-        padding: 2px;
-      }
-    }
+<style lang="scss">
+.field-selector {
+  .vs__selected-options {
+    flex-wrap: nowrap;
   }
+
+  .vs__selected {
+    max-width: 200px;
+    display: inline-block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+}
+
+.vs__dropdown-menu .vs__dropdown-option {
+  text-overflow: ellipsis;
+  overflow: hidden !important;
+}
 </style>
--- a/client/web/compose/src/components/PageBlocks/RecordEditor.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordEditor.vue
@@ -9,6 +9,7 @@
     >
       <b-spinner />
     </div>
+
     <div
       v-else-if="module"
       class="px-3"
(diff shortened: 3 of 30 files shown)


client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,11 +117,55 @@ export default {
...
+          label: f.label,
+          fieldID: f.fieldID,
+          isRequired: f.isRequired,
+        }
+      })
+      return fields
+    },
+  },
+
+  created () {
```
Fajfa: No need right?

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -411,6 +419,23 @@ export default {
       this.page.blocks = this.blocks
     },
 
+    checkState ({ items, indicator }) {
```
Fajfa: What are we doing here? The conditional fields are only used in record pages, and not while making them (builder)
Also not really sure what the point of this function is

lib/js/src/compose/types/page-block/record.ts
```diff
@@ -2,12 +2,19 @@ import { PageBlock, PageBlockInput, Registry } from './base'
 
 const kind = 'Record'
 
+interface Conditions {
```
Fajfa: This would be a FieldCondition

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            Condition Fields
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1 mr-auto"
+              @click="add"
```
Fajfa: add what (better naming)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
         style="max-height: 52vh;"
       />
     </b-form-group>
+
+    <div>
+      <b-form-group
+        :label="$t('sanitizers.label')"
```
Fajfa: probably not the correct key for the label (guessing its copy pasted)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
         style="max-height: 52vh;"
       />
     </b-form-group>
+
+    <div>
```
Fajfa, line 34: As mentioned style this to have a bigger space, consider adding an "\<hr\>" element between them

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+      >
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            Condition Fields
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1 mr-auto"
```
Fajfa: I would prefer if you did ml auto on the examples section

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+      >
+        <b-tbody>
+          <b-tr
+            v-for="(condition, i) in block.options.fieldConditions"
+            :key="i"
+          >
+            <b-td>
+              <vue-select
+                v-model="condition.field.name"
+                :options="fieldOptions"
```
Fajfa: you should look at how a field selector vue-select is configured (there is a couple of props that you need/should use here)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+                class="bg-white"
+                @close="(() => $root.$emit('checkState', { items: options.fieldConditions, indicator: 'field' }))"
```
Fajfa: You dont need this

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+                </b-input-group-prepend>
+                <b-form-input
+                  v-model="condition.condition"
+                />
+              </b-input-group>
+            </b-td>
+
+            <b-td>
+              <c-input-confirm
+                class="mt-2 ml-2"
```
Fajfa: no margin here, you probably want to align it, use the proper td/tr styles for that. Look at the module edit

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,11 +117,55 @@ export default {
...
+
+  computed: {
+    documentationURL () {
+      // eslint-disable-next-line no-undef
+      const [year, month] = VERSION.split('.')
+      return `https://docs.cortezaproject.org/corteza-docs/${year}.${month}/integrator-guide/compose-configuration/index.html`
+    },
+
+    fieldOptions () {
+      const nonRequiredFields = this.module.fields.filter(f => !f.isRequired)
```
Fajfa: We dont need to filter two times to do this. Also you should not allow all module fields, just use the ones that are selected in the above field picker!

The key is disabling the already selected ones and the required fields, there is a prop for that for vue-select (selectable prop)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,11 +117,55 @@ export default {
...
+    },
+  },
+
+  created () {
+    console.log(this.fieldOptions)
+  },
+
+  methods: {
+    add () {
+      if (this.options.fieldConditions.length >= this.module.fields.length) return
```
Fajfa: No need for this check, you won't be able to select a field anyhow, since the selectable prop will disable that

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+              @click="add"
+            >
+              + Add
+            </b-button>
+
+            <b-button
+              variant="link"
+              :href="`${documentationURL}#value-sanitizers`"
+              target="_blank"
+              class="p-0 ml-1"
```
Fajfa: I dont think we need ml-1 right?

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,86 @@
...
+    <div>
+      <b-form-group
+        :label="$t('sanitizers.label')"
+        label-size="lg"
+      >
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            Condition Fields
```
Fajfa: Consider making this look like a b-form-group label

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,11 +117,55 @@ export default {
...
+
+  created () {
+    console.log(this.fieldOptions)
+  },
+
+  methods: {
+    add () {
+      if (this.options.fieldConditions.length >= this.module.fields.length) return
+      this.options.fieldConditions.push({
+        field: {},
```
Fajfa: {} is a valid value! You shouldnt be pushing a valid value

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,11 +117,55 @@ export default {
...
+
+    deleteCondition (i) {
+      this.options.fieldConditions.splice(i, 1)
+      this.$root.$emit('checkState', { items: this.options.fieldConditions, indicator: 'field' })
+    },
+  },
 }
 </script>
+
 <style lang="scss" scoped>
```
Fajfa: Delete this whole style section, not sure why its still here (I know you didn't add it but its not needed)

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Builder.vue
EmmyMay: It caters to an edge case where a person tries to save a recordconfigurator with a non-configured condition. This method will also replace the one I am using in pageblock tabs to disable the save and close button

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
EmmyMay: image

This has kind of been the convention we have been using for these kind of situations...

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
EmmyMay: 😅 Pilot generated. Will be amended.

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
EmmyMay: It caters to an edge case where a person tries to save a recordconfigurator with a non-configured condition. This method will also replace the one I am using in pageblock tabs to disable the save and close button

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
Fajfa: Bad convention, it should be changed there aswell, since in the context of recordConfigurator add doesn't mean anything (are you adding a record?)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
EmmyMay: I think the user should not be making field conditions even after it is clear that they have run out of fields.

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,88 @@
...
+        <b-tbody>
+          <b-tr
+            v-for="(condition, i) in block.options.fieldConditions"
+            :key="i"
+          >
+            <b-td>
+              <vue-select
+                v-model="condition.field"
+                :options="fieldOptions"
+                :clearable="false"
```
Fajfa: Why not clearable? Also searchable?

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,88 @@
...
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            {{ $t('record.fieldConditions.label') }}
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1 mr-auto"
+              @click="addCondition"
```
Fajfa: Are we adding a condition here? or should it be called something else

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,88 @@
...
+                  v-model="condition.condition"
+                  :placeholder="$t('record.fieldConditions.placeholder')"
+                />
+              </b-input-group>
+            </b-td>
+
+            <b-td>
+              <c-input-confirm
+                size="lg"
+                @confirmed="deleteCondition(i)"
```
Fajfa: Are we deleting condition here? or should it be called something else

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,31 +119,41 @@ export default {
...
   extends: base,
+
+  computed: {
+    documentationURL () {
+      // eslint-disable-next-line no-undef
+      const [year, month] = VERSION.split('.')
+      return `https://docs.cortezaproject.org/corteza-docs/${year}.${month}/integrator-guide/compose-configuration/index.html`
+    },
+
+    fieldOptions () {
```
Fajfa: vue-select can be configured so you dont need the map here, you could also avoid the filter by extending the selectable prop

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -42,31 +119,41 @@ export default {
...
+        }
+      })
+    },
+  },
+
+  methods: {
+    addCondition () {
+      if (this.options.fieldConditions.length >= this.fieldOptions.length) return
+      this.options.fieldConditions.push({
+        field: null,
```
Fajfa: undefined or ''

lib/js/src/compose/types/page-block/record.ts
```diff
@@ -2,12 +2,19 @@ import { PageBlock, PageBlockInput, Registry } from './base'
 
 const kind = 'Record'
 
+interface FieldConditions {
```
Fajfa: Interface is never plural, this defines one field condition

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
Fajfa: No need for that. Even if you wanted to do that you would disable the button that adds a new rule and not prevent it from being added in the function itself

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
EmmyMay: image

This is what ml-auto looks like.

client/web/compose/src/components/PageBlocks/RecordEditor.vue
```diff
@@ -14,56 +14,62 @@
...
         :key="field.id"
         class="mt-3"
       >
         <field-editor
-          v-if="isFieldEditable(field)"
+          v-if="isFieldEditable(field) && canDisplay(index)"
           v-bind="{ ...$props, errors: fieldErrors(field.name) }"
           class="field"
           :field="field"
+          :evaluating="evaluating"
```
Fajfa: Instead of passing the prop and impacting the field editors all over the place. Either do the spinner in this file or don't do it at all, since why would you even display a spinner if you don't know that field will even be displayed?
I would v-if them and that is it

Also beware, the v-if you added now only hides the editing and not the viewing, you'll need a wrap like on the record base and that wrap will have to have the class from the v-for element, since ghost margins can happen otherwise

client/web/compose/src/components/PageBlocks/RecordBase.vue
```diff
@@ -18,37 +18,41 @@
         :key="index"
         class="d-flex flex-column mb-3 px-3"
```
Fajfa: Since the body of the v-for element is now optional the classes on it will probably have to be moved to the optional section. Otherwise there will be ghost margins and paddings

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+
+  methods: {
+    evaluateExpressions () {
+      if (!this.block.options.fieldConditions) return
+      this.evaluating = true
+      const { expressions, variables } = this.prepareFieldConditionsData()
+      return this.$SystemAPI
+        .expressionEvaluate({ variables, expressions })
+        .then(res => {
+          this.fieldConditions = res
```
Fajfa: How about computing an array of field names that are hidden? And just checking if that field is inside it, you can even use a set, to avoid duplicate items. And it's very fast when it comes to checking if it has an item.

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+    return {
+      fieldConditions: {},
+      evaluating: false,
+    }
+  },
+
+  methods: {
+    evaluateExpressions () {
+      if (!this.block.options.fieldConditions) return
+      this.evaluating = true
```
Fajfa: I dont think evaluation should be visible to the user, since if field is hidden then you don't need to show a spinner. Otherwise if it gets hidden then it just disappears. So no need for evaluating flag.

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+      const { expressions, variables } = this.prepareFieldConditionsData()
+      return this.$SystemAPI
+        .expressionEvaluate({ variables, expressions })
+        .then(res => {
+          this.fieldConditions = res
+          this.evaluating = false
```
Fajfa: The reason why finally exists is precisely this, so you don't have to have two of the same pieces of code in then and catch

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+          return res
+        }).catch(err => {
+          this.evaluating = false
+          return err
```
Fajfa: Look at how error handling is done in other cases, we should display a notification here

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+  data () {
+    return {
+      fieldConditions: {},
+      evaluating: false,
+    }
+  },
+
+  methods: {
+    evaluateExpressions () {
+      if (!this.block.options.fieldConditions) return
```
Fajfa: This should probably check for length no? Since that's an array

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+      this.block.options.fieldConditions.forEach(fc => {
+        expressions[fc.field] = fc.condition
+      })
+      return { expressions, variables }
+    },
+
+    canDisplay (index) {
+      let canDisplay
+      // ToDO: check if fieldID can be used instead of index
+      const similar = Object.keys(this.fieldConditions).find(k => k === this.fields[index].fieldID)
```
Fajfa: So as mentioned above, this check should be a simple set.has(fieldName), and call the method isFieldHidden for example. One of the reasons is, how to handle system fields, which don't have fieldID

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -156,6 +156,7 @@
       :ok-title="$t('label.saveAndClose')"
       ok-variant="primary"
       :cancel-title="$t('label.cancel')"
+      :ok-disabled="anyInvalidState"
```
Fajfa: We dont need changes in this file right?

client/web/compose/src/components/Public/Page/Block/Modal.vue
```diff
@@ -145,6 +145,7 @@ export default {
 .position-initial {
   position: initial;
 }
+
```
Fajfa: Guessing auto lint?

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+          <div
+            class="d-flex"
+          >
+            {{ $t('record.fieldConditions.label') }}
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1"
+              :disabled="block.options.fields.filter(f => !f.isRequired).length === block.options.fieldConditions.length"
+              @click="addRules"
```
Fajfa: Either add a placeholder like we discussed or preselect the first available field.

Also the method adds a rule, not multiple

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+            >
+              {{ $t('record.fieldConditions.help') }}
+            </b-button>
+          </div>
+        </template>
+      </b-form-group>
+
+      <b-table-simple
+        v-if="block.options.fieldConditions.length"
+        borderless
```
Fajfa, line 67: Remember i said do it like in module edit. It uses table small, so that means using the small prop here, it is also responsive="lg"

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+            <b-th />
+          </b-tr>
+        </b-thead>
+
+        <b-tbody>
+          <b-tr
+            v-for="(condition, i) in block.options.fieldConditions"
+            :key="i"
+          >
+            <b-td>
```
Fajfa: i suggest you add a "width: 33%; min-width: 200px" here

Also look at how namespace sidebar handles long names inside vue-select (elipsis)

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+                :options="block.options.fields"
+                append-to-body
+                :selectable="option => isSelectable(option)"
+                :get-option-label="getOptionLabel"
+                :reduce="option => option.fieldID"
+                class="bg-white"
+              />
+            </b-td>
+
+            <b-td>
```
Fajfa: i suggest you add a min-width: 300px here

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+                </b-input-group-prepend>
+                <b-form-input
+                  v-model="condition.condition"
+                  :placeholder="$t('record.fieldConditions.placeholder')"
+                />
+              </b-input-group>
+            </b-td>
+
+            <b-td
+              class="align-middle"
```
Fajfa: add 'text-right text-nowrap' also, width: 1%;

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+        <template #label>
+          <div
+            class="d-flex"
+          >
+            {{ $t('record.fieldConditions.label') }}
+
+            <b-button
+              variant="link"
+              class="p-0 ml-1"
+              :disabled="block.options.fields.filter(f => !f.isRequired).length === block.options.fieldConditions.length"
```
Fajfa: text-decoration-none, otherwise people think its a link, if you remember we talked about this

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -27,11 +27,109 @@
...
+            :key="i"
+          >
+            <b-td>
+              <vue-select
+                v-model="condition.field"
+                :options="block.options.fields"
+                append-to-body
+                :selectable="option => isSelectable(option)"
+                :get-option-label="getOptionLabel"
+                :reduce="option => option.fieldID"
```
Fajfa: The reason why I recommended the field name is that system fields for example don't have a fieldID.
Thats why selecting a system field doesn't work and produces warnings in console.

I would recommend using fieldID for user defined fields and using the name/alias for system fields.
That would ensure maximum compatibility. The easier way is to use just name/alias, but its not resistant to field name changes

client/web/compose/src/mixins/conditionalFields.js
```diff
@@ -0,0 +1,54 @@
...
+        }).catch(err => {
+          this.evaluating = false
+          return err
+        })
+    },
+
+    prepareFieldConditionsData () {
+      const expressions = {}
+      const variables = {
+        record: {
```
Fajfa: What about system fields, make sure you include them aswell. 
In normal filters they can be accessed via the record.recordID for example.

Fajfa (changes requested):
A couple of bugs, a couple of style improvements. And a different (simpler) approach

client/web/compose/src/components/PageBlocks/RecordEditor.vue
EmmyMay: I believe I am hiding the record viewing on recordbase and the editing in editor. Or what do you mean?

Fajfa (approved):

PR #776, opened 2023-01-13
Fix Number type record list filter

### LINKED ISSUE #759: IN operator not working on number multi value field
## Description of the bug

_When you have a multi value number field and you want to filter with `contains` or `not contains` the filter doesn't return anything. (It is working fine for the other fields, it looks like the problem is only with the number field)_

## Reproduction steps

1. Create a multi value number field
2. Go on the public page and create couple multi value records
3. Try to filter out with the contains filter
4. Notice that no records are returned

## Expected behavior 

_You should be able to filter multi value number fields as you can with text or other fields._

## Video

https://user-images.githubusercontent.com/53232926/211555503-26db0e8d-be6e-4e79-8d1c-0b9550143935.webm

## Versions

Branch: 2023.3.x and 2022.9.x

vicpatel:
Uses string values of number kind fields to build query for record-list filter.

Please take a look.


vicpatel:
This is cherry-picked to base branch!

PR #780, opened 2023-01-13
Show Warning Message for Unsaved Changes in Compose

### LINKED ISSUE #733: Show warning message for unsaved changes before navigating to another page
## Ref
_Block and Page need to both be Saved in Order to save a Block change only_ section - 

## TODO
- [x] detect if a block has been updated in page builder (compare current and previous object) and "save" btn was pressed
- [x] add the warning message
- [x] prevent the route update / route param update upon pressing "cancel"
- [x] keep block changes upon pressing "cancel"
- [x] update route / route params upon pressing "ok"

yonkov:
image
Image 1: Block has beem updated but the page has not been saved yet

image
Image 2: Browser confirmation message when user leaves the page without saving the changes.

### DIFF
--- a/client/web/compose/src/components/Common/Grid.vue
+++ b/client/web/compose/src/components/Common/Grid.vue
@@ -29,6 +29,8 @@
         }"
         style="touch-action: none;"
         v-bind="{ ...item }"
+        @moved="onBlockUpdated(index)"
+        @resized="onBlockUpdated(index)"
       >
         <slot
           :block="blocks[item.i]"
@@ -186,6 +188,11 @@ export default {
         ({ x, y, w, h, i }) => new compose.PageBlockMaker({ ...this.blocks[i], xywh: [x, y, w, h] }),
       ))
       this.recalculateBoundingRect()
+    },
+
+    // emit event when block has been moved or resized
+    onBlockUpdated (index) {
+      this.$emit('item-updated', index)
     },
   },
 }
--- a/client/web/compose/src/views/Admin/Pages/Builder.vue
+++ b/client/web/compose/src/views/Admin/Pages/Builder.vue
@@ -50,6 +50,7 @@
       :blocks="page.blocks"
       editable
       @change="updatePageBlockGrid"
+      @item-updated="onBlockUpdated"
     >
       <template
         slot-scope="{ boundingRect, block, index }"
@@ -63,6 +64,17 @@
             class="toolbox border-0 p-2 m-0 text-light text-center"
             data-test-id="block-toolbox"
           >
+            <div
+              v-if="unsavedBlocks.has(index)"
+              :title="$t('tooltip.unsavedChanges')"
+              class="btn border-0"
+            >
+              <font-awesome-icon
+                :icon="['fas', 'exclamation-triangle']"
+                class="text-warning"
+              />
+            </div>
+
             <b-button
               data-test-id="button-edit"
               :title="$t('tooltip.edit.block')"
@@ -96,7 +108,6 @@
                 :icon="['far', 'copy']"
               />
             </b-button>
-
             <c-input-confirm
               class="ml-1"
               size="md"
@@ -272,6 +283,7 @@ export default {
       page: undefined,
       blocks: [],
       board: null,
+      unsavedBlocks: new Set(),
     }
   },
 
@@ -365,6 +377,7 @@ export default {
       immediate: true,
       handler (pageID) {
         this.page = undefined
+        this.unsavedBlocks.clear()
 
         if (pageID) {
           const { namespaceID, name } = this.namespace
@@ -381,6 +394,14 @@ export default {
 
   mounted () {
     window.addEventListener('paste', this.pasteBlock)
+  },
+
+  beforeRouteUpdate (to, from, next) {
+    this.checkUnsavedBlocks(next)
+  },
+
+  beforeRouteLeave (to, from, next) {
+    this.checkUnsavedBlocks(next)
   },
 
   destroyed () {
@@ -409,20 +430,30 @@ export default {
     deleteBlock (index) {
       this.blocks.splice(index, 1)
       this.page.blocks = this.blocks
+      this.unsavedBlocks.add(index)
     },
 
     updatePageBlockGrid (blocks) {
       this.blocks = blocks
     },
 
+    onBlockUpdated (index) {
+      this.unsavedBlocks.add(index)
+    },
+
     updateBlocks () {
       const block = compose.PageBlockMaker(this.editor.block)
       this.page.blocks = this.blocks
 
+      /**
+       * Check if an existing block has been updated or a new block has been added
+       */
       if (this.editor.index !== undefined) {
         this.page.blocks.splice(this.editor.index, 1, block)
+        this.unsavedBlocks.add(this.editor.index)
       } else {
         this.page.blocks.push(block)
+        this.unsavedBlocks.add(this.page.blocks.length - 1)
       }
 
       this.editor = undefined
@@ -461,6 +492,7 @@ export default {
           const mergedPage = new compose.Page({ namespaceID, ...page, blocks: this.blocks })
 
           this.updatePage(mergedPage).then((page) => {
+            this.unsavedBlocks.clear()
             this.toastSuccess(this.$t('notification:page.saved'))
             if (closeOnSuccess) {
               this.$router.push({ name: 'admin.pages' })
@@ -558,7 +590,6 @@ export default {
         const maxY = this.blocks.map((block) => block.xywh[1]).reduce((acc, val) => {
           return acc > val ? acc : val
         }, 0)
-
         block.xywh = [0, maxY + 2, 3, 3]
       }
 
@@ -572,6 +603,11 @@ export default {
       } else {
         this.toastErrorHandler(this.$t('notification:page.duplicateFailed'))
       }
+    },
+
+    // Trigger browser dialog on page leave to prevent unsaved changes
+    checkUnsavedBlocks (next) {
+      next(!this.unsavedBlocks.size || window.confirm(this.$t('build.unsavedChanges')))
     },
   },
 }
--- a/locale/en/corteza-webapp-compose/page.yaml
+++ b/locale/en/corteza-webapp-compose/page.yaml
@@ -16,6 +16,7 @@ copyOf: "Copy of {{title}}"
 build:
   addBlock: Add block
   selectBlockTitle: Select type of the new block
+  unsavedChanges: Unsaved changes will be lost. Do you wish to leave the page?
 createLabel: Create page
 edit:
   create: Create page
@@ -45,6 +46,7 @@ pageMoveFailed: Could not move this page
 reordered: Page reordered
 title: List of Pages
 tooltip:
+  unsavedChanges: Unsaved Changes
   view: View page
   edit:
     page: Edit page
--- a/client/web/compose/src/components/faIcons.js
+++ b/client/web/compose/src/components/faIcons.js
@@ -60,6 +60,7 @@ import {
   faKey,
   faQuestion,
   faSync,
+  faExclamationTriangle,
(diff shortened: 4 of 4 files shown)


Fajfa (changes requested):
We should detect the change differently to avoid actually comparing the objects

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -29,6 +29,7 @@
         }"
         style="touch-action: none;"
         v-bind="{ ...item }"
+        @moved="movedEvent(item.i)"
```
Fajfa: Use the index from the for loop above

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -187,6 +188,10 @@ export default {
       ))
       this.recalculateBoundingRect()
     },
+
+    movedEvent (index) {
+      this.$root.$emit('has-unsaved-changes', +index + 1)
```
Fajfa: no need to +index +1
Also this shouldn't be a root event, just emit to parent (Builder)

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -6,7 +6,7 @@
       :class="blockClass"
     >
       <b-card-header
-        v-if="showHeader"
+        v-if="showHeader || hasUnsavedChange"
```
Fajfa: this should not be here, showHeader is there for a reason

client/web/compose/src/components/PageBlocks/Wrap/Card.vue
```diff
@@ -24,6 +24,16 @@
               <slot name="title-badge" />
             </h5>
 
+            <div
+              v-if="hasUnsavedChange"
```
Fajfa: This should be the first icon of before the options section(not part of its if but on that side), no need for extra div

client/web/compose/src/components/PageBlocks/Wrap/base.vue
```diff
@@ -18,6 +18,12 @@ export default {
     },
   },
 
+  data () {
+    return {
+      unsavedChange: false,
```
Fajfa: Please no, you can pass a prop to it that will determine this. And you evaluate it on the parent (builder)

client/web/compose/src/components/PageBlocks/Wrap/base.vue
```diff
@@ -49,6 +55,7 @@ export default {
         this.block.description,
         this.block.options.magnifyOption,
         this.block.options.refreshEnabled,
+        this.unsavedChange,
```
Fajfa: this would just be the prop

client/web/compose/src/components/PageBlocks/Wrap/base.vue
```diff
@@ -58,6 +65,23 @@ export default {
         this.block.options.refreshEnabled,
       ].some(c => !!c)
     },
+    hasUnsavedChange () {
```
Fajfa: Will not be needed

client/web/compose/src/components/PageBlocks/Wrap/base.vue
```diff
@@ -58,6 +65,23 @@ export default {
...
+      this.unsavedChanges()
+      return this.unsavedChange
+    },
+  },
+
+  methods: {
```
Fajfa: Will not be needed

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -272,6 +276,7 @@ export default {
       page: undefined,
       blocks: [],
       board: null,
+      hasUpdatedBlocks: false,
```
Fajfa: Not like this.

1. Create a Set, that will hold indexes of changed blocks
2. if it has any items that means blocks have been modified/added
3. On block save or move the blockIndex(this index wont change if you move blocks around) of that block should be added to the set
4. You can always check if set contains that index very easily, to determine if a certain block has been modified
3. Reset that set on save

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -241,6 +241,10 @@ export default {
     namespaces: 'page',
   },
 
+  beforeRouteLeave (to, from, next) {
```
Fajfa: This should be with the rest of the navigation guards

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -573,6 +597,21 @@ export default {
         this.toastErrorHandler(this.$t('notification:page.duplicateFailed'))
       }
     },
+
+    checkBlocksForUnsavedChanges (next) {
+      if (this.hasUpdatedBlocks) {
```
Fajfa: Here you will check if set contains any items

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -573,6 +597,21 @@ export default {
...
+        const okMessage = window.confirm(this.$t('build.unsavedChanges'))
```
Fajfa: next(window.confirm(this.$t('build.unsavedChanges'))) simplifies all of this, please use it

locale/en/corteza-webapp-compose/page.yaml
```diff
@@ -16,6 +16,7 @@ copyOf: "Copy of {{title}}"
 build:
   addBlock: Add block
   selectBlockTitle: Select type of the new block
+  unsavedChanges: Do you really want to leave? you have unsaved changes!
```
Fajfa: This needs to be less condescending

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -272,6 +278,7 @@ export default {
       page: undefined,
       blocks: [],
       board: null,
+      unsavedBlockIds: new Set([]),
```
Fajfa: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -50,6 +50,7 @@
       :blocks="page.blocks"
       editable
       @change="updatePageBlockGrid"
+      @has-unsaved-changes="hasUnsavedChanges"
```
Fajfa: Better event/function names

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -573,6 +593,20 @@ export default {
         this.toastErrorHandler(this.$t('notification:page.duplicateFailed'))
       }
     },
+
+    checkBlocksForunsavedBlockIds (next) {
```
Fajfa: checkUnsavedBlocks

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -573,6 +593,20 @@ export default {
...
+      if (this.unsavedBlockIds.size) {
```
Fajfa: Get it to 4 lines

Fajfa (changes requested):

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -29,6 +29,7 @@
         }"
         style="touch-action: none;"
         v-bind="{ ...item }"
+        @moved="movedEvent(index)"
```
Fajfa: to keep consitency lets rename movedEvent to onMove

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -187,6 +188,10 @@ export default {
       ))
       this.recalculateBoundingRect()
     },
+
+    movedEvent (index) {
+      this.$emit('move-block', index)
```
Fajfa: events should always say what happened. In your case move-block is saying something will happen.
For example `moved`, `item-moved`(since grid doesnt know its items are blocks)

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -50,6 +50,7 @@
       :blocks="page.blocks"
       editable
       @change="updatePageBlockGrid"
+      @move-block="moveBlock"
```
Fajfa: same here with event name, onBlockMoved instead of moveBlock, since moveBlock implies that method will move it

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -424,7 +445,7 @@ export default {
       } else {
         this.page.blocks.push(block)
       }
-
+      this.unsavedBlocks.add(this.editor.index)
```
Fajfa: I have a feeling this wont work with adding new blocks.
Also new line between blocks (ifs, functions, basically anything that has a {})

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -29,6 +29,7 @@
         }"
         style="touch-action: none;"
         v-bind="{ ...item }"
```
Fajfa, line 31: I can see we'll need a resize event aswell, when a block is resized

Fajfa, line 31: I can see we'll need a resize event aswell, when a block is resized

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -63,6 +64,13 @@
             class="toolbox border-0 p-2 m-0 text-light text-center"
             data-test-id="block-toolbox"
           >
+            <font-awesome-icon
```
Fajfa: I suggest you wrap the icon in a div and give it the "btn border-0" and just keep text-warning on the icon, it will make it look better (like the rest of the toolbox). Also move the title and v-if to the wrapping div please.

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -365,7 +373,7 @@ export default {
       immediate: true,
       handler (pageID) {
         this.page = undefined
-
+        this.unsavedBlocks.clear()
```
Fajfa, line 380: Do respect new lines around ifs and such

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -573,6 +594,10 @@ export default {
         this.toastErrorHandler(this.$t('notification:page.duplicateFailed'))
       }
     },
+
+    checkUnsavedBlocks (next) {
```
Fajfa, line 609: I know we talked about not commenting code, but I think it would be beneficial if there was a comment here describing what this is for

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -425,6 +451,14 @@ export default {
         this.page.blocks.push(block)
       }
 
+      // if a new block has been added
+      if (this.page.blocks[this.page.blocks.length - 1].blockID === NoID) {
```
Fajfa: Instead of checking it twice do it in the above ifs, since they check the exact same thing

client/web/compose/src/components/Common/Grid.vue
```diff
@@ -187,6 +189,11 @@ export default {
       ))
       this.recalculateBoundingRect()
     },
+
+    // emit event when block has been moved or resized
+    onBlockUpdated (index) {
+      this.$emit('moved-resized', index)
```
Fajfa: probably should be called 'item-updated'

Fajfa (changes requested):

PR #805, opened 2023-01-20
Fix error for unconfigured sources

### LINKED ISSUE #766: Unconfigured source in geometry field causing issues
## Steps to reproduce
1. Create a geometry page block
2. Add a source event without selecting a module
3. See error in the console

Image

## Expected behaviour
We shouldn't save unconfigured sources

### Requirements

- [x] Geometry field should be required for isValid check
- [x] Change order of inputs to Select, Geometry field, title, prefilter ....
- [x] Change Select module to Select
- [x] Module and Geometry field should be asterisked

EmmyMay:
https://github.com/cortezaproject/corteza/issues/766

### DIFF
--- a/client/web/compose/src/components/PageBlocks/GeometryBase.vue
+++ b/client/web/compose/src/components/PageBlocks/GeometryBase.vue
@@ -149,7 +149,7 @@ export default {
         zoomMax,
       }
 
-      Promise.all(this.options.feeds.map((feed, idx) => {
+      Promise.all(this.options.feeds.filter(f => f.isValid()).map((feed, idx) => {
         return this.findModuleByID({ namespace: this.namespace, moduleID: feed.options.moduleID })
           .then(module => {
             // Interpolate prefilter variables
--- a/client/web/compose/src/components/PageBlocks/GeometryConfigurator/FeedSource/configs/Record.vue
+++ b/client/web/compose/src/components/PageBlocks/GeometryConfigurator/FeedSource/configs/Record.vue
@@ -28,16 +28,21 @@
           horizontal
           :label-cols="3"
           breakpoint="md"
-          :label="$t('geometry.recordFeed.colorLabel')"
+          :label="$t('geometry.recordFeed.geometryFieldLabel')"
         >
-          <b-input-group>
-            <b-form-input
-              v-model="feed.options.color"
-              style="max-width: 50px;"
-              type="color"
-              debounce="300"
-            />
-          </b-input-group>
+          <b-form-select
+            v-model="feed.geometryField"
+            :options="geometryFields | optionizeFields"
+          >
+            <template slot="first">
+              <option
+                disabled
+                value=""
+              >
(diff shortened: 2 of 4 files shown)


client/web/compose/src/components/PageBlocks/GeometryBase.vue
```diff
@@ -149,7 +149,7 @@ export default {
         zoomMax,
       }
 
-      Promise.all(this.options.feeds.map((feed, idx) => {
+      Promise.all(this.options.feeds.filter(({ options }) => options.moduleID !== NoID).map((feed, idx) => {
```
Fajfa: I doubt this is the only thing needed for it to function no? You also need other configurations so, I would suggest you add a isValid method to the feeds and do it like this.options.feed.filter(f => f.isValid())

Fajfa (changes requested):

lib/js/src/compose/types/page-block/geometry/feed.ts
```diff
@@ -53,4 +53,8 @@ export default class Feed {
       }
     }
   }
+
+  isValid (): boolean {
+    return this.options.moduleID !== NoID
```
Fajfa: This is not the only thing needed in the feed for it to be valid. Try configuring it and see, might also be worth adding a * next to the required values (in the form next to labels))

Otherwise this is the right approach

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/GeometryConfigurator/FeedSource/configs/Record.vue
```diff
@@ -28,16 +28,21 @@
...
-          <b-input-group>
-            <b-form-input
-              v-model="feed.options.color"
-              style="max-width: 50px;"
-              type="color"
-              debounce="300"
-            />
-          </b-input-group>
+          <b-form-select
+            v-model="feed.geometryField"
```
EmmyMay, line 34: Here you can see the v-model is tied to the geometryField not the deeper options.geometryField

lib/js/src/compose/types/page-block/geometry/feed.ts
```diff
@@ -53,4 +53,8 @@ export default class Feed {
       }
     }
   }
+
+  isValid (): boolean {
+    return this.options.moduleID !== NoID && this.geometryField !== ''
```
EmmyMay: https://github.com/cortezaproject/corteza/blob/8a7a10edaccb4b9d271ec4e100e22201c5c3e27a/lib/js/src/compose/types/page-block/geometry/feed.ts#L10

I noticed that the geometryField options is not updated when a field is not selected. My question is should it even be in the options? It either looks like we have two identical things and we only really need one or we are not updating it as we should. Or maybe I am the one missing something

Fajfa (approved):

PR #826, opened 2023-01-27
Fix record list filter issue inside a modal

### LINKED ISSUE #779: Record list filter issue inside modals
## Description of the bug

_When you have a record list in a specific record which is set to open up in a modal, the filters are not shown as they should depending on where they are on the screen. The first filter for example is cuted out because it tries to stay inside of the modal but it is not fixed. (Screenshot will provide better context.)_

## Reproduction steps

1. Create a record list that opens up records in a modal
2. Have a record list also inside of the record 
3. once you open the record and it is shown in modal, try to use the filters on the record list below and see the outcome.

## Expected behavior 

_Filter modal inside of record modal should be displayed normally and not to be cuted._

## Screenshots

Image

## Versions

Branch: 2023.3.x

yonkov:


Fajfa (changes requested):
Please also define the css class here. So we don't forget where its defined.
We'll merge the global classes into one file soon

Fajfa (approved):

PR #830, opened 2023-01-30
2022.9.x fix pgsql id columns datatype

### LINKED ISSUE #682: Properly handle IDs for PGSQL
## Description

Corteza uses `uint64` for the ID but PGSQL does not support unsigned integers so the maximum number it can save is `2^63-1` where the uint64 is `2^64-1`.

PGSQL driver knows how to convert `uint64` to `int64` before saving will start failing once the number passes `2^63` range.

## Solution

This should be fixed on the database driver side (`store/drivers/rdbms/pgsql`) and not in the service layer (ID should remain as `uint64`).
The most probable solution will require you to change the data type of ID and reference table columns (ref where to start `store/adapters/rdbms/drivers/postgres/dialect.go` and `store/adapters/rdbms/drivers/sqlserver`).

Utilise `numeric` data type (https://www.postgresql.org/docs/current/datatype-numeric.html) to solve this issue (check how I've done it for mssql).

This must be fixed on `2023.3.x`, `2022.9.x`.
This would optimally be addressed on `2022.3.x`.

## Potential issues

- how will we do the migration code if any at all

KinyaElGrande:
Change PSQL ID column datatype from `int8` to `numeric` to enable support for  Corteza ID ` uint64` data type.

Fixes: https://github.com/cortezaproject/corteza/issues/682

### DIFF
--- a/server/store/adapters/rdbms/upgrade_fixes.go
+++ b/server/store/adapters/rdbms/upgrade_fixes.go
@@ -42,6 +42,7 @@ var (
 		fix_2022_09_00_addRevisionOnComposeRecords,
 		fix_2022_09_00_addMetaOnComposeRecords,
 		fix_2022_09_00_addMissingNodeIdOnFederationMapping,
+		fix_2022_09_07_changePostgresIdColumnsDatatype,
 	}
 )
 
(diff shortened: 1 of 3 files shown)


server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -429,6 +430,34 @@ func fix_2022_09_00_addMissingNodeIdOnFederationMapping(ctx context.Context, s *
 	)
 }
 
+func fix_2022_09_07_changePostgresIdColumnsDatatype(ctx context.Context, s *Store) (err error) {
+	var tableName string
+	if s.DB.DriverName() != "postgres" {
+		return
+	}
+
+	rows, err := s.DB.QueryContext(ctx, "SELECT table_name FROM INFORMATION_SCHEMA.COLUMNS  WHERE column_name = 'id'")
```
tjerman: Consider somehow constraining these ID fields only to the tables that Corteza creates.
A potential edge case is when someone would add some of their own tables for DAL and so forth.

tjerman (approved):
Apart from that one comment; lgtm 👍

tjerman (commented):
Looks like it'll do the job 👍 

I like how you've utilised SQL filters to weed out the tables we don't want, but I'm not the biggest fan of how you've constructed the DB filter. Namely, firstly pushing all of the models into one slice, then making a string slice from the first slice, then making a string from that string slice.
Can you see if you can shorten it out a bit? Hint: IIRC, you can pass string slices to the DB level 😉

PR #836, opened 2023-02-03
Add navigation for previous and next page on the record page

### LINKED ISSUE #480: Allow users to navigate to previous and next page [PlanID:2023.3/15/1]
## About

Expand navigation when inside a record page to allow you to go to the next/prev record.

This could be achieved by extending what the record read endpoint returns by including the next/prev page cursors.
If we do decide to use paging cursors we need to be careful how to handle update like operations since they could break the order and not work as expected.

## Important

if you're editing records it might not be stable to rely on page cursors since they are applied as filters.
In some cases, the next page cursor could mean the exact same record.

## User stories

As a user I can use prev/next buttons on record page when coming from a record-list because  I want to be able to access next/prev page

kelanik8:
Expand navigation when inside a record page to allow you to go to the next/prev record.

image

 - [x] Code builds
 - [x] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/Common/RecordToolbar.vue
+++ b/client/web/compose/src/components/Common/RecordToolbar.vue
@@ -33,28 +33,24 @@
       >
         <b-button-group v-if="recordNavigation.prev || recordNavigation.next">
           <b-button
-            id="tooltip-target-prev-page"
-            v-b-tooltip.hover
             pill
             size="lg"
             variant="outline-primary"
             class="mr-2"
-            :disabled="!recordNavigation.prev"
+            :disabled="!record || processing || !recordNavigation.prev"
             :title="$t('recordNavigation.prev')"
-            @click="$emit('update-navigation', recordNavigation.prev)"
+            @click="navigateToRecord(recordNavigation.prev)"
           >
             <font-awesome-icon :icon="['fas', 'angle-left']" />
           </b-button>
 
           <b-button
-            id="tooltip-target-next-page"
-            v-b-tooltip.hover
             size="lg"
             pill
             variant="outline-primary"
-            :disabled="!recordNavigation.next"
+            :disabled="!record || processing || !recordNavigation.next"
             :title="$t('recordNavigation.next')"
-            @click="$emit('update-navigation', recordNavigation.next)"
+            @click="navigateToRecord(recordNavigation.next)"
           >
             <font-awesome-icon :icon="['fas', 'angle-right']" />
           </b-button>
@@ -175,6 +171,7 @@
 
 <script>
 import { compose, NoID } from '@cortezaproject/corteza-js'
+import { throttle } from 'lodash'
 
 export default {
   i18nOptions: {
@@ -284,6 +281,12 @@ export default {
 
       return this.isDeleted && this.record.canUndeleteRecord && !this.processing && this.record.recordID !== NoID
     },
+  },
+
+  methods: {
+    navigateToRecord: throttle(function (recordID) {
+      this.$emit('update-navigation', recordID)
+    }, 500),
   },
 }
 </script>
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -1217,29 +1217,18 @@ export default {
       this.processing = false
     },
 
-    handleRowClicked ({ r: { recordID } }) {
-      const { moduleID, namespaceID } = this.recordListModule
+    async handleRowClicked ({ r: { recordID } }) {
+      if ((this.options.editable && this.editing) || (!this.recordPageID && !this.options.rowViewUrl)) {
+        return
+      }
 
-      if (this.block.options.enableRecordPageNavigation) {
-        const { pageCursor, nextPage, prevPage } = this.filter
-
-        this.loadPaginationRecords({
-          recordListModule: this.recordListModule,
-          moduleID,
-          namespaceID,
-          filterCursors: [prevPage, pageCursor, nextPage],
+      if (this.options.enableRecordPageNavigation) {
+        await this.loadPaginationRecords({
           filter: {
             ...this.filter,
             limit: Math.min(this.pagination.count, 100),
           },
-          options: {
-            showRecordNavigationTooltip: this.block.options.showRecordNavigationTooltip,
-          },
         })
-      }
-
-      if ((this.options.editable && this.editing) || (!this.recordPageID && !this.options.rowViewUrl)) {
-        return
       }
 
       const pageID = this.recordPageID
--- a/client/web/compose/src/store/ui.js
+++ b/client/web/compose/src/store/ui.js
@@ -5,8 +5,8 @@ const types = {
   completed: 'completed',
   setRecordPagination: 'setRecordPagination',
   clearRecordPagination: 'clearRecordPagination',
-  clearRecordPageNavigation: 'clearRecordPageNavigation',
-  setClearRecordPageNavigation: 'setClearRecordPageNavigation',
+  recordPaginationUsable: 'recordPaginationUsable',
+  setRecordPaginationUsable: 'setRecordPaginationUsable',
 }
 
 export default function (ComposeAPI) {
@@ -16,9 +16,8 @@ export default function (ComposeAPI) {
     state: {
       loading: false,
       pending: false,
-      recordPaginationIds: [],
-      recordPageVisited: null,
-      clearRecordPageNavigation: true,
+      recordPaginationIDs: [],
+      recordPaginationUsable: false,
     },
 
     getters: {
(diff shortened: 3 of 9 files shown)


client/web/compose/src/store/ui.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+
+    actions: {
+      async loadPaginationRecords ({ commit }, { filter, moduleID, namespaceID, recordListModule, queries, incTotal, incPageNavigation } = {}) {
+        commit(types.pending)
+
+        const [prevRecords, nextRecords] = await Promise.all(queries.map((query) => {
+          return ComposeAPI.recordList({ ...filter, moduleID, namespaceID, query, incTotal, incPageNavigation })
+            .then(({ set }) => {
+              const records = set.map(r => new compose.Record(r, recordListModule))
```
Fajfa: No need to map anything here since all you do is return the recordID's anyway, that also saves you from passing the recordListModule param

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
       this.processing = false
     },
 
-    handleRowClicked ({ r: { recordID } }) {
+    handleRowClicked ({ r: { recordID, values, ...record } }) {
+      const { moduleID, namespaceID } = this.recordListModule
+      let prevAndNextRecordFilters = []
+      const field = this.options.presort.split(' ')[0] || 'createdAt'
```
Fajfa: I was thinking about this, and it could cause issues. The safest bet is probably just using the this.filter.sort. Since the sort can change. Also you shouldn't fallback to createdAt, since that could mess up the order.

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
+      const value = values[field] || record[field]
+
+      const moduleField = this.recordListModule.findField(field)
+
+      prevAndNextRecordFilters.push({
+        condition: 'Where',
+        name: moduleField.name,
+        operator: '>=',
```
Fajfa: Should the operator change based on if sort is DESC or ASC?
I think it should since the order changes if based on the sort

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
-    handleRowClicked ({ r: { recordID } }) {
+    handleRowClicked ({ r: { recordID, values, ...record } }) {
+      const { moduleID, namespaceID } = this.recordListModule
+      let prevAndNextRecordFilters = []
+      const field = this.options.presort.split(' ')[0] || 'createdAt'
+      const value = values[field] || record[field]
+
+      const moduleField = this.recordListModule.findField(field)
+
+      prevAndNextRecordFilters.push({
```
Fajfa: Instead of creating an array and pushing, just define the array from the start, with the two objects

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
+        condition: 'Where',
```
Fajfa: You dont need the condition

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
     },
 
-    handleRowClicked ({ r: { recordID } }) {
+    handleRowClicked ({ r: { recordID, values, ...record } }) {
+      const { moduleID, namespaceID } = this.recordListModule
+      let prevAndNextRecordFilters = []
+      const field = this.options.presort.split(' ')[0] || 'createdAt'
+      const value = values[field] || record[field]
+
+      const moduleField = this.recordListModule.findField(field)
```
Fajfa: You can deconstruct to just get the name and kind here 
`{ name, kind } = findField || {}`

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
+
+      prevAndNextRecordFilters.push({
+        condition: 'Where',
+        name: moduleField.name,
+        operator: '>=',
+        value: value,
```
Fajfa: no need to do value: value, you can just do value,

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
+        condition: 'Where',
+        name: moduleField.name,
+        operator: '<',
+        value: value,
+        kind: moduleField.kind,
+      })
+
+      const query = queryToFilter(this.query, this.prefilter, this.recordListModule.filterFields(this.options.fields), this.recordListFilter)
+
+      prevAndNextRecordFilters = prevAndNextRecordFilters.map(filter => getFieldFilter(filter.name, filter.kind, filter.value, filter.operator))
```
Fajfa: Instead of maping , just use this getFieldFilter when you define the initial array, and provide (name, kind, value, operator)

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1168,7 +1172,51 @@ export default {
...
+
+      const queries = prevAndNextRecordFilters.map(filterQuery => query ? [query, filterQuery].join(' AND ') : filterQuery)
+
+      this.loadPaginationRecords({
+        recordListModule: this.recordListModule,
+        moduleID,
+        namespaceID,
+        queries,
+        incTotal: false,
```
Fajfa: this should auto to false anyway, just provide the things you absolutley need.
moduleID, namespaceID, queries, filter.

incTotal, incPageNav and pageCursor should be false/undefined if you dont provide them so.

Also prevAndNextRecordFilters looks to me you don't need that

client/web/compose/src/store/ui.js
```diff
@@ -0,0 +1,81 @@
+import { compose } from '@cortezaproject/corteza-js'
+
+const types = {
+  loading: 'loading',
+  loaded: 'loaded',
+  pending: 'pending',
+  completed: 'completed',
+  setRecordIds: 'setRecordIds',
```
Fajfa: setRecordPagination, clearRecordPagination

client/web/compose/src/store/ui.js
```diff
@@ -0,0 +1,81 @@
...
+      recordPaginationIds (state) {
+        return state.recordPaginationIds
+      },
+    },
+
+    actions: {
+      async loadPaginationRecords ({ commit }, { filter, moduleID, namespaceID, recordListModule, queries, incTotal, incPageNavigation } = {}) {
+        commit(types.pending)
+
+        const [prevRecords, nextRecords] = await Promise.all(queries.map((query) => {
```
Fajfa: Instead of awaiting, please return the promise and handle anything that needs to be handled in the `.then` for the Promise.all

Look at any other async action in any another store

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -41,18 +41,22 @@
         :hide-clone="inCreating"
         :hide-add="inCreating"
         :show-record-modal="showRecordModal"
+        :next-prev-navigation-options="nextPrevNavigationOptions"
```
Fajfa, line 44: Since you'll do it with a slot, you won't need a prop or events

Fajfa (changes requested):

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -27,6 +27,29 @@
         </b-button>
       </div>
 
+      <b-button-group class="ml-auto">
+        <b-button
+          v-if="nextPrevNavigationOptions.prev"
```
Fajfa: Make sure the whole section is not shown (button group), if the prop is missing. Also prop should be required: false, by default. So you can rely on it

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -27,6 +27,29 @@
         </b-button>
       </div>
 
+      <b-button-group class="ml-auto">
```
Fajfa: I think ml-auto can fail if right or left section is hidden. That means it will be centered right if right section doesnt exist. I suggest you use justify-content-between for the row class

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -27,6 +27,29 @@
...
+        <b-button
+          v-if="nextPrevNavigationOptions.prev"
+          pill
+          size="lg"
+          variant="outline-primary"
+          class="mr-2"
+          @click="$emit('update-navigation:prev')"
```
Fajfa: Make sure button is disabled if there is no previous/next record.
Also a :title will be required since the button is icon only

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -205,6 +228,11 @@ export default {
       type: Boolean,
       required: false,
     },
+
+    nextPrevNavigationOptions: {
```
Fajfa: lets keep the name simple, record-navigation

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -27,6 +27,29 @@
...
+        >
+          <font-awesome-icon :icon="['fas', 'angle-left']" />
+        </b-button>
+
+        <b-button
+          v-if="nextPrevNavigationOptions.next"
+          size="lg"
+          pill
+          variant="outline-primary"
+          @click="$emit('update-navigation:next')"
```
Fajfa: I suggest having a single event, and using value to indicate the prev/next.
For example -1 for prev, 1 for next.

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -112,6 +116,14 @@ export default {
   },
 
   computed: {
+    ...mapGetters({
+      recordIDs: 'ui/recordPaginationIds',
+    }),
+
+    recordIdx () {
```
Fajfa: I would define this in the store and use a getter for it. 
for example getRecordNavigationIndex

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -146,6 +158,23 @@ export default {
 
       return this.$t(`page:public.record.${titlePrefix}.title`, { name: name || handle, interpolation: { escapeValue: false } })
     },
+
+    recordListModule () {
```
Fajfa: Pretty sure you don't need this

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -146,6 +158,23 @@ export default {
...
+      if (this.page.moduleID) {
+        return this.getModuleByID(this.page.moduleID)
+      } else {
+        return undefined
+      }
+    },
+
+    nextPrevNavigationOptions () {
+      const index = this.recordIdx !== undefined ? this.recordIdx : 1
```
Fajfa: How about having a getter in store again, that always just returns a boolean of next or previous item. And then you can just return an object with those two getters.

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -245,6 +274,23 @@ export default {
         this.$router.push({ name: 'page.record.edit', params: this.$route.params })
       }
     },
+
+    handleRedirectToPrevOrNext (index) {
```
Fajfa: i would accept 'prev'/'next' param. And just use the getter from the store to get the apropriate ID.

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -245,6 +274,23 @@ export default {
...
+        const route = {
+          name: 'page.record',
+          params: {
+            pageID: this.page.pageID,
+            recordID,
+          },
+          query: null,
+        }
+
+        this.$router.push(route)
```
Fajfa: It will probably have to be handled differently than this. Since you can have this open in a modal aswell. You'll need to come up with a solution, or atleast think about it and we can resolve it together if you're not sure.

client/web/compose/src/views/Admin/Modules/Records/View.vue
```diff
@@ -131,6 +134,14 @@ export default {
   },
 
   computed: {
+    ...mapGetters({
+      recordIDs: 'ui/recordPaginationIds',
+      getRecordNavigationIndex: 'ui/getRecordNavigationIndex',
+      nextRecordNavigationPrev: 'ui/nextRecordNavigationPrev',
+      nextRecordNavigationNext: 'ui/nextRecordNavigationNext',
+      getNextOrPrevRecordId: 'ui/getNextOrPrevRecordId',
```
Fajfa: You only need one getter for the navigation

client/web/compose/src/components/Common/RecordToolbar.vue
```diff
@@ -25,112 +26,140 @@
...
-            type="grow"
-          />
+        <b-button-group v-if="recordNavigation.prev || recordNavigation.next">
+          <b-button
+            pill
+            size="lg"
+            variant="outline-primary"
+            class="mr-2"
+            :disabled="!recordNavigation.prev"
+            @click="$emit('update-navigation', 'prev')"
```
Fajfa: You can emit the recordID itself here

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1187,6 +1191,24 @@ export default {
...
+        recordListModule: this.recordListModule,
+        moduleID,
+        namespaceID,
+        filterCursors: [pageCursor, nextPage],
+        filter: {
+          ...this.filter,
+          limit: Math.min(this.pagination.count, 100),
+        },
+        options: {
+          showRecordNavigationTooltip: this.block.options.showRecordNavigationTooltip,
```
Fajfa: Only use this option to check if navigation should be computed at the start of this method

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -245,6 +265,29 @@ export default {
         this.$router.push({ name: 'page.record.edit', params: this.$route.params })
```
Fajfa: On page change, check the flag in the store you set when you compute the recordID's
If its true, toggle it to false, if its false clear the array in store.

THIS IS ON VIEW PAGE

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -245,6 +265,29 @@ export default {
...
       }
     },
+
+    handleRedirectToPrevOrNext (value) {
+      const recordID = this.getNextOrPrevRecordId(this.recordID, value)
```
Fajfa: Since you'll emit value you won't need to recompute it here, call it recordID

`if (!recordID) return`

Fajfa (changes requested):

PR #845, opened 2023-02-07
Fix select field options translations showing when not saved

### LINKED ISSUE #815: Select field translations not show unless module is saved
## Branch version
23.3

## Steps to reproduce
1. Create a select field
2. Add an option
3. Save the change from the configurator modal
4. Reopen the field
5. Click on the translation button
6. See that no translations are shown

## Video with reproduction steps

Image

## Cause of the issue
The translations don't have a key so they're not saved on the BE

## Expected behaviour
Show the translations only if there are such (they're saved)

## Notes
I also noticed that the translations modal doesn't have a title

kelanik8:
Fix select field options translations showing when not saved

image

 - [x] Code builds
 - [x] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/ModuleFields/Configurator/Select.vue
+++ b/client/web/compose/src/components/ModuleFields/Configurator/Select.vue
@@ -30,7 +30,7 @@
               :module="module"
               :highlight-key="`meta.options.${option.value}.text`"
               size="sm"
-              :disabled="isNew || option.new"
+              :disabled="isNew || !defaultFieldOptionsState.includes(option.value)"
               button-variant="light"
             />
             <b-button
@@ -139,6 +139,16 @@ export default {
 
     isNew () {
       return this.module.moduleID === NoID || this.field.fieldID === NoID
+    },
+
+    defaultFieldOptionsState () {
+      const field = this.moduleDefaultState.fields.find(({ name }) => name === this.f.name)
+
+      if (field) {
+        return field.options.options.map(({ value }) => value)
+      }
+
+      return []
     },
   },
 
--- a/client/web/compose/src/views/Admin/Modules/Edit.vue
+++ b/client/web/compose/src/views/Admin/Modules/Edit.vue
@@ -414,6 +414,7 @@
           :field.sync="updateField"
           :namespace="namespace"
           :module="module"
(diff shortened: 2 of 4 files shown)


client/web/compose/src/views/Admin/Modules/Edit.vue
```diff
@@ -703,6 +703,16 @@ export default {
 
     handleFieldEdit (field) {
       this.updateField = compose.ModuleFieldMaker({ ...field })
+
+      if (this.updateField.kind === 'Select') {
```
Fajfa, line 707: This is a very "bandage" kind of solution. 
I suggest you take a look at `lib/js/select.ts`, and see if you can persist the new property there so that compose.ModuleFieldMaker wont "remove" it

Fajfa: This is a very "bandage" kind of solution. 
I suggest you take a look at `lib/js/select.ts`, and see if you can persist the new property there so that compose.ModuleFieldMaker won't "remove" it

Only select.ts requires changes

Fajfa (changes requested):

Fajfa (approved):

client/web/compose/src/components/ModuleFields/Configurator/Select.vue
```diff
@@ -140,6 +140,16 @@ export default {
     isNew () {
       return this.module.moduleID === NoID || this.field.fieldID === NoID
     },
+
+    defaultFieldOptions () {
```
Fajfa: Since these are not really default field options but just the default state, can you please rename to defaultFieldOptionsState

client/web/compose/src/components/ModuleFields/Configurator/Select.vue
```diff
@@ -140,6 +140,16 @@ export default {
...
+      const field = this.moduleDefaultState.fields.find(({ fieldID }) => fieldID === this.f.fieldID)
```
Fajfa: For all new (not yet saved fields) the fieldID will be the same ('0').
So you'll need to make sure you're not returning the wrong options here.

Fajfa (changes requested):

Fajfa (approved):

PR #847, opened 2023-02-07
Fix date record list filter

### LINKED ISSUE #797: Date and Time record field issues
Tests done on QC `2022.9.6-rc.1`

- [x] 1. When you have date only field, and you set the filter to be equal to the correct day, it is not showing any record. (it only shown for not equal and greater then)

- [x] 2. When you have time only and past values configured, it looks like the configuration is only taking the hour in its configuration, because even if you set minutes in the past it will say invalid form as long as you don't also change the hour too. (not sure if this is the expected behavior). Same goes for future time only, if you set minutes in the past but the hour is still the same, the time will work.

- [x] 3. When you have time only value, and you try to filter with a time filter, this is the error popping out `Could not load record list: pq: cannot cast type time with time zone to timestamp with time zone`

kelanik8:
Fix date record list filter

 - [x] Code builds
 - [x] All existing tests pass

### DIFF
--- a/client/web/compose/src/lib/record-filter.js
+++ b/client/web/compose/src/lib/record-filter.js
@@ -74,17 +74,20 @@ export function getFieldFilter (name, kind, query = '', operator = '=') {
 
   if (['DateTime'].includes(kind)) {
     // Build different querries if date, time or datetime
-    const date = moment(query, ['YYYY-MM-DDTHH:mm:ssZ', 'YYYY-MM-DD'])
-    const time = moment(query, ['HH:mm'])
+    const dateTime = moment(query, 'YYYY-MM-DDTHH:mm:ssZ', true)
+    const date = moment(query, 'YYYY-MM-DD', true)
+    const time = moment(query, 'HH:mm', true)
 
     // @note tweaking the template a bit:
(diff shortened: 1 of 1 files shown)


client/web/compose/src/lib/record-filter.js
```diff
@@ -74,17 +74,22 @@ export function getFieldFilter (name, kind, query = '', operator = '=') {
...
+    const time = moment(query, 'HH:mm', true)
 
     // @note tweaking the template a bit:
     // * adding %f to include fractions; mysql sometimes forces them when formatting date
     // * changing Z to +00:00
     // * doing the same for time-only fields
-    if (date.isValid()) {
-      return `TIMESTAMP(DATE_FORMAT(${name}, '%Y-%m-%dT%H:%i:00.%f+00:00')) ${operator} TIMESTAMP(DATE_FORMAT('${date.format()}', '%Y-%m-%dT%H:%i:00.%f+00:00'))`
+    if (dateTime.isValid()) {
+      const left = `TIMESTAMP(DATE_FORMAT(${name}, '%Y-%m-%dT%H:%i:00.%f+00:00'))`;
```
Fajfa: No need to define the vars, just pass it straight to build

Fajfa (approved):

PR #865, opened 2023-02-10
Fix Post-processing function not working for gauge chart

### LINKED ISSUE #822: Post processing function not working for gauge chart
## Steps to reproduce
1. Create a gauge chart or use an existing
2. Add `n/2` to the 'post processing function' field (n is the current value in the chart, not the chart legend!)
3. See that the values in the chart don't change

## Expected behaviour
The added expression should affect the value in the chart. If you go to a generic chart and add the same expression, the values will change.

yonkov:
image

### What was fixed?
Post processing function in Gauge chart in Compose did not work as expected. The end result of the Gauge chart was not updated based on the post processing function field. In contrast. this works for standard charts.

### How was fixed?
Post processing method was added to Gauge Chart compose type in `lib/corteza-js`. 

### Why was fixed?
Users should be able to do basic computations and write expressions based on the result of the chart, in the same way as in standard charts.

### DIFF
--- a/lib/js/src/compose/types/chart/base.ts
+++ b/lib/js/src/compose/types/chart/base.ts
@@ -6,6 +6,7 @@ import {
   Report,
   dimensionFunctions,
   makeAlias,
+  TemporalDataPoint,
 } from './util'
 
 import {
@@ -16,6 +17,10 @@ import {
 } from '../../../cast'
 
 export type PartialChart = Partial<BaseChart>
+
+// The default dataset post processing function to use.
+// This one simply returns the current value.
+const defaultFx = 'n'
 
 /**
  * BaseChart represents a structure that stores any configuration data.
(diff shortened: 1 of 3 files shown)


lib/js/src/compose/types/chart/gauge.ts
```diff
@@ -36,13 +41,38 @@ export default class GaugeChart extends BaseChart {
...
+    }
+    else {
+      data = fx(data)
+    }
+
+    return data
+  }
+
   makeDataset (m: Metric, d: Dimension, data: Array<number|any>, alias: string) {
     const steps = (d.meta?.steps || [])
```
Fajfa, line 42: Instead of defining the function here i would move the one from chart.ts to base.ts and provide it with data parameter(same as it is used in chart.ts).
Also look at where the datasetPostProc is called in base chart,(start of makeDataset) since you did it after reduce here. The function will receive the data which is an array in all cases.

Fajfa (changes requested):

lib/js/src/compose/types/chart/gauge.ts
```diff
@@ -36,10 +41,12 @@ export default class GaugeChart extends BaseChart {
     return (d.meta?.steps || []).map(({ label }: any) => label)
   }
 
-  makeDataset (m: Metric, d: Dimension, data: Array<number|any>, alias: string) {
+  makeDataset (m: Metric, d: Dimension, data: Array<number|TemporalDataPoint>, alias: string) {
     const steps = (d.meta?.steps || [])
 
-    const value = data.reduce((acc, cur) => {
+     const initialValue = this.datasetPostProc(data, m)
```
Fajfa: just do data = this.datasetPostProc(data, m), so we keep it the same as chart

lib/js/src/compose/types/chart/gauge.ts
```diff
@@ -4,9 +4,14 @@ import {
   Report,
   Dimension,
   ChartType,
+  TemporalDataPoint,
 } from './util'
 import { getColorschemeColors } from '../../../shared'
 
+// The default dataset post processing function to use.
+// This one simply returns the current value.
+const defaultFx = 'n'
```
Fajfa: We dont need this section right?

lib/js/src/compose/types/chart/base.ts
```diff
@@ -41,6 +46,59 @@ export class BaseChart {
...
+   * It works with a simple equation written in javascript (example: n + m).
+   * Available variables to use:
+   * * n - current value
+   * * m - previous value (undefined in case of the first element)
+   * * r - entire data array.
+   *
+   * @param data Array of values in the given data set
+   * @param m Metric for the given dataset
+   */
+  public datasetPostProc (data: Array<number|TemporalDataPoint>, m: Metric): Array<number|TemporalDataPoint> {
```
Fajfa: No need to make it public

Fajfa (changes requested):

Fajfa (approved):

PR #885, opened 2023-02-21
Add Record Page Links to Reminders

### LINKED ISSUE #852: Cannot use links in reminders
## Description
There is no way to use the link of the reminder - when you create a reminder there’s an option to link it to a record, but that’s not exposed later on

## Ref to forum
https://forum.cortezaproject.org/t/how-can-i-modify-the-reminder-time-dropdown/1253/3?u=katrinyordanova

yonkov:
image

Summary: This PR improves the way how linked reminders function in Corteza. When there is a linked reminder to a specific record, you can now go to the Record Page from the reminder directly. Before that, there was a link on the record page to the linked reminder but not vice vicera.

### What was added?
When there is a linked reminder, a link to the linked Record is added in the reminder. 

### How was added?
In Compose, the edit reminder tempate was updated at `src/components/namespaces/reminders/edit` to add button input group as well the list template at `src/components/namespaces/reminders/edit`.

### Why was added?
Before this feature, when the users link a record page to a reminder, the link was shown in the record page but when you visit the reminder, there was no way to navigate quickly to the record. This functionality improves UX, since users can now easily navigate from the linked reminder to the linked record page directly.

### DIFF
--- a/client/web/compose/src/components/Namespaces/Reminders/Edit.vue
+++ b/client/web/compose/src/components/Namespaces/Reminders/Edit.vue
@@ -50,10 +50,24 @@
           v-if="reminder.payload.link"
           :label="$t('reminder.routesTo')"
         >
-          <b-form-input
-            v-model="reminder.payload.link.label"
-            data-test-id="input-link"
-          />
+          <b-input-group>
+            <b-form-input
+              v-model="reminder.payload.link.label"
+              data-test-id="input-link"
+            />
+
+            <b-input-group-append>
+              <b-button
+                :disabled="!recordViewer"
+                :to="recordViewer(reminder.payload.link.params)"
+                :title="$t('label.recordPageLink')"
+                variant="light"
+                class="d-flex align-items-center"
+              >
+                <font-awesome-icon :icon="['fas', 'external-link-alt']" />
+              </b-button>
+            </b-input-group-append>
+          </b-input-group>
         </b-form-group>
       </b-form>
     </b-list-group-item>
@@ -177,6 +191,10 @@ export default {
   },
 
   methods: {
+    recordViewer (params) {
+      return params ? { name: 'page.record', params } : undefined
+    },
+
     save () {
       // @todo support for updating times
       let r = {}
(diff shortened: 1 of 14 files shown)


client/web/compose/src/components/Namespaces/Reminders/Edit.vue
```diff
@@ -50,10 +50,23 @@
...
-          />
+          <b-input-group>
+            <b-form-input
+              v-model="reminder.payload.link.label"
+              data-test-id="input-link"
+            />
+
+            <b-input-group-append>
+              <b-button
+                :to="recordViewer"
```
Fajfa: Would probably be a good idea to disable the button if link is not valid

client/web/compose/src/components/Namespaces/Reminders/List.vue
```diff
@@ -54,6 +54,16 @@
           class="ml-1"
         />
 
+        <b-button
+          v-if="r.payload.link"
+          :to="recordViewer(r.payload.link.params)"
```
Fajfa: Same here (disable)

client/web/compose/src/components/Namespaces/Reminders/List.vue
```diff
@@ -127,6 +137,9 @@ export default {
     makeTooltip ({ remindAt }) {
       return fmt.fullDateTime(remindAt)
     },
+    recordViewer (link) {
```
Fajfa: I suggest you just call the param "params", so you don't have to do `{ params: link }`, you can just do `{ params }`

locale/en/corteza-webapp-compose/general.yaml
```diff
@@ -55,6 +55,7 @@ label:
   processing: Processing...
   record: Record
   recordPage: Record page
+  recordPageLink: Go to Record page
```
Fajfa: Probably best to name it view record, since you're not just going to record page you're actually viewing record here

Fajfa (changes requested):

PR #887, opened 2023-02-23
2023.3.x feature users manage profile photo

### LINKED ISSUE #483: Allow users to manage their profile photo [PlanID: 2023.3/27/1.5]
Administrators can enable (default) or disable if users can set their own profile photo in the auth/profile pages. The profile photo is either an uploaded photo or custom initials (1-3 ASCII characters) with a custom background and text colour. Administrators can also upload and set the initials with colours for any user via the admin panel.

The profile photo is then used on the top navigation on the right side.

-----

 - [x] for image - we accept JPEG and PNG formats, max size (propose)MB, resize to (propose)x(propose)px dimensions.
 - [x] admin can enable/disable photo upload (enabled by default)
 - [x] admin can enable/disable "the initials" feature (enabled by default)
 - [x] create a special endpoint to serve profile photos for users, accepts user-id and dimensions (resized on-demand)
 - [ ] in-browser resizing and cropping is a bonus.

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/483

This pull request introduces several changes that take effect when the "Profile Photo Avatar" feature is enabled. These changes include:

- [x]  Ability for admins to enable/disable avatar upload and avatar generation, which is enabled by default.
- [x]  Ability for users to upload their own avatars.
- [x]  Generation of avatar initials using default font path, avatar initial background color, and avatar initial text color, if environment variables are not configured by the admin.
- [x]  Ability for users to set their own avatar's initial background color and avatar's initial text color.
- [x]  Generation of avatar initials using the user's name, handle, or email, in order of priority.
- [x]  Generation of avatar initials every time there is an update on the initial, initial text color, or the background color.
- [x]  Uploaded avatar is given priority over generated avatar on display.
- [x]  Ability for users to update their avatar's initial text color and background color when an uploaded avatar is in place.
- [x]  Ability for users to delete their uploaded avatar.
- [x]  Ability for admins to upload and set the avatar initial background and text colors for any user via the admin panel.
- [x]  Ability for admins to delete an avatar for any user.
- [x]  Usage of the profile avatar on the top navigation on the right side.
 
 The accepted avatar image format is only JPEG and PNG formats with a maximum size is **1 MB** and dimensions are  (300x300)px.

The custom initials consist of up to 3 letter characters.

These changes provide users with greater flexibility and customization options for their profile avatars, while also allowing admins to manage avatars on behalf of users.

## Avatar Generation Rules

1.  If the user's first name is available, the initials will be the first two letters of the name (e.g., `John => JO`).
2. If the user's full name is available, the initials will be the first character of the first and last names (e.g., `John Doe => JD`).
3. If the user has three or more names, the first three characters of the first, second, and third names are taken to form the initials (e.g., `John Doe Test => JDT`, `John Doe User Test => JDU`).
4. If the user's handle exists and their name doesn't, the initials will be the first letter of the handle (e.g., `john => J`).If the handle contains delimiters such as `., _, or -`, the handle name is split using those delimiters, and the first letter from the first and second parts is used as the initials (e.g., `john.doe => JD`, `john-doe.me => JD`, `john_doe.me-you => JD`).
5. If both the name and handle don't exist, the initials will be the first letter of the email (e.g., `john.doe@tld.dev => J`). If the email name contains delimiters before the @ symbol, the email name is split using those delimiters, and the first letter from the first and second parts is used as the initials (e.g., `john.doe@tld.dev => JD`, `john-doe.me@tld.dev => JD`, `john_doe.me-you@tld.dev => JD`).

## Environment Variable

1. `AVATAR_INITIALS_FONT_PATH`: Avatar initials font file path: default value is `./auth/assets/public/fonts/poppins/Poppins-Regular.ttf`
2. `AVATAR_INITIALS_BACKGROUND_COLOR`: Avatar initials background color. default value is `#F3F3F3`
3. `AVATAR_INITIALS_COLOR`: Avatar initials text color. default value is `#162425`
4. `ATTACHMENT_AVATAR_MAX_FILE_SIZE`: Avatar image maximum upload size. default value is `1MB`

 - [x] Commits are tidied up, squashed if needed, and follow guidelines in CONTRIBUTING.md
 - [x] Code builds
 - [x] All existing tests pass


server/auth/assets/templates/profile.html.tpl
```diff
@@ -18,6 +20,66 @@
...
+        {{ end }}
+
+
+        {{ if .initialsEnabled }}
+        <div class="mb-3">
+            <label for="avatarInitials">{{ tr "profile.template.form.avatar-initial.label" }}</label>
+            <div class="d-block">
+                <div class="d-flex justify-content-start align-items-center">
+                    <div>
+                        <input class="col-10 form-control" type="text" id="avatarInitials" name="avatar-initials" value="{{ .form.avatarInitial }}">
```
tjerman: Should we take the initials from the user's name instead?

server/auth/handlers/handle_oauth2.go
```diff
@@ -437,6 +439,14 @@ func (h AuthHandlers) handleTokenRequest(req *request.AuthReq, client *types.Aut
 
 	response := h.OAuth2.GetTokenData(ti)
 
+	// include user's avatar info
+	response["profile"] = map[string]string{
+		"avatarID":                strconv.FormatUint(user.Meta.AvatarID, 10),
```
tjerman: What is the avatarID set to when using initials?

server/system/rest/user.go
```diff
@@ -146,19 +146,42 @@ func (ctrl User) Create(ctx context.Context, r *request.UserCreate) (interface{}
...
+				return nil, service.UserErrInvalidInitialsCharacter()
+			}
+		}
+
+		// Check for avatar initials character length
+		if len(r.Meta.AvatarInitials) > 3 {
+			return nil, service.UserErrInvalidInitialsLength()
+		}
+	}
+
```
tjerman: Brave.
Why these limitations and the checks being done the way they are?

server/system/service/attachment.go
```diff
@@ -198,8 +200,49 @@ func (svc attachment) CreateApplicationAttachment(ctx context.Context, name stri
...
+	)
+
+	// check the file type
+	ext := filepath.Ext(name)
+	if ext != ".jpg" && ext != ".jpeg" && ext != ".png" {
+		return nil, AttachmentErrInvalidAvatarFileType()
+	}
+
+	//check the file size
+	if size > 3000000 {
```
tjerman: Try to avoid magic numbers.
Should we keep file size limitations consistent and have it configurable?

server/system/service/attachment.go
```diff
@@ -198,8 +200,49 @@ func (svc attachment) CreateApplicationAttachment(ctx context.Context, name stri
...
 
+func (svc attachment) CreateAuthAttachment(ctx context.Context, name string, size int64, fh io.ReadSeeker, labels map[string]string) (att *types.Attachment, err error) {
+	var (
+		aaProps       = &attachmentActionProps{}
+		currentUserID = intAuth.GetIdentityFromContext(ctx).Identity()
+	)
+
+	// check the file type
+	ext := filepath.Ext(name)
+	if ext != ".jpg" && ext != ".jpeg" && ext != ".png" {
```
tjerman, line 220: Should we allow gifs as well?

server/system/service/user.go
```diff
@@ -1035,3 +1042,86 @@ func toLabeledUsers(set []*types.User) []label.LabeledResource {
...
+				return
+			}
+		}
+
+		att, err := svc.att.CreateAuthAttachment(
+			ctx,
+			upload.Filename,
+			upload.Size,
+			file,
+			map[string]string{"key": "profile-photo-avatar"},
```
tjerman: Should we rather utilize the attachment type?

tjerman (changes requested):
I only checked the code under `/server`.

Are we sure we want to handle initials like so? I imagined we'd make a `.png` like here: https://ui-avatars.com/ (I'm not saying we need to use this service).
How would the initials be presented in the OAuth flows when authenticating other applications -- they won't know how to show initials.

server/auth/assets/templates/profile.html.tpl
KinyaElGrande: Yes, I believe we can do this:

1. When the handle exists and the name doesn't we can use the first letter of the handle for example `testuser` would be `T`.
2. When the name exists, we can use the first two letters if the user provided only the first name for example 'Francis' would be `FR`, or the first characters for the first and last name when the user provides their full name, for example `Francis Mumbi` would be `FM`.
3. Users can set their own initials.

server/system/service/attachment.go
KinyaElGrande: I think we shouldn't allow gifs as profile avatars since they could be distracting and in cases where Corteza is used as a professional tool, I think it may be inappropriate.

server/system/types/user.go
```diff
@@ -37,8 +37,16 @@ type (
 	}
 
 	UserMeta struct {
-		// @todo remove, obsolete.
-		Avatar string `json:"avatar,omitempty"`
+		// User's profile avatar photo attachment ID
+		AvatarID uint64 `json:"avatarID,string"`
+
+		// User's profile avatar initials attachment ID
+		AvatarInitialsID uint64 `json:"AvatarInitialsID,string"`
```
tjerman: Don't you think this is just going to complicate things? Why not have it under AvatarID?
If you really want to differentiate between them, you've added a special kind for generated avatars already.

server/system/types/user.go
```diff
@@ -37,8 +37,16 @@ type (
...
+		// User's profile avatar custom initials
+		AvatarInitials string `json:"avatarInitials,omitempty"`
+		// User's profile Initial background and text color
+		AvatarInitialsTextColor string `json:"avatarInitialsTextColor,omitempty"`
+		AvatarInitialsBgColor   string `json:"avatarInitialsBgColor,omitempty"`
```
tjerman: Do we need to remember all of this?
I wouldn't save these. If you really want this, you can keep it or you can utilize .png metadata

tjerman (changes requested):

server/auth/handlers/handle_profile.go
```diff
@@ -48,10 +46,9 @@ func (h *AuthHandlers) profileForm(req *request.AuthReq) (err error) {
 			"name":              u.Name,
 			"preferredLanguage": preferredLanguage,
 			"avatarUrl":         avatarUrl,
-			"avatarInitialsUrl": avatarInitialsUrl,
-			"avatarInitial":     u.Meta.AvatarInitials,
-			"initialTextColor":  u.Meta.AvatarInitialsTextColor,
-			"initialBgColor":    u.Meta.AvatarInitialsBgColor,
+			//"avatarInitial":     u.Meta.AvatarInitials,
+			//"initialTextColor":  u.Meta.AvatarInitialsTextColor,
+			//"initialBgColor":    u.Meta.AvatarInitialsBgColor,
```
tjerman: Should these be removed instead?

server/system/service/user.go
```diff
@@ -1139,3 +1150,23 @@ func (svc user) DeleteAvatar(ctx context.Context, userID uint64) (err error) {
 
 	return svc.recordAction(ctx, uaProps, UserActionDeleteAvatar, err)
 }
+
+func processAvatarInitials(u *types.User) (initials string) {
```
tjerman: You've mentioned you'd use the email also; that no longer the case?

server/system/types/user.go
```diff
@@ -30,6 +30,8 @@ type (
 		SuspendedAt *time.Time `json:"suspendedAt,omitempty"`
 		DeletedAt   *time.Time `json:"deletedAt,omitempty"`
 
+		AvatarInitialsMeta *UserAvatarInitialsMeta `json:"avatarInitialsMeta"`
```
tjerman: Are we storing this both in the PNG and the DB? If so, should we?

tjerman (changes requested):

server/system/types/user.go
KinyaElGrande: We should store the fields `[initials, bgcolor, textcolor]` on the database to maintain form interactivity on the front end.

server/system/service/user.go
KinyaElGrande: That is no longer the case. When a user is created, their handle is generated, and since a handle has a higher priority in this case, I opted to remove the user's email from the ` processAvatarInitials` function.

server/.env.example
```diff
@@ -1290,3 +1290,32 @@
...
 # Default: <no value>
 # DISCOVERY_BASE_URL=<no value>
+
+###############################################################################
+###############################################################################
+# attachment
+#
+
+###############################################################################
+# Avatar image maximum upload size, default value is 3MB
```
petergrlica: That's a lot for an avatar image, but ok :)

server/auth/handlers/handle_profile.go
```diff
@@ -25,6 +32,10 @@ func (h *AuthHandlers) profileForm(req *request.AuthReq) error {
 		}
 	}
 
+	if u.Meta.AvatarID != 0 {
+		avatarUrl = fmt.Sprintf("/api/system/attachment/avatar/%d/original/%s", u.Meta.AvatarID, types.AttachmentKindAvatar)
```
petergrlica: The `/api` prefix can in some configurations be omitted or totally different, see more: https://github.com/cortezaproject/corteza/blob/2023.3.x/server/app/servers.go#L71

server/system/service/user.go
```diff
@@ -1035,3 +1051,119 @@ func toLabeledUsers(set []*types.User) []label.LabeledResource {
...
+				upload.Size,
+				file,
+				map[string]string{"key": types.AttachmentKindAvatar},
+			)
+
+			if err != nil {
+				return err
+			}
+		} else {
+			initial := processAvatarInitials(u)
```
petergrlica: Make sure to handle sensible defaults for people with initials above char 127, like mr. Mesut Özil (for example).
What happens with him, does `CreateAvatarInitialsAttachment` throw an error?

locale/en/corteza-webapp-admin/system.users.yaml
```diff
@@ -20,6 +20,17 @@ editor:
...
+  avatar:
+      title: Profile avatar
+      preview:
+        label: Avatar
+        title: Avatar
+      initial:
+        color: Initial Color
+        backgroundColor: Background color
+      uploader:
+        instructions: 'Click or drop user avatar here to upload'
```
petergrlica: You can drop the `'` like you did with the rest of the strings

server/system/service/user.go
KinyaElGrande: The new initials don't check whether the characters are of ASCII type. Instead, they check whether the character is a letter. If it is not a letter, the default initial is set to the letter 'CU' for Corteza User.

PR #913, opened 2023-03-03
Fix reports link step sort for numeric 

### LINKED ISSUE #761: Sorting in reports with link steps doesn't work as expected
When sorting a table which is showing a link step, the sort seems to order numbers based on text sorting rules, not on number sorting rules.

See the column `bb n1`; the values are numeric but the descending order was done using character order.

Image

KinyaElGrande:
Fixes https://github.com/cortezaproject/corteza/issues/761
 
**Issue**: When sorting a table showing a link step, the sort seems to order numbers based on text sorting rules, not number sorting rules.

### DIFF
--- a/server/pkg/dal/exec_link_left.go
+++ b/server/pkg/dal/exec_link_left.go
@@ -3,10 +3,10 @@ package dal
 import (
 	"context"
 	"fmt"
+	"github.com/cortezaproject/corteza/server/pkg/filter"
 	"io"
 	"sort"
-
-	"github.com/cortezaproject/corteza/server/pkg/filter"
+	"strconv"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


petergrlica (approved):
lgtm

manuelnelson:
@KinyaElGrande This still seems to be happening?  I've confirmed that the "Days Since Needs Assessment" is a Number field, so I'm not sure why it's sorting like the image below? Let me know if there's an extra step I need to take to get the sorting to work on the number field.

image

KinyaElGrande:
@manuelnelson  Thank you for reporting this issue.
I'll work on reproducing it, then formally open an issue for it after I confirm the behaviour.

Could you please provide the version of Corteza you're running on.

KinyaElGrande:
@manuelnelson,
I have tried reproducing this issue on Corteza `2024.9.3` and everything appears to be working as expected as shown in the GIF below.

Could you please double-check your Corteza version? It might be older than the version where this was fixed.

ScreenRecording2025-06-20at12 51 53-ezgif com-video-to-gif-converter

KinyaElGrande:
I have tried reproducing this issue on Corteza `2024.9.3 ` and everything appears to be working as expected, as shown in the GIF below.
Could you please double-check your Corteza version? It might be older than the version where this was fixed.

ScreenRecording2025-06-20at14 08 08-ezgif com-video-to-gif-converter

PR #928, opened 2023-03-13
Use Magnifying Glass in Page Builder

### LINKED ISSUE #816: Can't use magnifying glass in page builder
## Steps to reproduce
1. Add a PB that has the magnification option
2. Enable it
3. Go to page builder mode
4. Click on the magnification icon
5. Nothing happens

## Expected behaviour
Make blocks magnifiable in the builder

yonkov:
image

### What was improved?
The block magnification icon was not functional in the page builder, only on public pages. It was decided that this should not be the case, the icon should work from everywhere, all the time.

### How was improved?
Inside the compose, the page block modal was appended to the the page builder component.

### Why was improved?
It was impractical to have a magnification icon that does not magnify. Adding it improved the usability of the feature and its adoption throughout Corteza.

### DIFF
--- a/client/web/compose/src/components/Public/Page/Block/Modal.vue
+++ b/client/web/compose/src/components/Public/Page/Block/Modal.vue
@@ -99,18 +99,20 @@ export default {
   },
 
   created () {
-    this.$root.$on('magnify-page-block', ({ blockID, block } = {}) => {
-      this.customBlock = block
-      blockID = blockID || (block || {}).blockID
-      this.$router.push({ query: { ...this.$route.query, blockID } })
-    })
+    this.$root.$on('magnify-page-block', this.magnifyPageBlock)
   },
 
-  beforeDestroy () {
-    this.$root.$off('magnify-page-block')
+  destroyed () {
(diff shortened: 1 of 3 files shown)


client/web/compose/src/components/Public/Page/Block/Modal.vue
```diff
@@ -99,18 +99,19 @@ export default {
...
+  destroyed () {
+    this.$root.$off('magnify-page-block', this.magnifyPageBlock)
   },
 
   methods: {
+    magnifyPageBlock ({ blockID, block } = {}) {
+      this.customBlock = block
+      blockID = blockID || (block || {}).blockID
+      this.$router.push({ query: { ...this.$route.query, blockID } })
+    },
```
Fajfa, line 114: Empty line below please

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -262,6 +262,9 @@
         </template>
       </editor-toolbar>
     </portal>
```
Fajfa, line 264: Empty line below please

Fajfa (changes requested):

PR #950, opened 2023-03-16
2023.3.x feature recordlistconfigurator layout

### LINKED ISSUE #918: RecordListConfigurator Layout Change
Update the styles and layout of the recordlist configurator's interface according to internally discussed designs.

EmmyMay:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass


lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+    <div
+      class="d-flex"
```
Fajfa, line 3: align-items-center, also the gap is not needed just do mx on checkbox

lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
...
+      style="gap: 1.5rem;"
+    >
+      <h6
```
Fajfa: use spans instead, font-weight-bold and text-primary when value

lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
...
+        class="mb-0 text-primary"
+        :class="{ 'text-muted': value }"
+      >
+        No
+      </h6>
+      <b-form-checkbox
+        v-bind="$attrs"
+        v-model="value"
```
Fajfa: you shouldn't use v-model here since it updates the prop, this is a display only component that means you only emit events when you update and don't v-model

lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
...
+        switch
+        @input="handleInput"
+      />
+      <h6
+        class="mb-0 strong text-primary"
+        :class="{ 'text-muted': !value }"
+      >
+        Yes
```
Fajfa: Should be translated using the labels prop

lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
...
+        class="mb-0 text-primary"
+        :class="{ 'text-muted': value }"
+      >
+        No
+      </h6>
+      <b-form-checkbox
+        v-bind="$attrs"
+        v-model="value"
+        switch
+        @input="handleInput"
```
Fajfa: @input will trigger on non user interaction aswell, so I'm not sure you want this event

lib/vue/src/components/input/CInputCheckbox.vue
```diff
@@ -0,0 +1,38 @@
...
+
+</template>
+
+<script>
+export default {
+  props: ['value'],
+
+  methods: {
+    handleInput() {
+      this.$emit('input', this.value)
```
Fajfa: formatting, for example spaces and so on between () and function title

Also since the event gets the param that has been updated you should use that instead of this.value

Fajfa (changes requested):

PR #955, opened 2023-03-20
Add geolocation capture feature on record geometry field

### LINKED ISSUE #875: Enable geo location capture from Compose Webapp
Extend the current geo module field to support users capturing their current location, either on demand or as a default value.
This feature is primarily designed to be used with mobile devices as the location is most accurate there.

## What to use

Use the [geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API) provided by the browser.
Check if and how exceptions in regards to the API not supported, GPS not available, user denying location permissions, ...

You are free to use some NPM module but the standard API should be enough.

### Geometry field default current location

In the geometry module field configuration, add a new option for the "default value" checkbox which will take the user's current location and use it to prefill the input.

Image

**Important concern**: When should the prefill value be captured? When opening the record page, before closing it? The concern here is if I'd have to capture locations on my mobile and I'd open the record page before hand, the location I'd want is the one when saving the record, not when opening it. If we want to capture it at the end, what if I was in a hurry and was planning on filling the rest out a bit later?

A potential soliton is to capture the location at the start then before saving, if the location changed from the previous one (we shouldn't use strict equality but rather a delta of x -- the user can walk around a bit but the location is still about the same), ask the user if they want to update the location. Additionally, you can capture some locations every n seconds and provide those as options to chose from.

**Note**: You are free to do some of your own research and propose a different UX for the current location prefill.

### Geometry field insert user location

When editing a record with a geometry field, provide a "use my current location" button which would use the geolocation API to get and update the field's value.

Image

## Steps

- [x] Research and determine how error/exception handling should be handled
- [x] Research how the prefill value option should behave
- [x] Implement the prefill value
- [x] Implement the button to insert the current location

KinyaElGrande:
ref: https://github.com/cortezaproject/corteza/issues/875

This PR adds a geolocation capture feature on the record geometry field while using the [geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API) provided by the browser.

### Features added
- Show the current location button with a `location-arrow` icon.

   Screenshot 2023-03-23 at 11 11 09

   
   Screenshot 2023-03-23 at 11 22 59

- An option to prefill the current location and to hide the `current location` button.
   
   Screenshot 2023-03-22 at 10 39 06

- Ability to fly to current location while on the map.

When the `Show Current Location` button is clicked or when the `Prefill With Current Location`  option is enabled, both the latitude and longitude input fields are prefilled with the user's current coordinates obtained from the [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API).

### Errors captured
1.  [Geolocation API permission denied error](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError/code)
2. [Geolocation API position unavailable error](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError/code)
3. [Geolocation API timeout error](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError/code)
4. Geolocation API unknown error
5. Geolocation API not supported error
6. An error occurred while getting user's location

### DIFF
--- a/client/web/compose/src/components/ModuleFields/Configurator/Geometry.vue
+++ b/client/web/compose/src/components/ModuleFields/Configurator/Geometry.vue
@@ -1,21 +1,44 @@
 <template>
   <div>
+    <b-form-checkbox v-model="f.options.prefillWithCurrentLocation">
+      {{ $t('prefillWithCurrentLocation') }}
+    </b-form-checkbox>
+
+    <b-form-checkbox v-model="f.options.hideCurrentLocationButton">
+      {{ $t('hideCurrentLocationButton') }}
+    </b-form-checkbox>
+
     <b-form-group
-      :label="$t('kind.geometry.initialZoomAndPosition')"
+      :label="$t('initialZoomAndPosition')"
       class="mb-0"
     >
       <l-map
+        ref="map"
         :zoom="zoom"
         :center="center"
         class="w-100"
         style="height: 60vh;"
         @update:zoom="f.options.zoom = $event"
         @update:center="f.options.center = $event"
+        @locationfound="onLocationFound"
       >
         <l-tile-layer
           url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
           :attribution="attribution"
         />
+        <l-control class="leaflet-bar">
+          <a
+            :title="$t('tooltip.goToCurrentLocation')"
+            role="button"
+            class="d-flex justify-content-center align-items-center"
+            @click="goToCurrentLocation"
+          >
+            <font-awesome-icon
+              :icon="['fas', 'location-arrow']"
+              class="text-primary"
+            />
+          </a>
+        </l-control>
       </l-map>
     </b-form-group>
   </div>
@@ -23,10 +46,16 @@
 
 <script>
 import base from './base'
+import { LControl } from 'vue2-leaflet'
 
 export default {
   i18nOptions: {
     namespaces: 'field',
+    keyPrefix: 'kind.geometry',
+  },
+
+  components: {
+    LControl,
   },
 
   extends: base,
@@ -44,6 +73,17 @@ export default {
 
     zoom () {
       return this.f.options.zoom || 3
+    },
+  },
+
+  methods: {
+    goToCurrentLocation () {
+      this.$refs.map.mapObject.locate()
+    },
+
+    onLocationFound ({ latitude, longitude }) {
+      const zoom = this.$refs.map.mapObject._zoom >= 13 ? this.$refs.map.mapObject._zoom : 13
+      this.$refs.map.mapObject.flyTo([latitude, longitude], zoom)
     },
   },
 }
--- a/client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
+++ b/client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
@@ -33,13 +33,14 @@
     <div class="d-flex w-100">
       <b-button
         v-if="field.isMulti"
-        variant="primary"
-        rounded
+        :title="$t('tooltip.openMap')"
+        variant="light"
         class="w-100"
         @click="openMap"
       >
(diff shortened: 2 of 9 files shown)


lib/js/src/compose/types/module-field/geometry.ts
```diff
@@ -7,13 +7,15 @@ interface GeometryOptions extends Options {
   center: number[];
   zoom: number;
   multiDelimiter: string;
+  presetWithCurrentLocation: boolean;
```
Fajfa: Add an option to hide the current location button aswell

locale/en/corteza-webapp-compose/field.yaml
```diff
@@ -143,6 +143,7 @@ kind:
     longitude: Longitude
     clickToPlaceMarker: Click to place a marker or click on a marker to remove it
     initialZoomAndPosition: Set initial map zoom and position
+    presetWithCurrentLocation: Preset with current location
```
Fajfa: Can you rename this one and the user one to prefill

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -85,12 +85,23 @@
...
           >
             <font-awesome-icon
               :icon="['fas', 'map-marked-alt']"
             />
           </b-button>
+
+          <b-button
+            variant="light"
+            rounded
+            @click="currentLocation"
```
Fajfa: You also need to add this to the <multi> section above, since otherwise it wont work with multivalue geometry(you'll need to use index)

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -229,6 +244,43 @@ export default {
         this.localValue = { coordinates: [] }
       }
     },
+
+    currentLocation () {
+      try {
+        if (navigator.geolocation) {
+          navigator.geolocation.getCurrentPosition(
+            position => {
+              this.localValue = {
```
Fajfa: For multi value geometry this wont work

client/web/compose/src/components/faIcons.js
```diff
@@ -62,6 +62,7 @@ import {
   faSync,
   faExclamationTriangle,
   faEllipsisV,
+  faLocationArrow,
```
Fajfa, line 65: Since we dont have the location icon in free version, use the compass icon(fa-compass)

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -85,12 +85,23 @@
           <b-button
             variant="light"
             rounded
+            class="mr-2"
```
Fajfa: Remove mr-2 please

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -229,6 +244,43 @@ export default {
...
+                case error.TIMEOUT:
+                  this.toastErrorHandler(this.$t('notification:field-geometry.geolocationErrors.timeout'))()
+                  break
+                default:
+                  this.toastErrorHandler(this.$t('notification:field-geometry.geolocationErrors.unknownError'))()
+                  break
+              }
+            },
+          )
+        } else {
```
Fajfa: Move this to the top and to !navigator.geolocation instead and return inside it, so its less nesting

client/web/compose/src/components/ModuleFields/Configurator/Geometry.vue
```diff
@@ -1,5 +1,11 @@
 <template>
   <div>
+    <b-form-group>
```
Fajfa: No need for a form group on checkboxes

Fajfa (changes requested):

PR #990, opened 2023-03-27
Embed avatar initials font path and relocate it

### LINKED ISSUE #983: Error:  font file not found, please ensure that the correct AVATAR_INITIALS_FONT_PATH is set
## Issue

Image

## When does the issue occur
- after enabling profile settings in system settings and trying to add/edit a user

## TODO
- [x] fix bug
- [x] improve error message

KinyaElGrande:
Fixes: https://github.com/cortezaproject/corteza/issues/983 bug which occurs on production.

Screenshot 2023-03-23 at 15 44 32

### Solution approach:
Replaced the current method of loading the font file from the disk with embedding it in the binary this will ensure that it's always available to the application and can be loaded directly from memory.


tjerman (approved):
lgtm :+1:

tjerman (approved):
lgtm :+1:

tjerman (commented):
lgtm :+1:

PR #1018, opened 2023-03-29
Update field kind after it's selected

### LINKED ISSUE #1002: New field configs not taken into account
## Description of the bug

_When adding a new field with a different kind and if you don't open its configuration, the configuration is not updated, it is just inherited from the default kind/ from the previous field kind._

## Reproduction steps

1. Create a number field but don't open its config
2. save the field and go in all records
3. add a float number for example 1.21
4. Notice that the number is rounded 
5. Now go in modules again and open the number field config and just open the number tab and don't touch anything just press save 
6. Now go in all records and add the same entry, 1.21, notice that now the number is not rounded.

## Expected behavior 

_When you add a new field, the configurations of that field kind should be automatically applied to the new field._

## Screenshot/Video

https://user-images.githubusercontent.com/53232926/228258439-c3a7844f-17ba-4cea-b55d-65e905fe139e.webm

## Version

Branch: 2023.3.0

katrinDY:
Field kind is updated after it's selected

None

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/views/Admin/Modules/Edit.vue
+++ b/client/web/compose/src/views/Admin/Modules/Edit.vue
@@ -301,6 +301,7 @@
                           :is-duplicate="!!duplicateFields[index]"
                           @edit="handleFieldEdit(module.fields[index])"
                           @delete="module.fields.splice(index, 1)"
+                          @updateKind="handleFieldKindUpdate(index)"
                         />
                       </draggable>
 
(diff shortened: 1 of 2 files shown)


client/web/compose/src/views/Admin/Modules/Edit.vue
```diff
@@ -301,6 +301,7 @@
                           :is-duplicate="!!duplicateFields[index]"
                           @edit="handleFieldEdit(module.fields[index])"
                           @delete="module.fields.splice(index, 1)"
+                          @updateKind="handleFieldKind(module.fields[index])"
```
Fajfa: update based on index and splice same a handleFieldSave, also rename handleFieldKind to handleFieldKindUpdate

Fajfa (changes requested):

PR #1020, opened 2023-03-29
Fix dedicated record revisions table

### LINKED ISSUE #911: Error on rec update when using revision feature with MySQL Backend
### Description

When using revisions i cannot update record anymore, its rasing error:

“Could not update record: failed to complete transaction: Error 1366: Incorrect decimal value: ‘updated’ for column ‘operation’ at row 1”

Table has been created just fine:

```
CREATE TABLE 'record_revisions_begleiter' (
 'id' bigint unsigned NOT NULL,
  'ts' datetime NOT NULL,
  'revision' decimal(10,0) NOT NULL,
  'operation' decimal(10,0) NOT NULL,
  'rel_resource' bigint unsigned NOT NULL,
  'rel_user' bigint unsigned NOT NULL,
  'delta' json NOT NULL,
  'comment' text NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci

```
record is beeing updated in background, but error is thrown because revision row in DB could not be created, it looks like insert order in code could be wrong. it seems to try to insert string "updated" into operation column which is beeing created as decimal

### Version
2022.9.7

### Steps to reproduce
With Docker installation using MySQL as a backend:
 - Activate revisions in module
 - try to update record in this module

### Expected behaviour

revision insert in MySQL table should work

KinyaElGrande:
Fixes: https://github.com/cortezaproject/corteza/issues/911

### Issue
When a user set's a  custom dedicated record revision table, then tries to update a record, the following error was appearing:  _“Could not update record: failed to complete transaction: Error 1366: Incorrect decimal value: ‘updated’ for column ‘operation’ at row 1”_

### Solution
Rearranged the order of the `rel_resource`, `revision`, and `operation` columns when custom dedicated record revision table is created and changed the `operation` column datatype to `text`.

### DIFF
--- a/server/pkg/revisions/model.go
+++ b/server/pkg/revisions/model.go
@@ -23,9 +23,9 @@ func Model() *dal.Model {
 		Attributes: dal.AttributeSet{
 			&dal.Attribute{Ident: "id", PrimaryKey: true, Store: &dal.CodecPlain{}, Type: &dal.TypeID{}},
 			&dal.Attribute{Ident: "ts", Store: &dal.CodecPlain{}, Type: &dal.TypeTimestamp{}},
-			&dal.Attribute{Ident: "revision", Store: &dal.CodecPlain{}, Type: &dal.TypeNumber{}},
-			&dal.Attribute{Ident: "operation", Store: &dal.CodecPlain{}, Type: &dal.TypeNumber{}},
 			&dal.Attribute{Ident: "rel_resource", Store: &dal.CodecPlain{}, Type: &dal.TypeID{}},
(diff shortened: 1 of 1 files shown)


tjerman:
Why the rearrange?

KinyaElGrande:
@tjerman  In order to match the `compose_record_revisions` table, since the input order for both tables is similar.

tjerman:
I see; nice catch.
LGTM :+1:

tjerman (approved):

PR #1072, opened 2023-04-04
Fix preview for icon attachment

### LINKED ISSUE #904: Uploading .ico file results in not showing the picture on record list
## Description of the bug

_When you have a record upload field in your record list page and you try to upload an icon with .ico configuration/extension, the icon is not shown. The other types like, png, svg are working the only problem looks like it is with .ico._

## Reproduction steps

1. Create a `File upload` field and configure it to be like single image and hide file name from the file upload configuration
2. After that try adding a record with uploading an .ico picture
3. After saving the record, notice that the file upload is blank, if you press on the record it is blank again, and if you press on edit record, then you can see the file uploaded with only the name of it

## Expected behavior 

_You should be able to upload an .ico file._

## Screenshots

Image

Image

Image

## Version

Branch: 2022.9.8-rc.1 (Tested on QC)

vicpatel:
It will fall back to original image meta and return preview accordingly.

Please take a look.


server/compose/service/attachment.go
```diff
@@ -685,6 +692,25 @@ func (svc attachment) processImage(original io.ReadSeeker, att *types.Attachment
...
+func (svc attachment) processIcon(original io.ReadSeeker, att *types.Attachment) (err error) {
+	if _, err = original.Seek(0, 0); err != nil {
+		return
+	}
+
+	if att.Meta.Preview == nil {
+		att.Meta.Preview = &att.Meta.Original
+		att.Meta.Preview.Image = &types.AttachmentImageMeta{
+			Width:    attachmentPreviewMaxWidth,
+			Height:   attachmentPreviewMaxHeight,
```
tjerman: This sounds unsafe to me -- shouldn't we get the dimensions from the icon itself?
If there are issues with that, we could also display icons like we do other non-image files (the little font awesome icon)

tjerman (changes requested):

server/compose/service/attachment.go
```diff
@@ -685,6 +693,65 @@ func (svc attachment) processImage(original io.ReadSeeker, att *types.Attachment
...
+	// if att.Meta.Preview == nil {
+	// 	att.Meta.Preview = &att.Meta.Original
+	// 	att.Meta.Preview.Image = &types.AttachmentImageMeta{
+	// 		Width:    attachmentPreviewMaxWidth,
+	// 		Height:   attachmentPreviewMaxHeight,
+	// 		Animated: false,
+	// 	}
+	// }
+
+	// return
```
tjerman: This can be deleted now yea?

PR #1074, opened 2023-04-04
Prove this is a bigger issue with the bootstrap library

### LINKED ISSUE #981: When you edit a content of tab block, it opens the last tab of that block
## Description of the bug

_When I edit the content of a page block within a tab, it opens the last tab of that block configuration._

## Reproduction steps

1.	Create or edit a page
2.	Add a record list
3.	Save the record list
4.	Add a new tabs block
5.	Add the record list created above to the tabs block
6.	Save and exit
7.	Open the tabs block again in editing mode
8.	Click on the tab “Tabs” (the second tab) to see the list of configured tabs
9.	Click on the “edit” button of the only tab that’s in there (the record list)
10.	 Now the block configuration on the record list opens, but not in the “General” tab, but in the “Automation” tab.

## Expected behavior 

_When the record block configuration opens I expect it to always show the “General” tab (so, not the last tab)_

## Screenshot/Video

Image

Image

## Version

Branch: 2023.3.0-dev.5.1 
Reporter: Lenny (Test fest)
Browser: Firefox

EmmyMay:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/PageBlocks/ContentConfigurator.vue
+++ b/client/web/compose/src/components/PageBlocks/ContentConfigurator.vue
@@ -1,5 +1,12 @@
 <template>
   <b-tab :title="$t('content.label')">
+    <b-button
+      variant="primary"
+      @click="editBlockTest"
+    >
+      Edit Block
+    </b-button>
+
     <b-form-group>
       <c-rich-text-input
         v-model="options.body"
@@ -29,5 +36,11 @@ export default {
   },
 
   extends: base,
+
+  methods: {
+    editBlockTest () {
+      this.$root.$emit('tab-editRequest', '4' /* Your block ID goes here for manual testing */)
+    },
+  },
 }
 </script>
--- a/client/web/compose/src/components/PageBlocks/TabsConfigurator.vue
+++ b/client/web/compose/src/components/PageBlocks/TabsConfigurator.vue
@@ -3,7 +3,12 @@
     <h5>
       {{ $t('tabs.displayTitle') }}
     </h5>
-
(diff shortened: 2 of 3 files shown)


EmmyMay:
You can pull this branch to see why this issue appears, in my opinion, to be external to bootstrap-vue. This issue only occurs when you try to open a configurator (à la "tabs in a modal") using a `$root` emit. 

### Approach

On this branch, I made a page with a `Tabs` `Content` `Content` `Recorlist`(this is where the issue is easily noticeable) which means my recordlist has a blockID of 3. I proceeded to tab all the blocks, and manually make a separate edit button as you would find in the `TabsConfigurator.vue` file, the `Builder.vue` file, and `ContentConfigurator.vue` file. The manual editing works by targeting the recordlist block (so pay attention to its blockID in your own test). 

### Findings

The manual edit buttons in the configurator files all rely on `$root` emits and this the only situation where opening the modal causes it to go to the last tab instead of the first one, as is normal when you click the manual edit button in the builder file. 

### Conclusion and Suggestion

I think this would be hard to reproduce externally, hence we would need to brainstorm a bit if we are going to make an issue on bootstrap-vue notifying them of this. Going through their issues page, I discovered they've had [problems with tabs in modals before](https://github.com/bootstrap-vue/bootstrap-vue/issues/3361) so they are not infallible in this area. We could do this and wait for a fix and simply let the bug live on till then.

A second suggestion would be for us to find a hacky way around, whereby we force the active tab to be the first one every time the configurator is opened.

PR #1083, opened 2023-04-06
Fix description of edited role shown when creating a new role

### LINKED ISSUE #1043: Description of edited role shown when creating a new role
## Reproduction steps
1 . Create a role with a description
2. Click on ‘New’ button
3.  See the previous role’s description is shown

## Expected behavior
The description from the prev role shouldn't be present when creating a new one

## Tested on
Safari on  MacOS Ventura 13.2.1

## Version
2023.3.0-dev.5

yonkov:


Fajfa (changes requested):
The change should happen in the role.ts class since thats the culprit
The defaultMeta should be shallow cloned when assigned in the constructor

PR #1088, opened 2023-04-10
 Error Handling Message In Workflows

### LINKED ISSUE #810: Error Handling Message In Workflows
## About

We would like to improve error handling inside workflows by letting configurators access the context of the error message.
Currently, we can only define what path the WF should take in case of error but we don't really have any way to figure out what happened.

Inside of the catch branch, we'd like to define some variables that specify what the error was, what step it was on, ...

## How

Up to you.
I'd suggest the error handler provides a way for you to define into what variable the error message (and surrounding context is placed in) so the configurator has more freedom in defining error handling.

Other options would be to just define a (set of) variable(s) that hold the error message and the context.

## Ref forum post

https://forum.cortezaproject.org/t/error-handling-message-in-workflows/1219

vicpatel:
Add support for error handler results

It adds error wrapper, error message, error step ID in result of error handler.

Please take a look.


server/pkg/wfexec/session.go
```diff
@@ -689,7 +688,18 @@ func (s *Session) exec(ctx context.Context, log *zap.Logger, st *State) (nxt []*
...
+			if ehr != nil {
+				if _, ok := ehr.GetValue()["error"]; ok {
+					_ = expr.Assign(scope, ehr.GetValue()["error"].Get().(string), expr.Must(expr.NewAny(st.err)))
+				}
+				if _, ok := ehr.GetValue()["errorMessage"]; ok {
+					_ = expr.Assign(scope, ehr.GetValue()["errorMessage"].Get().(string), expr.Must(expr.NewString(st.err.Error())))
+				}
+				if _, ok := ehr.GetValue()["errorStepID"]; ok {
+					_ = expr.Assign(scope, ehr.GetValue()["errorStepID"].Get().(string), expr.Must(expr.NewInteger(st.step.ID())))
+				}
```
tjerman: Let's reduce the number of `GetValue` calls.
The change probably won't make much of a difference in performance, but still.

server/pkg/wfexec/step.go
```diff
@@ -10,6 +10,8 @@ type (
 		ID() uint64
 		SetID(uint64)
 		Exec(context.Context, *ExecRequest) (ExecResponse, error)
+		// Results() interface{}
+		// SetResults(result interface{})
```
tjerman: Should this be removed?

tjerman (changes requested):
Apart from those comments; this looks good.
Can we add some test cases for this please.

tjerman (approved):
lgtm; squash and go

PR #1089, opened 2023-04-10
Fix error handling for avatar font file in auth profile

### LINKED ISSUE #996: Can't change the preferred language
## Description of the bug

_You cannot change the preferred language. If you select another language and hit save, the chosen language resets to the default one (English) and not the one you just selected and saved._

## Reproduction steps

1. Open your profile settings
2. Under preferred languages, select another language from the dropdown
3. Press save
4. Notice that the chosen language is not saved

## Expected behavior 

_You should be able to change your app language according to your preference._

## Screenshot/Video

https://user-images.githubusercontent.com/53232926/228196057-48a9287a-4f52-40f9-9550-aa8b8a2802ef.webm

## Version

Branch: 2023.3.0-dev.5.1
Reporter: Mariya (Test fest)
Browser: Chrome

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/996

Fix error handling for missing font file when generating user avatar initial in auth profile.

Previously, if a user attempted to update their own profile and the update triggered avatar initial generation with invalid font file configuration, the system did not handle the resulting error properly. This PR adds proper error handling to catch invalid font file configuration and inform the user why their update was not persisted.

### DIFF
--- a/server/auth/handlers/handle_profile.go
+++ b/server/auth/handlers/handle_profile.go
@@ -149,7 +149,7 @@ func (h *AuthHandlers) profileProc(req *request.AuthReq) error {
 		service.UserErrNotAllowedToUpdate().Is(err),
 		service.AttachmentErrInvalidAvatarFileType().Is(err),
 		service.AttachmentErrInvalidAvatarFileSize().Is(err),
-		service.AttachmentErrInvalidInitialsLength().Is(err):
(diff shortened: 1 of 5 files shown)


tjerman (commented):
Are you sure you want to wrap all of those error cases under a single error? Would it be better to have an error for each case? Perhaps just wrapping the error with the new one you created would do the trick.

PR #1142, opened 2023-04-17
Add optional invite email to new admin-created users

### LINKED ISSUE #469: Optionally send invitation email when a new user is created by an administrator
- [x] Add template that will be used for sending (similar to one we have for password-reset)
- [x] Extend `users add` CLI command with `--send-email` flag
- [x] Extend the user-create endpoint and add prop to request to trigger email sending when new user is created
- [x] Checkbox for admin console that should be checked for email to be sent

 Ref: #430

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/469

This PR adds the option for admins to send invitation emails to new users created with `users add` CLI command using `--send-email` flag. Users receive a link with an invitation token that expires after 72 hours, leading them to a password reset form. If the token is invalid, they are redirected to the login form with an error message.

 - [x] Code builds
 - [x] All existing tests pass


server/system/service/auth_credentials.go
```diff
@@ -228,6 +229,10 @@ func (svc *auth) createUserToken(ctx context.Context, u *types.User, kind string
 
 			// random password string, "3i[g0|)z"
 			token = fmt.Sprintf("%s", rand.Password(credentialsTokenLength))
+		case credentialsTypeInviteEmailToken:
+			// set the expiration time for all invited user tokens sent via email to 72 hours.
+			expiresAt = now().Add(time.Minute * 4320)
```
tjerman: Move this magic number to a constant; or better yet, make it configurable (probably the same way as the "is enabled" checkbox)

server/auth/handlers/links.go
```diff
@@ -74,6 +75,7 @@ func GetLinks() Links {
 		RequestPasswordReset:     b + "auth/request-password-reset",
 		PasswordResetRequested:   b + "auth/password-reset-requested",
 		ResetPassword:            b + "auth/reset-password",
+		InviteEmail:              b + "auth/invite-email",
```
tjerman: Can we think of a better endpoint name? I don't think `invite-email` fits here

server/system/commands/users.go
```diff
@@ -195,6 +202,12 @@ func Users(ctx context.Context, app serviceInitializer) *cobra.Command {
 		nil,
 		"Add user to roles (use ID or handle, repeat for multiple roles)")
 
+	addCmd.Flags().BoolVar(
+		&flagSendEmail,
+		"send-email",
```
tjerman: perhaps `send-invite` would be better?

tjerman (changes requested):
I'm ok with you defining a new thing for email invites (in addition to password resets), but I don't think I'm ok with you mixing and matching the two (so using new token types but reusing views).

Can you re-think how it should be (separate thing, updated pwd. reset, ...) and get to me?

server/auth/handlers/links.go
KinyaElGrande: I think `auth/invite` will make perfect sense.

client/web/admin/src/components/Settings/System/CSystemEditorAuth.vue
```diff
@@ -240,6 +240,37 @@
           <b-form-input v-model="settings['auth.mail.from-name']" />
         </b-input-group>
       </b-form-group>
+
+      <hr>
+
+      <h5>
+        Invite Email
```
tjerman: Shouldn't this be a translation?

client/web/admin/src/components/Settings/System/CSystemEditorAuth.vue
```diff
@@ -240,6 +240,37 @@
...
+          v-model="settings['auth.internal.send-user-invite-email.enabled']"
+          :value="true"
+          :unchecked-value="false"
+        >
+          {{ $t('internal.send-user-invite-email.enabled') }}
+        </b-form-checkbox>
+      </b-form-group>
+      <b-form-group
+        :label="$t('mfa.emailOTP.expires.label')"
+        description="How long will the invite token be valid before it expires"
```
tjerman: Shouldn't this be a translation?

server/auth/handlers/links.go
```diff
@@ -74,6 +75,7 @@ func GetLinks() Links {
 		RequestPasswordReset:     b + "auth/request-password-reset",
 		PasswordResetRequested:   b + "auth/password-reset-requested",
 		ResetPassword:            b + "auth/reset-password",
+		Invite:                   b + "auth/invite",
```
tjerman: perhaps `auth/accept-invite` would fit better (like (`auth/reset-password` and `auth/logout`)?

server/system/commands/users.go
```diff
@@ -195,6 +202,12 @@ func Users(ctx context.Context, app serviceInitializer) *cobra.Command {
 		nil,
 		"Add user to roles (use ID or handle, repeat for multiple roles)")
 
+	addCmd.Flags().BoolVar(
+		&flagSendEmail,
+		"send-invite",
+		false,
+		"Send invite email to user with password reset link")
```
tjerman: It's not a password reset link is it since you're setting it for the first time

server/system/service/auth_credentials.go
```diff
@@ -228,6 +229,15 @@ func (svc *auth) createUserToken(ctx context.Context, u *types.User, kind string
 
 			// random password string, "3i[g0|)z"
 			token = fmt.Sprintf("%s", rand.Password(credentialsTokenLength))
+		case credentialsTypeInviteEmailToken:
+			expSec := svc.settings.Auth.Internal.SendUserInviteEmail.Expires
+			if expSec == 0 {
+				expSec = 72
```
tjerman: Prefer to avoid magic numbers.
In this case, it's ok (should be trivial to figure out) but for more complex & used multiple times cases, define a (package/global) constant

tjerman (changes requested):
Last few from my end; looking good @KinyaElGrande

PR #1188, opened 2023-05-09
Add https to-all urls input values across corteza

### LINKED ISSUE #979: Add "http://" to all URLs input values across Corteza
## Notes
- Add "http://" to all URLs input values right before they are used, if they don't include (https:// or http://).
- Do not change original value in input!
- If there is anything before the ://, leave it. (https://, http://, mailto://, irc://). If not, add "http://" 

### Plan:

- [x] Write a function / class method that checks if url starts with https and append it if it doesn't.
- [x] Investigate twitter feed block component and add the logic to it
- [x] investigate iframe page block component and add the logic to it
- [x] investigate navigation page block links and add the logic to it
- [x] investigate admin user inteface settings and add the logic to it

yonkov:
Url inputs have been prefixed with "https" to the following places:

- Compose: twitter feed block
- Compose: iframe page block
- Compose: navigation page block
- Admin: admin user inteface settings

### DIFF
--- a/client/web/compose/src/components/PageBlocks/IFrameBase.vue
+++ b/client/web/compose/src/components/PageBlocks/IFrameBase.vue
@@ -8,7 +8,7 @@
       v-if="src"
       ref="iframe"
       class="h-100 w-100 border-0"
-      :src="src"
+      :src="src | checkValidURL"
     />
   </wrap>
 </template>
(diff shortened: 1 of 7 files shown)


Fajfa (changes requested):
The function should really be in vue/filters/url
And it shouldnt change the data from the resource, it should only be ran right before its used and shouldn't change/save anything

PR #1223, opened 2023-05-24
Add ability to select and perform operations on all records

### LINKED ISSUE #1167: Add select all items in record list (includes records on all pages)
Add the capability to select all records in the record list from all pages, and allow bulk editing/deleting/exporting/automation and other things you can do with selected records.

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1167

This PR introduces the capability to select all records and perform batch operations for editing, deleting, and undeleting records. It also includes the ability to exclude specific ones from the operation.

Key Features:

- Perform edit, delete, and undelete operations on all records in the recordlist
- Select all records and exclude specific ones from the operation
- Batch processing of records in groups of 500 for improved efficiency and performance

### Screenshots and videos

1. Select all records bulk edit sample video

https://github.com/cortezaproject/corteza/assets/32476094/5fd4164e-c1f7-4955-80d3-5b7cbf0965b0

2.  Automation video for prompting selected record ids from multiple pages

https://github.com/cortezaproject/corteza/assets/32476094/e1e1324d-6c0f-484f-84e6-4573e7c1f252

### DIFF
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -72,6 +72,7 @@
                 :query="query"
                 :prefilter="prefilter"
                 :selection="selected"
+                :selected-all-records-count="selectedAllRecords ? pagination.count : 0"
                 :processing="processing"
                 :preselected-fields="fields.map(({ moduleField }) => moduleField)"
                 class="mr-1 float-left"
@@ -164,37 +165,37 @@
         </div>
 
         <div
-          class="d-none flex-wrap align-items-center mt-1"
-          :class="{ 'd-flex': options.selectable && selected.length }"
+          v-if="options.selectable && selected.length"
+          class="d-flex align-items-center flex-wrap align-items-center mt-2"
         >
-          <div
-            class="d-flex align-items-baseline my-auto pt-1 text-nowrap h-100"
-          >
-            {{ $t('recordList.selected', { count: selected.length, total: items.length }) }}
-            <b-button
-              variant="link"
-              class="p-0 text-decoration-none"
-              @click.prevent="handleSelectAllOnPage({ isChecked: false })"
-            >
-              ({{ $t('recordList.cancelSelection') }})
-            </b-button>
+          <div class="mr-1">
+            {{ selectedRecordsDisplayText }}
           </div>
+
+          <b-button
+            size="sm"
+            variant="outline-light"
+            class="text-primary border-0"
+            @click="selectAllRecords()"
+          >
+            {{ selectedAllRecords ? $t('recordList.unselectAllRecords') : $t('recordList.selectAllRecords') }}
+          </b-button>
 
           <div class="d-flex align-items-center ml-auto">
             <automation-buttons
               class="d-inline m-0 mr-2"
               :buttons="options.selectionButtons"
               :module="recordListModule"
-              :extra-event-args="{ selected, filter }"
+              :extra-event-args="{ selected, filter}"
               v-bind="$props"
               @refresh="refresh()"
             />
 
             <bulk-edit-modal
-              v-show="options.bulkRecordEditEnabled && canUpdateSelectedRecords"
+              v-show="options.bulkRecordEditEnabled && canUpdateSelectedRecords && !showingDeletedRecords"
               :module="recordListModule"
               :namespace="namespace"
-              :selected-records="selected"
+              :selected-records="selectedAllRecords ? [] : selected"
               @save="onBulkUpdate()"
             />
 
@@ -433,7 +434,7 @@
                     :extra-options="options"
                   />
                   <div
-                    v-if="options.inlineRecordEditEnabled && field.canEdit"
+                    v-if="options.inlineRecordEditEnabled && field.canEdit && !showingDeletedRecords"
                     class="inline-actions"
                   >
                     <b-button
@@ -844,6 +845,7 @@ export default {
       customPresetFilters: [],
       currentCustomPresetFilter: undefined,
       showCustomPresetFilterModal: false,
+      selectedAllRecords: false,
     }
   },
 
@@ -1020,6 +1022,14 @@ export default {
     authUserRoles () {
       return this.$auth.user.roles
     },
+
+    selectedRecordsDisplayText () {
+      const count = this.selectedAllRecords ? (this.options.showTotalCount ? this.pagination.count : undefined) : this.selected.length
+      const total = this.items.length
+      const key = this.selectedAllRecords ? 'selectedFromAllPages' : 'selected'
+
+      return this.$t(`recordList.${key}`, { count, total })
+    },
   },
 
   watch: {
@@ -1150,6 +1160,8 @@ export default {
           return
         }
         this.selected.splice(i, 1)
+
+        this.selectedAllRecords = false
       }
     },
 
@@ -1174,6 +1186,7 @@ export default {
 
     handleShowDeleted () {
       this.showingDeletedRecords = !this.showingDeletedRecords
+      this.selectedAllRecords = false
       this.refresh(true)
     },
 
@@ -1381,7 +1394,7 @@ export default {
         query: {
           fields: e.fields,
           // url.Make already URL encodes the the values, so the filter shouldn't be encoded
-          filter: filter,
+          filter: this.selectedAllRecords ? '' : filter,
           jwt: this.$auth.accessToken,
           timezone: timezone ? timezone.tzCode : undefined,
         },
@@ -1468,7 +1481,13 @@ export default {
         this.selected = this.items.map(({ id }) => id)
       } else {
         this.selected = []
+        this.selectedAllRecords = isChecked
       }
+    },
+
+    selectAllRecords () {
+      this.selectedAllRecords = !this.selectedAllRecords
+      this.handleSelectAllOnPage({ isChecked: this.selectedAllRecords })
     },
 
     handleRestoreSelectedRecords () {
@@ -1482,15 +1501,19 @@ export default {
       } else {
         const { moduleID, namespaceID } = this.items[0].r
 
-        // filter undeletable records from the selected list
-        const recordIDs = this.items
-          .filter(({ id, r }) => r.canUndeleteRecord && this.selected.includes(id))
-          .map(({ id }) => id)
+        let query = ''
 
-        this.processing = true
+        if (!this.selectedAllRecords) {
+          // filter deletable records from the selected list
+          const recordIDs = this.items
+            .filter(({ id, r }) => r.canUndeleteRecord && this.selected.includes(id))
+            .map(({ id }) => id)
+
+          query = recordIDs.map(r => `recordID='${r}'`).join(' OR ')
+        }
 
         this.$ComposeAPI
-          .recordBulkUndelete({ moduleID, namespaceID, recordIDs })
+          .recordBulkUndelete({ moduleID, namespaceID, query })
           .then(() => {
             this.refresh(true)
             this.toastSuccess(this.$t('notification:record.undeleteBulkSuccess'))
@@ -1498,6 +1521,7 @@ export default {
           .catch(this.toastErrorHandler(this.$t('notification:record.undeleteBulkFailed')))
           .finally(() => {
             this.processing = false
+            this.selectedAllRecords = false
           })
       }
     },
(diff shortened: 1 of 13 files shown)


server/compose/rest.yaml
```diff
@@ -1095,6 +1095,9 @@ endpoints:
         name: ext
         required: true
         title: Export format
+      - type: "bool"
+        name:  excludeRecords
```
tjerman: how about just `exclude`?

server/compose/service/record.go
```diff
@@ -714,6 +718,147 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+		if err != nil {
+			return err
+		}
+
+		ids := records.IDs()
+		totalRecords := recordFilter.Total
+
+		for _, r = range records {
+			// check if the excluded ID is present in the fetched list and remove it from the records
+			excludeRec, index := contains(excludeIds, r.ID)
```
tjerman: consider indexing excludeIds OR filtering them out when fetching the records.
The list could (probably won't) be long.

server/compose/service/record.go
```diff
@@ -714,6 +718,147 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+		}
+
+		// prepare the next cursor
+		if f.PageCursor == nil {
+			f.PageCursor = &filter.PagingCursor{}
+		}
+		f.PageCursor.Set("ID", ids[len(ids)-1], false)
+
+		cursor = f.PageCursor.Encode()
+		cursor, _ = strconv.Unquote(cursor)
```
tjerman: This can be done much cleaner; have a think about what this search function returns, what the returned filter defines and see if you can better utilize that

server/compose/service/record.go
```diff
@@ -714,6 +718,147 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+		i += f.Limit
+
+		if int(f.Limit) > len(ids) {
+			i += uint(int(f.Limit) - len(ids))
+		}
+
+		if i >= totalRecords {
+			break
+		}
+	}
```
tjerman: You can get rid of all of this if you better utilize what the search function returns and how it works

server/compose/service/record.go
```diff
@@ -714,6 +718,147 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+			NamespaceID: namespaceID,
+			ModuleID:    moduleID,
+			Deleted:     filter.State(1),
+		}
+	)
+
+	return svc.processUpdateAll(ctx, f, nil, types.OperationTypeUndelete, recordIds)
+}
+
+func (svc record) processUpdateAll(ctx context.Context, f types.RecordFilter, values types.RecordValueSet, operation types.OperationType, excludeIds []uint64) (err error) {
```
tjerman: should this function get wrapped in a transaction? Should it all be run in a transacton?

tjerman (changes requested):

Fajfa (changes requested):
1. Make sure you use the query approach for all bulk endpoints
2. Decide if we want to either select all or just worry about the selected on the current page
3. Change text for number of selected records to be more concise
4. Change button and text for selecting all records

server/compose/automation/records_handler.go
```diff
@@ -232,6 +234,17 @@ func (h recordsHandler) each(ctx context.Context, args *recordsEachArgs) (out wf
...
+		for _, idTypeValue := range args.Exclude {
+			if idStr, ok := idTypeValue.Get().(string); ok {
+				id := payload.ParseUint64(idStr)
+				excludeRec, index := slice.HasUint64(i.buffer.IDs(), id)
+				if excludeRec {
+					i.buffer = append(i.buffer[:index], i.buffer[index+1:]...)
+				}
+			}
+		}
+
```
tjerman: Clean this up, please

* constant parsing from TypedValue -> string -> uint64 is just not needed
* `slice.HasUint64` is a linear check; this will probably have a bit of impact if the exclude list is long (probably won't be but could be).
* would it be more efficient to fetch records into a separate slice (not `i.buffer`) and then selectively copy the records over (into `i.buffer`)

server/compose/service/record.go
```diff
@@ -714,6 +719,133 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
 	}
 }
 
+func (svc record) UpdateAll(ctx context.Context, namespaceID uint64, moduleID uint64, values types.RecordValueSet, recordIds ...uint64) (err error) {
```
tjerman: `recordIds` is later refered as `excludeIds` inside `processUpdateAll`. It'll throw people off

server/compose/rest.yaml
```diff
@@ -1163,6 +1166,7 @@ endpoints:
       post:
       - { name: records, type: "[]string", title: Records to update }
       - { name: values, type: "types.RecordValueSet", title: Fields to update and their values }
+      - { name: includeAllRecords, type: "bool", title: Flag value for selecting all records }
```
tjerman: I can get behind adding this new parameter; I'm just not sure about us changing the meaning of what `records` does when this is set (when set, it becomes the exclude list). What do you think? Should we have an extra parameter just for excluded records?

tjerman: This would apply to all of affected endpoints

server/compose/rest/record.go
```diff
@@ -588,8 +614,39 @@ func (ctrl *Record) Export(ctx context.Context, r *request.RecordExport) (interf
...
+
+					if !foundMatch {
+						exportBuffer.WriteString(line)
+						exportBuffer.WriteString("\n")
+					}
+				}
+
+				s.Source = &exportBuffer
+			}
+
```
tjerman: We should apply the filter when getting the records out of the DB, not cutting things out of the output file.
We can take a look at it later (tomorrow or monday probably)

tjerman (changes requested):
First bit; the second comes tomorrow

server/compose/service/record.go
```diff
@@ -714,6 +719,133 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+	return store.Tx(ctx, svc.store, func(ctx context.Context, s store.Storer) error {
+		// load both the namespace and module
+		if ns, m, err = loadModuleCombo(ctx, svc.store, f.NamespaceID, f.ModuleID); err != nil {
+			return err
+		}
+
+		aProps.setNamespace(ns)
+		aProps.setModule(m)
+
+		f.Limit = 10
```
tjerman: Why did you decide to go with such a low number?
I usually go with 300-500 (that was the sweet spot for me).

server/compose/service/record.go
```diff
@@ -714,6 +719,133 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+		cursor := ""
+		f.IncPageNavigation = true
+		f.IncTotal = true
```
tjerman: Let's have a re-think how we can clean this up.

What does `svc.Find` return? Records and a filter which has a `NextPage *PagingCursor` field.
Can we straight up use the `NextPage` field to get rid of getting the cursor to a string and then back into a cursor?

Also, why do we need `IncPageNavigation` and `IncTotal`?

server/pkg/slice/ints.go
```diff
@@ -11,14 +11,18 @@ type (
 	UInt64s []uint64
 )
 
-func HasUint64(ss []uint64, s uint64) bool {
-	for i := range ss {
-		if ss[i] == s {
-			return true
+func HasUint64(ss []uint64, s uint64) (bool, int) {
```
tjerman: does this need a bool and an int? -1 is an invalid index which could be used to replace the bool `false`

server/pkg/slice/strings.go
```diff
@@ -1,15 +1,15 @@
 package slice
 
-func ContainsAny[C comparable](haystack []C, needles ...C) bool {
-	for _, h := range haystack {
+func ContainsAny[C comparable](haystack []C, needles ...C) (bool, int) {
```
tjerman: same comment as with the uint64 slice.
you could also consider replacing the contents of `HasUint64` with a call to this, since it's a generic.

tjerman (changes requested):

server/compose/rest/record.go
```diff
@@ -233,34 +230,29 @@ func (ctrl *Record) Create(ctx context.Context, r *request.RecordCreate) (interf
...
-	}
+	err = ctrl.record.BulkModifyByFilter(ctx, f, req.Values, types.OperationTypePatch)
 
-	results, err := ctrl.record.Bulk(ctx, true, oo...)
 	if rve := types.IsRecordValueErrorSet(err); rve != nil {
 		return ctrl.handleValidationError(rve), nil
 	}
 
-	return ctrl.makeRecordBulkPatchPayload(ctx, results, err)
+	return nil, err
```
tjerman: Let's return something here; if all else fails we can return `api.OK()`

server/compose/service/record.go
```diff
@@ -714,6 +716,88 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+			module:    &types.Module{ID: f.ModuleID},
+		}
+		action func(props ...*recordActionProps) *recordAction
+
+		valueError *types.RecordValueErrorSet
+	)
+
+	return store.Tx(ctx, svc.store, func(ctx context.Context, s store.Storer) error {
+		// load both the namespace and module
+		if ns, m, err = loadModuleCombo(ctx, svc.store, f.NamespaceID, f.ModuleID); err != nil {
```
tjerman: use `s` over `svc.store`

server/pkg/slice/ints.go
```diff
@@ -12,8 +12,12 @@ type (
 )
 
 func HasUint64(ss []uint64, s uint64) bool {
-	for i := range ss {
-		if ss[i] == s {
+	if len(ss) == 0 {
+		return false
+	}
```
tjerman: What does this exception cover? If `ss` is empty/`nil`, the loop on `19` won't iterate ever so it'll be `false`. Is it something else I've missed?
I'm not sure if this improves performance at all so I'd consider removing it (less is more, sometimes).

I'll leave the decision with you, either works for me.

tjerman (changes requested):
Last few bits; then the server side of things looks good to me.
Good job :+1:

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -155,22 +156,36 @@
...
+                v-if="!selectedAllRecords"
+                variant="link"
+                class="p-0 text-decoration-none"
+                @click.prevent="handleSelectAllOnPage({ isChecked: false })"
+              >
+                ({{ $t('recordList.cancelSelection') }})
+              </b-button>
+            </div>
+
+            <div class="ml-3">
```
Fajfa: No need for this wrapper, just do ml-3 on the button itself

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -155,22 +156,36 @@
...
-            <b-button
-              variant="link"
-              class="p-0 text-decoration-none"
-              @click.prevent="handleSelectAllOnPage({ isChecked: false })"
-            >
-              ({{ $t('recordList.cancelSelection') }})
-            </b-button>
+            <div>
+              {{ selectedAllRecords ? $t('recordList.selectedFromAllPages', { count: getPagination.count }) : $t('recordList.selected', { count: selected.length, total: items.length }) }}
+              <b-button
```
Fajfa: You can remove the cancel button since it's a bit confusing now with Select on all pages button.

client/web/compose/src/mixins/record.js
```diff
@@ -281,8 +281,14 @@ export default {
 
       const { moduleID, namespaceID } = this.module
 
+      let query = ''
+
+      if (!includeAllRecords) {
```
Fajfa: Wouldn't it make sense to always evaluate the query? If records is empty then you know you're processing them all. So in essence, `query = records.map(...` in all cases

Fajfa (changes requested):

Fajfa (approved):

server/compose/service/record.go
```diff
@@ -714,6 +716,85 @@ func (svc record) Bulk(ctx context.Context, skipFailed bool, oo ...*types.Record
...
+			return err
+		}
+
+		aProps.setNamespace(ns)
+		aProps.setModule(m)
+
+		f.Limit = 500
+		if f.Paging, err = filter.NewPaging(f.Limit, ""); err != nil {
+			return err
+		}
```
tjerman: Why? If you don't pass in a paging cursor that means it gives you the first page (no filtering). Can we not abuse that here?

tjerman (commented):
One tiny bit, else lgtm; good job :ok_hand:

tjerman (approved):
See my prior comment; clicked on the wrong thing

katrinDY:
QC is happy

PR #1227, opened 2023-05-24
Add record filter preset saving

### LINKED ISSUE #1172: Add saving of record filter presets per user
## Description of the bug

User can save a record list filter as preset.

- Add `save as a preset` button on the record list filter
- which saves the current filter as filter preset
- a modal to is required to add a name to the filter preset
- the saved filter preset can be retrieved from the localstorage and added to the filter preset
- Filter presets can be configurable in the record list configurator

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/Common/RecordListFilter.vue
+++ b/client/web/compose/src/components/Common/RecordListFilter.vue
@@ -215,13 +215,23 @@
             {{ $t('general:label.reset') }}
           </b-button>
 
-          <b-button
-            ref="btnSave"
-            variant="primary"
-            @click="onSave"
-          >
-            {{ $t('general.label.save') }}
-          </b-button>
+          <div class="d-flex">
+            <b-button
+              v-if="allowFilterPresetSave"
+              variant="outline-primary"
+              class="mr-2"
+              @click="onSave(true, 'filter-preset')"
+            >
+              {{ $t('recordList.filter.addFilterToPreset') }}
+            </b-button>
+            <b-button
+              ref="btnSave"
+              variant="primary"
+              @click="onSave"
+            >
+              {{ $t('general.label.save') }}
+            </b-button>
+          </div>
         </b-card-footer>
       </b-card>
 
@@ -287,6 +297,11 @@ export default {
     buttonStyle: {
       type: String,
       default: '',
+    },
+
+    allowFilterPresetSave: {
+      type: Boolean,
+      default: false,
     },
   },
 
@@ -592,13 +607,8 @@ export default {
       }
     },
 
-    onSave (close = true) {
-      if (close) {
-        this.$refs.popover.$emit('close')
-      }
-
-      // Emit only value and not whole record with every filter
-      this.$emit('filter', this.componentFilter.map(({ groupCondition, filter = [], name }) => {
+    processFilter () {
+      return this.componentFilter.map(({ groupCondition, filter = [], name }) => {
         filter = filter.map(({ record, ...f }) => {
           if (record) {
             f.value = record[f.name] || record.values[f.name]
@@ -615,7 +625,16 @@ export default {
         })
 
         return { groupCondition, filter, name }
-      }))
+      })
+    },
+
+    onSave (close = true, type = 'filter') {
+      if (close) {
+        this.$refs.popover.$emit('close')
+      }
+
+      // Emit only value and not whole record with every filter
+      this.$emit(type, this.processFilter())
     },
 
     updateFilterProperties (filter) {
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -78,22 +78,32 @@
                 @export="onExport"
               />
 
-              <b-dropdown
-                v-if="filterPresets.length"
-                size="lg"
-                variant="light"
-                right
-                :text="$t('recordList.filter.filters.label')"
+              <b-button
+                :id="`${uniqueID}-filter-preset`"
+                class="btn dropdown-toggle btn-light btn-lg"
               >
-                <b-dropdown-item
-                  v-for="(f, idx) in filterPresets"
-                  :key="idx"
-                  :disabled="activeFilters.includes(f.name)"
-                  @click="updateFilter(f.filter, f.name)"
-                >
-                  {{ f.name }}
-                </b-dropdown-item>
-              </b-dropdown>
+                {{ $t('recordList.filter.filters.label') }}
+              </b-button>
+              <b-popover
+                :target="`${uniqueID}-filter-preset`"
+                triggers="click blur"
+                placement="bottom"
+              >
+                <ul class="list-style-none">
+                  <li
+                    v-for="(f, idx) in filterPresets"
+                    :key="idx"
+                  >
+                    <button
+                      class="dropdown-item"
+                      :disabled="activeFilters.includes(f.name)"
+                      @click="updateFilter(f.filter, f.name)"
+                    >
+                      {{ f.name }}
+                    </button>
+                  </li>
+                </ul>
+              </b-popover>
 
               <column-picker
                 v-if="!options.hideConfigureFieldsButton"
@@ -293,8 +303,10 @@
                       :namespace="namespace"
                       :module="recordListModule"
                       :record-list-filter="recordListFilter"
+                      :allow-filter-preset-save="options.customFilterPresets"
                       class="d-print-none ml-1"
                       @filter="onFilter"
+                      @filter-preset="onSaveFilterPreset"
                       @reset="activeFilters = []"
                     />
 
@@ -609,6 +621,13 @@
         open-on-select
         @save="onInlineEdit()"
       />
+
+      <!-- Modal for naming custom filter -->
+      <custom-filter-preset
+        :visible="showCustomPresetFilterModal"
+        @save="setStorageRecordListFilterPreset"
+        @close="showCustomPresetFilterModal = false"
+      />
     </template>
 
     <template
@@ -733,6 +752,7 @@ import draggable from 'vuedraggable'
 import RecordListFilter from 'corteza-webapp-compose/src/components/Common/RecordListFilter'
 import ColumnPicker from 'corteza-webapp-compose/src/components/Admin/Module/Records/ColumnPicker'
 import BulkEditModal from 'corteza-webapp-compose/src/components/Public/Record/BulkEdit'
+import CustomFilterPreset from 'corteza-webapp-compose/src/components/Public/Record/CustomFilterPreset'
 
 const { CInputSearch } = components
 
@@ -752,6 +772,7 @@ export default {
     ColumnPicker,
     CInputSearch,
     BulkEditModal,
+    CustomFilterPreset,
   },
 
   extends: base,
@@ -815,6 +836,9 @@ export default {
       items: [],
       showingDeletedRecords: false,
       activeFilters: [],
+      customPresetFilters: [],
+      currentCustomPresetFilter: undefined,
+      showCustomPresetFilterModal: false,
     }
   },
 
@@ -982,7 +1006,10 @@ export default {
     },
 
     filterPresets () {
-      return this.options.filterPresets.filter(({ name, roles }) => name && this.isUserRoleMember(roles))
+      return [
+        ...this.options.filterPresets.filter(({ name, roles }) => name && this.isUserRoleMember(roles)),
+        ...this.customPresetFilters,
+      ]
     },
 
     authUserRoles () {
@@ -1008,6 +1035,7 @@ export default {
       handler () {
         this.createEvents()
         this.getStorageRecordListFilter()
+        this.getStorageRecordListFilterPreset()
         this.prepRecordList()
         this.refresh(true)
       },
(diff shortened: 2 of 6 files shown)


client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -989,7 +991,11 @@ export default {
     },
 
     filterPresets () {
-      return this.options.filterPresets.filter(({ name, roles }) => name && this.isUserRoleMember(roles))
+      let presets = []
+      presets = presets.concat(this.options.filterPresets.filter(({ name, roles }) => name && this.isUserRoleMember(roles)))
```
Fajfa:
```
[... array1, ...array2 ]
```
```
return [
   ...this.options.filterPresets.filter(({ name, roles }) => name && this.isUserRoleMember(roles)),
   ...this.customPresetFilters
]
```
You can concat arrays like this if you wish

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1085,6 +1102,16 @@ export default {
...
     },
 
+    onSaveFilterPreset (filter = []) {
+      this.customPresetFilters = [{
+        name: this.$t('recordList.customPresetFilter'),
+        filter,
+      }]
+
+      this.setStorageRecordListFilterPreset(filter)
+      this.refresh(true)
```
Fajfa, line 1127: I dont think you need to refresh here, you just need to fetch the presets

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1626,6 +1669,23 @@ export default {
...
+        // Get record list filters from localStorage
+        currentListFilters = getItem(`record-list-preset-${this.uniqueID}`)
+        currentListFilters = {
+          name: this.$t('recordList.customPresetFilter'),
+          filter,
+        }
+
+        setItem(`record-list-preset-${this.uniqueID}`, currentListFilters)
+      } catch (e) {
+        console.warning(this.$t('notification:record-list.corrupted-filter'))
```
Fajfa: you probably mean console.warn here

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1626,6 +1669,23 @@ export default {
       }
     },
 
+    setStorageRecordListFilterPreset (filter = []) {
+      let currentListFilters = []
+
+      try {
+        // Get record list filters from localStorage
+        currentListFilters = getItem(`record-list-preset-${this.uniqueID}`)
```
Fajfa: Do you need this, since you immediately overwrite it in the next line anyway

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -320,6 +322,7 @@ recordList:
     addField: Add new filter field
     addFilter: + Add filter
     byValue: Filter records based on field value
+    addFilterToPreset: Add filter to preset
```
Fajfa: Save as preset

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1792,6 +1852,10 @@ td:hover .inline-actions {
     color: $primary !important;
   }
 }
+
+::v-deep .b-form-tags-button {
```
Fajfa: You should use this is the style non scoped section, or even better add it to custom.scss

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1626,6 +1669,23 @@ export default {
       }
     },
 
+    setStorageRecordListFilterPreset (filter = []) {
+      let currentListFilters = []
```
Fajfa: Im thinking it might make sense to show a modal with input, so you can name your filter preset. Otherwise custom filter 1/2 and so on dont tell you much

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1613,6 +1641,21 @@ export default {
       }
     },
 
+    getStorageRecordListFilterPreset () {
+      try {
+        // Get record list filters from localStorage
+        const currentFilters = getItem(`record-list-preset-${this.uniqueID}`)
```
Fajfa: This is probably currentFilterPresets right?

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1613,6 +1641,21 @@ export default {
...
+
+        // Set the custom preset filters
+        this.customPresetFilters = [currentFilters]
```
Fajfa: isn't currentFilters and array? That would mean you would have to use [...currentFilters], i might be looking at this wrong tho

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1071,12 +1078,22 @@ export default {
...
-          })
+          const filterPresets = this.filterPresets.find(p => p.name === f.name)
+
+          if (filterPresets) {
+            filterPresets.filter.forEach((filterPreset) => {
+              if (!isEqual(f.filter, filterPreset.filter)) {
+                const filterIndex = this.activeFilters.indexOf(f.name)
+                this.activeFilters.splice(filterIndex, 1)
+
+                this.activeFilters.push(this.$t('recordList.customFilter'))
```
Fajfa: Since the same code is in lines 1095 and 1096 it makes me think you could rewrite this to only use the code once (lines 1089/1090 are same as 1095/1096)

Fajfa (approved):

Fajfa (changes requested):
Mistakenly selected Approve.

client/web/compose/src/components/PageBlocks/RecordListBase.vue
kelanik8: This format follows the same way the preset filter are saved on the options object, but in this case all the custom filter are the same

client/web/compose/src/components/Common/RecordListFilter.vue
```diff
@@ -615,7 +619,25 @@ export default {
...
+      })
+    },
+
+    onSave (close = true) {
+      if (close) {
+        this.$refs.popover.$emit('close')
+      }
+
+      // Emit only value and not whole record with every filter
+      this.$emit('filter', this.processFilter())
```
Fajfa: This looks like you can just have a onSave function with (close = true, type = 'filter').
So that code doesn't repeat

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+<script>
+export default {
+  i18nOptions: {
+    namespaces: 'block',
+  },
+
+  name: 'CustomFilterPreset',
+
+  props: {
+    showCustomPresetFilterModal: {
```
Fajfa: can this be called `visible`, since based on where it is it is obvious it's the custom filter preset modal (component name)

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+      type: Boolean,
+      default: false,
+    },
+
+    customFilter: {
```
Fajfa: this can just be called filter

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+
+  data () {
+    return {
+      showModal: false,
+      filterName: '',
+    }
+  },
+
+  watch: {
+    showCustomPresetFilterModal (val) {
```
Fajfa: Do the visible: { immediate: true, handler () } syntax please

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+      this.showModal = val
+    },
+  },
+
+  created () {
+
```
Fajfa: No need for this if its empty

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+  },
+
+  methods: {
+    onModalHide () {
+      this.showModal = false
+    },
+
+    onReset () {
```
Fajfa: We dont need a reset on this modal, just a save and cancel

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+  <div>
+    <b-modal
+      :visible="showModal"
+      :title="$t('recordList.presetFilter.addFilterPresetName')"
+      body-class="p-0"
+      footer-class="flex-column align-items-stretch"
```
Fajfa: d-flex w-100 align-items center and justify-content-between

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -263,6 +263,12 @@ record:
 recordList:
   addRecord: Add
   cancelSelection: Cancel
+  customFilter: Custom filter
+  customPresetFilter: Custom preset filter
+  presetFilter:
+    reset: Reset
+    filterName: Filter name
+    addFilterPresetName: Add filter preset name
```
Fajfa: Save filter as preset

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -263,6 +263,12 @@ record:
 recordList:
   addRecord: Add
   cancelSelection: Cancel
+  customFilter: Custom filter
+  customPresetFilter: Custom preset filter
```
Fajfa: i dont think you use this anywhere

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -263,6 +263,12 @@ record:
...
+  presetFilter:
+    reset: Reset
```
Fajfa: This also wont be needed since you don't need reset

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -263,6 +263,12 @@ record:
 recordList:
   addRecord: Add
   cancelSelection: Cancel
+  customFilter: Custom filter
+  customPresetFilter: Custom preset filter
+  presetFilter:
```
Fajfa: filterPresets, presetFilter means something else

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+      centered
+      @hide="onModalHide"
+    >
+      <b-card
+        class="pt-0"
+      >
+        <b-form-group
+          :label="$t('recordList.presetFilter.filterName')"
```
Fajfa: label-class primary

client/web/compose/src/components/Public/Record/CustomFilterPreset/index.vue
```diff
@@ -0,0 +1,120 @@
...
+        >
+          <b-form-input
+            v-model="filterName"
+            required
```
Fajfa: the required here doesn't have any effect, you should however prevent saving(button should be disabled) if there is no filter name. You can also indicate its required by having a * in the label

Fajfa (changes requested):
1. If i close the save as preset naming modal I cannot save another preset
2. Apply filter that you save as preset
3. If I press reset, the Custom filter still stays
4. This should be configurable, that means a new option in the record list configurator `customFilterPresets`, should be a checkbox in the filter preset section
5. Make sure the record list filter component is also configurable to be able to hide that save as preset button since its not just used in the record list. And make sure its disabled in other places record list filter is used. We only want custom presets on record list block.

Fajfa (changes requested):
1. If i close the save as preset naming modal I cannot save another preset
2. Saving as preset without applying filter doesn't apply and/or save the not applied one in the preset
3. If I press reset, the Custom filter still stays
4. This should be configurable, that means a new option in the record list configurator `customFilterPresets`, should be a checkbox in the filter preset section
5. Make sure the record list filter component is also configurable to be able to hide that save as preset button since its not just used in the record list. And make sure its disabled in other places record list filter is used. We only want custom presets on record list block.

katrinDY:
Issues found after QC check:
- on clone/copy block, preset and "normal" filters are not copied into the new block config. The value of the option to set custom filters is cloned/copied correctly. Steps to reproduce: in a block click on copy block or copy block button -> in the newly copied/clone block notice how both of the filters are not present. _This is not supported for now_
- "normal" filters are not shown after page refresh
- after adding a preset filter, it’s shown in active filter section but not applied to the RL. Steps to reproduce: add a filter -> set it as preset -> remove the "normal filter" -> apply the preset filter -> notice it's not respected

Fajfa (approved):

katrinDY:
Last thing:
- if I add a  preset filter, it's applied normally but the same filter is added as a normal filter. Please see the gif. Notice how after choosing the preset filter, a normal filter is added and after removing the normal filter, the preset filter is removed. Steps to reproduce: add a filter -> set it as prefilter -> remove the normal filter -> the preset filter is removed _This is expected_

Fajfa (approved):

PR #1238, opened 2023-06-01
Fixed From header on env SMTP vars

### LINKED ISSUE #1149: service.auth-notification auth notification "could not send email"}

### Version of Corteza

2022.9.9

### Current Behavior

While deploying a fresh corteza project with proper SMTP settings, the value SMTP_FROM is not set in the auth-notification.

It is exactly(!) the same behavior as described in #372. Is there a quick fix? Yaren-IT has already asked this question in #372. There may be an answer by now.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment and versions

_No response_

### Anything else?

_No response_

petergrlica:
Fixes #1149 as the default provisioned value for an email is invalid on some smtp servers (providers).

Removed the email that is fetched from settings as the `.env` vars are not added to settings at that time, the default SMTP dialer does set the default `From` value to the correct one.

Omitted the email from template as it did not add any benefits to the user experience and would only complicate the solution.

### DIFF
--- a/server/app/boot_levels.go
+++ b/server/app/boot_levels.go
@@ -973,6 +973,7 @@ func setupSmtpDialer(log *zap.Logger, servers ...types.SmtpServers) {
 		zap.String("host", s.Host),
 		zap.Int("port", s.Port),
 		zap.String("user", s.User),
+		zap.String("from", s.From),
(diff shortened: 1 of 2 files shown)


tjerman:
lgtm :+1:

PR #1239, opened 2023-06-02
Improve Reminders Layout

### LINKED ISSUE #1056: Reminders improvement
## Description of the bug

UI/UX improvement, mostly not a bug I don’t find reminders intuitive.

1.	Should the + add a new reminder button become something cleaner (do we need it to be so verbose?

Image

Btn-outline-primary -> btn-primary
Text-center -> no alignment

Consistent with records and other add something lists like records

Image

2.	Should the toggle become a checkbox? A toggle tells me that something can be ON or OFF; so here enable/disable the reminder, not dismiss it – I think a checkbox would fit better (or a button that writes “dismiss”

Image

3.	Can we show more content or make the content accessible with less work? Z, for example, has a description. Can we show (a truncated) description under the label? Alternatively, making the reminder clickable to expand/collapse with details would (in my opinion) work also.

4. I don’t find the left two icons (that the reminder is snoozed and for what time it’s set) intuitive. Do we need to hide information behind icons/tooltip instead of just showing it? In addition, the left (that it’s snoozed) and the right one show the same data (bug?):

Image

Image

5.	When editing a reminder, should the save button be moved next to the back button (same setup as elsewhere)?

Image

6.	Should the fact the reminder was snoozed displayed on the reminder edit page, like it is on the view page?

Image

Image

7. Should the reminders panel change to feel the same as the navigation panel? When expanded (on large screens) should it push the content instead of rendering over the content?
Notes:
I really like how Apple’s reminders are displayed; can we utilize what they have going on?

Image

## Version

Branch: 2023.3.0-dev.5.1
Reporter: Tomaz (Test fest)
Browser: Firefox

yonkov:

Improvements in the compose reminders have been made, based on test fest feedback.

## Reminders List Page 
image

## Reminders Edit Page
image

### DIFF
--- a/client/web/compose/src/components/Namespaces/Reminders/List.vue
+++ b/client/web/compose/src/components/Namespaces/Reminders/List.vue
@@ -1,108 +1,98 @@
 <template>
   <div
-    class="h-100"
+    class="d-flex flex-column h-100"
   >
-    <div
-      v-for="(r, i) in sortedReminders"
-      :key="r.reminderID"
-      :style="`${!!r.dismissedAt ? 'opacity:.6;' : ''}`"
-    >
-      <hr v-if="r.dismissedAt && sortedReminders[i - 1] ? !sortedReminders[i - 1].dismissedAt : false ">
-
+    <div class="flex-fill overflow-auto px-2">
       <div
-        class="overflow-auto border card shadow my-2 p-1"
+        v-for="(r, i) in sortedReminders"
+        :key="r.reminderID"
+        :style="`${!!r.dismissedAt ? 'opacity:.6;' : ''}`"
       >
+        <hr v-if="r.dismissedAt && sortedReminders[i - 1] ? !sortedReminders[i - 1].dismissedAt : false ">
+
         <div
-          class="d-flex flex-row flex-nowrap align-items-center"
+          class="overflow-auto border card shadow-sm my-2 p-1"
         >
-          <b-form-checkbox
-            v-b-tooltip.hover.left.350
-            data-test-id="checkbox-dismiss-reminder"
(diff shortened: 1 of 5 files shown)


client/web/compose/src/components/Namespaces/Reminders/List.vue
```diff
@@ -2,108 +2,129 @@
...
-        size="sm"
-        variant="outline-primary"
-        @click="$emit('edit')"
-      >
-        + {{ $t('reminder.add') }}
-      </b-button>
-    </div>
-
-    <div
+    rounded-2    <div
```
Fajfa: Probably a lapsus

client/web/compose/src/components/Namespaces/Reminders/List.vue
```diff
@@ -152,6 +173,10 @@ export default {
     recordViewer ({ params } = {}) {
       return params ? { name: 'page.record', params } : undefined
     },
+
+    formatDate (d) {
+      return moment(d).format('L')
```
Fajfa: It also needs to show the time of the reminder and not just the date.
Let's avoid using moment, we have filters in vue for this purpose, or you can use native javascript.
We're trying to move away from moment since it stopped being maintained because es6 datetime works just as well. Also double check it properly maps to local timezone

Fajfa (changes requested):

client/web/compose/src/components/Namespaces/Reminders/List.vue
yonkov: Good point, I updated the code to use `locFullDateTime` filter.

PR #1321, opened 2023-07-11
Abort ongoing backend request on a record list block when page is switched

### LINKED ISSUE #994: Requests to backend are not dropped after page leave
The issue gets intensified on a slow network or browsing pages with bigger data sets.
Once the request is fetched, leaving the page will keep the connection and not cancel the Promise().

## Reproduction
Enable network throttling in developer toolbar
Navigate to a few of pages with bigger record lists
Check for connection responses in the toolbar

## Expected behavior
The connections should be dropped once a user leaves the page with the record list.

## Video
Image

Components

- [x] Record list
- [x] Automation
- [x] Comment
- [x] Record
- [x] Record Organizer
- [x] Report
- [x] Geometry
- [x] Metric
- [x] Calender

Pages
- [x] modules edit
- [x] modules create -> hasRecord request which enables field type editing
- [x] record view -> record values request
- [x] pages edit -> roles options reqeust on the configure layout modal
- [x] pages list -> page list tree request
- [x] page record -> load record values request

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/PageBlocks/AutomationBase.vue
+++ b/client/web/compose/src/components/PageBlocks/AutomationBase.vue
@@ -40,6 +40,8 @@ export default {
       processing: false,
 
       automationScripts: [],
+
+      abortRequests: [],
     }
   },
 
@@ -49,19 +51,37 @@ export default {
     },
   },
 
+  beforeDestroy () {
+    this.abortRequests.forEach((cancel) => {
+      cancel()
+    })
+  },
+
   created () {
     if (this.$UIHooks.set && !!this.$UIHooks.set.length) {
       return
     }
 
-    this.processing = true
-    return this.$ComposeAPI.automationList({ eventTypes: ['onManual'], excludeInvalid: true })
-      .then(({ set = [] }) => {
-        this.automationScripts = set
-      })
-      .finally(() => {
-        this.processing = false
-      })
+    this.fetchAutomationLists()
+  },
+
+  methods: {
+    fetchAutomationLists () {
+      this.processing = true
+
+      const { response, cancel } = this.$ComposeAPI
+        .automationListCancellable({ eventTypes: ['onManual'], excludeInvalid: true })
+
+      this.abortRequests.push(cancel)
+
+      return response()
+        .then(({ set = [] }) => {
+          this.automationScripts = set
+        })
+        .finally(() => {
+          this.processing = false
+        })
+    },
   },
 
   beforeDestroy () {
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -722,6 +722,7 @@
   </wrap>
 </template>
 <script>
+import axios from 'axios'
 import { debounce, isEqual } from 'lodash'
 import { mapGetters, mapActions } from 'vuex'
 import base from './base'
@@ -828,6 +829,8 @@ export default {
       currentCustomPresetFilter: undefined,
       showCustomPresetFilterModal: false,
       selectedAllRecords: false,
+
+      abortRequests: [],
     }
   },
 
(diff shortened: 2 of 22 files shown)


Fajfa (changes requested):
Change the trigger of when that happens (beforeDestroy), which should be good.

Fajfa:
It works great, i think if you can pinpoint a few other requests that would benefit from this
Maybe some record loading, page loading in builder, maybe the chart reporter endpoint.
If you can cover the important(ones who take the most time) until the next release that would be great for performance

Fajfa (changes requested):
This is a good start, for page blocks. There are still other places, for example views and module fields that would most likely benefit from this, aswell as initial namespace loading.

katrinDY:
LGTM except for one minor issue:

Screenshot 2023-08-03 at 12 31 32

client/web/compose/src/components/PageBlocks/AutomationBase.vue
```diff
@@ -61,23 +61,27 @@ export default {
...
+          { eventTypes: ['onManual'], excludeInvalid: true },
+          { cancelToken: this.cancelTokenSource.token },
+        )
+        .then(({ set = [] }) => {
+          this.automationScripts = set
+        })
+        .finally(() => {
+          this.processing = false
+        })
+    },
```
katrinDY, line 84: Let's add a line between the methods for better readability

client/web/compose/src/components/PageBlocks/RecordListBase.vue
tjerman, line 1: I think we should also cancel requests for the record list before we do any additional requests (such as changing filters, sorting, ...). Not cancelling old requests would cause issues such as seeing data from the previous request instead of the new request.

tjerman (changes requested):
I have two comments regarding this.
Here is the TL;DR but do read that wall of text I've prepared for you.

**TL;DR**
You're using a deprecated mechanism which could cause extra work later down the line (we'll probably eventually have to change it).
You're also giving Vue components a bunch of extra low-level knowledge since they have to tell the API client how to cancel and when; the how should be handled by the API client.

As things are, I'm fine with this staying as is, but we should seriously consider what I've thrown together in a later patch release.

I'll send this off as request changes; it's then up to you to decide what you'll do.
**EOF**


## 1. The mechanism you've opted to use is deprecated

From the [axios docs](https://axios-http.com/docs/cancellation), it's noted down that the `cancelToken` mechanism is deprecated.
Was there a specific reason you opted in to use the `cancelToken` vs. `signal` -- usage seems fairly similar to me and it'd fit into how you've designed your code just as well?

image

If possible, let's try to avoid deprecated features as much as possible to reduce tech dept as much as possible.

## 2. The integration introduces low-level complexity to the Vue component

Currently, your flow looks like this:

1. Component pings the API client and passes down some cancel bits (that token).
2. The API client makes the API request.
3. Component optionally cancels the API request directly via that mechanism.

TL;DR: your vue component asks the API client to get something as well as tells it how to abort that request.

Overall, this just adds more low-level complexity to the Vue component which, in reality, should be handled by the API client.

This works fine but I'd see a few issues:

1. You're forcing vue components to know what mechanism you're using. Especially now, when we're using a deprecated thing, we'll need to change all of them (if/when we decide to change it). 
2. You're coupling it to Axios which would cause problems if we decide to switch what we use for HTTP (unlikely but still).

TL;DR: your vue component asks the API client to get something; the API client gives us a promise as well as a way to interact with the request (cancel it).

### Possible change

#### Change the API client to provide a `cancel` closure along with the promise

```ts
async recordList (a: KV, extra: AxiosRequestConfig = {}): { response: Promise<KV>, cancel: function() } {...}
```

The `recordList` function in the API client would firstly prepare something to cancel that specific request, and then the cancel closure would invoke the cancelation.

#### Change the vue component to properly handle the new response format (the promise, closure combo)

To cancel requests from the vue component, you'd simply invoke the `cancel` closure; if we need to identify them, we can use some hashmap to index them.

### Problems with the alternate approach

1. If you change the API client function definition, you'll need to update all invocations.
This can be bypassed by either changing the function definition which is backwards-compatible, or by defining a new function (for example `recordListCancellable` which would leave the original function (`recordList`) intact.

PR #1354, opened 2023-07-20
Cast numeric datatype to text for 'like' and 'not like' filters in Postgre

### LINKED ISSUE #1305: Postgre error when trying to filter a record using 'Not equal', 'Like', 'Not like'
## Error

Image

## Reproduction steps

Image

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1305

The following PR fixes the issue with PostgreSQL's `like` and `not like` filters on numeric values.

### DIFF
--- a/server/store/adapters/rdbms/drivers/postgres/dialect.go
+++ b/server/store/adapters/rdbms/drivers/postgres/dialect.go
@@ -226,14 +226,20 @@ func (d postgresDialect) ExprHandler(n *ql.ASTNode, args ...exp.Expression) (exp
 		return drivers.OpHandlerNotIn(d, n, args...)
 
 	case "like", "nlike":
-		dalType := n.Args[0].Meta["dal.Attribute"].(*dal.Attribute).Type
-		// if the type is id (numeric) data type, then cast it to text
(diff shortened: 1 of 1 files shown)


server/store/adapters/rdbms/drivers/postgres/dialect.go
```diff
@@ -217,20 +217,24 @@ func (postgresDialect) AttributeToColumn(attr *dal.Attribute) (col *ddl.Column,
...
-		return exp.NewSQLFunctionExpression("CONCAT", aa...), nil
+		return castColumnDataToText("CONCAT", args...)
 
 	case "in":
 		return drivers.OpHandlerIn(d, n, args...)
 
 	case "nin":
 		return drivers.OpHandlerNotIn(d, n, args...)
 
+	case "like", "nlike":
```
tjerman, line 228: What about cases where the type is not ID? Should it return nothing? Is that OK?
Also, is the AttributeTypeID the only one? There are some other numeric types defined,

tjerman (commented):
Check my comment; if it's ok then this lgtm; else fix :ok_hand:

tjerman (approved):
LGTM; good job

PR #1355, opened 2023-07-24
2023.3.x fix mysql case insensitive search

### LINKED ISSUE #866: Make record search on MySQL databases case insensitive
When using MySQL and JSON encoded record values, the search function is case sensitive (all credits here go to Brian from KBR).

The issue is caused by the way MySQL handles JSON encoded strings

> "Strings produced by converting JSON values have a character set of utf8mb4 and a collation of utf8mb4_bin", "Because utf8mb4_bin is a binary collation, comparison of JSON values is case-sensitive."

Below is an example that illustrates the issue:

```sql
select

-- raw data

[a.id](http://a.id/), a.values->>'$.Number[0]' case_number


-- charset/collation coming out of json functions

, charset(a.values->'$.Number[0]') json_charset

, collation(a.values->'$.Number[0]') json_collation


-- can we match using standard like? no.

, case when a.values->>'$.Number[0]' like lower('%ticket%') then 'Yes' else 'No' end normal_match


-- can we match using collate, cast, lower, or json_search? yes.

, case when a.values->>'$.Number[0]' like lower('%ticket%') collate utf8mb4_0900_ai_ci then 'Yes' else 'No' end collate_match

, case when cast(a.values->>'$.Number[0]' as char) like lower('%ticket%') then 'Yes' else 'No' end cast_match

, case when lower(a.values->>'$.Number[0]') like lower('%ticket%') then 'Yes' else 'No' end lower_match

, case when json_search(a.values->'$.Number', 'one', lower('%ticket%') collate utf8mb4_0900_ai_ci) is not null then 'Yes' else 'No' end json_search_match


from compose_record a

where a.values->'$.Number[0]' is not null;
```

Below is the output of the example:

Image

## Todo

- [x] Pick a solution (consider benchmarking/researching which one would run faster)
- [x] Implement the fix

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/866

The following PR introduces MySQL's `like` and `not like` case-insensitive search on JSON values.

### DIFF
--- a/server/store/adapters/rdbms/drivers/mysql/dialect.go
+++ b/server/store/adapters/rdbms/drivers/mysql/dialect.go
@@ -241,7 +241,7 @@ func (d mysqlDialect) ExprHandler(n *ql.ASTNode, args ...exp.Expression) (expr e
 	case "nin":
 		return drivers.OpHandlerNotIn(d, n, args...)
 
-	case "like":
+	case "like", "nlike":
 		for a := range args {
 			args[a] = exp.NewLiteralExpression("LOWER(?)", args[a])
 		}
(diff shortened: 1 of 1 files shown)


tjerman (approved):
lgtm :ok_hand:

PR #1413, opened 2023-08-29
Remove handle field on signup

### LINKED ISSUE #1369: Invalid username and password combination on signup
### Description
Users that want to create an account some times receive the error `invalid username and password combination” “failed to authenticate with password`

Get rid of the handle input on sign up (it is generated automatically). We have duplicated checks on emails as well, we will just make sure that the auto-generated handle will not be duplicated (or at least it will not trigger an error).

petergrlica:
Handle field removed from signup page, since the user experience was not optimal, especially on duplicated handle values.
The handles are uniquely auto generated.

Closes https://github.com/cortezaproject/corteza/issues/1369

_Before_
image

_After_
Handle field is removed on the signup page.

image

### DIFF
--- a/server/auth/assets/templates/signup.html.tpl
+++ b/server/auth/assets/templates/signup.html.tpl
@@ -22,7 +22,7 @@
 				{{ tr "signup.template.form.email.label" }}
             </label>
             <input
-								data-test-id="input-email"
+				data-test-id="input-email"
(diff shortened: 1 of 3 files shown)


KinyaElGrande (approved):
LGTM 👍🏾

PR #1417, opened 2023-09-01
Fix issue with full-page navigation when show total is disabled

### LINKED ISSUE #1376: Full page navigation edge case
Properly implement full page navigation on FE and fix BE response

As you can see on the image the extra page is always there, even if there is only enough items for one page.
In cases where there are not enough items for a full new page the count is incorrect and therefore the page is not shown as clickable.

Image

- [x]  Item count on last page is incorrect
- [ ]  Return total when fullPageNav is turned on
- [x] Check if fullPage nav and count options (when turned on at the same time), do double work (count items twice),
If so, make sure it only happens once.

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1376

This PR resolves issues related to full-page navigation. It resolves cases when 'show total' is disabled, the last page has a count of 0 items and is never showed if the items on the last page are not equal to the limit, and an extra page consistently being returned in the `pageNavigation` object.

### DIFF
--- a/server/compose/dalutils/records.go
+++ b/server/compose/dalutils/records.go
@@ -200,7 +200,13 @@ func drainIterator(ctx context.Context, iter dal.Iterator, mod *types.Module, f 
 	return
 }
 
-// generatePageNavigation page nav for given record set using iterator
+// generatePageNavigation generates page navigation for a given record set using an iterator and filter limit.
+// If the limit is not defined, the page navigation will consist of only one page without a cursor.
+// If the limit is defined and is greater than the total number of records in the set,
+// the page navigation will consist of only one page without a cursor.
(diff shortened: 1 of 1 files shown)


tjerman (approved):
LGTM; just add some comments about the whole function please 🙏 (what we talked in DMs)

tjerman (approved):
lgtm ship it

tjerman (approved):
I'm not the biggest fan of cutting the mistake from `pageNavigation` -- I'd rather see the surrounding code change to exclude this edge case.

For me, it's ok as is but I'd suggest we add a little `@todo` to re-visit this function and clean it up properly -- it became a bit of a mess (my fault included :see_no_evil:)...

If it's tested and works as expected, it's mergable but do consider what I noted down above

PR #1435, opened 2023-09-12
Fix record selector, metric, chart values not refreshed when inline and bulk edit changes are applied to related records

### LINKED ISSUE #1208: Record selector value not refreshed when related record value is updated and on the same page
## Description of the bug

_When you have a text field for example in module1 and you have a record selector in a module2 pointing to that text field and you change the value of the text from module1, the record selector value won't change in module2 unless you refresh the page._

## Reproduction steps

1. Create 2 modules
2. The first module to have a text field and the second module a record selector pointing to the first module text field
3. Now go to the public page and add a text record
4. Then add a record selector record (which will be the text from the module1)
5. After that edit the text field from module 1 and notice that the record selector value is not updated unless you refresh the page

## Expected behavior 

_When you update the original value of the field the related record selector value should be also updated automatically._

## Screenshot/Video

https://user-images.githubusercontent.com/53232926/238965767-87a246ba-a842-4edc-b3a3-b9a3796c7c26.webm

## Version

Branch: 2023.3.x

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/ModuleFields/Viewer/Record.vue
+++ b/client/web/compose/src/components/ModuleFields/Viewer/Record.vue
@@ -85,8 +85,8 @@ export default {
   },
 
   beforeDestroy () {
-    this.setDefaultValues()
     this.destroyEvents()
+    this.setDefaultValues()
   },
 
   mounted () {
@@ -100,8 +100,8 @@ export default {
       resolveRecords: 'record/resolveRecords',
     }),
 
-    refreshOnRelatedModuleUpdate (module) {
-      if (this.relatedModuleID === module.moduleID) {
+    refreshOnRelatedModuleUpdate ({ moduleID }) {
+      if (this.relatedModuleID === moduleID) {
         this.formatRecordValues(this.value)
       }
     },
--- a/client/web/compose/src/components/ModuleFields/Viewer/Record.vue
+++ b/client/web/compose/src/components/ModuleFields/Viewer/Record.vue
@@ -48,6 +48,8 @@ export default {
       recordValues: {},
 
       relRecords: [],
+
+      relatedModuleID: undefined,
     }
   },
 
@@ -84,6 +86,11 @@ export default {
 
   beforeDestroy () {
     this.setDefaultValues()
+    this.destroyEvents()
+  },
+
+  mounted () {
+    this.$root.$on('module-records-updated', this.refreshOnRelatedModuleUpdate)
   },
 
   methods: {
@@ -92,6 +99,12 @@ export default {
       resolveUsers: 'user/resolveUsers',
       resolveRecords: 'record/resolveRecords',
     }),
+
+    refreshOnRelatedModuleUpdate (module) {
+      if (this.relatedModuleID === module.moduleID) {
+        this.formatRecordValues(this.value)
+      }
+    },
 
     linkToRecord (recordID) {
       if (!this.recordPage || !recordID) {
@@ -126,6 +139,8 @@ export default {
 
           const relatedModule = await this.findModuleByID({ namespaceID, moduleID: relatedField.options.moduleID })
           const relatedRecordIDs = new Set()
+
+          this.relatedModuleID = relatedModule.moduleID
 
           records.forEach(r => {
             const recordValue = relatedField.isMulti ? r.values[relatedField.name] : [r.values[relatedField.name]]
@@ -197,6 +212,11 @@ export default {
       this.processing = false
       this.recordValues = {}
       this.relRecords = []
+      this.relatedModuleID = undefined
+    },
+
+    destroyEvents () {
+      this.$root.$off('module-records-updated', this.refreshOnRelatedModuleUpdate)
     },
   },
 }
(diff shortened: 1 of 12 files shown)


client/web/compose/src/components/ModuleFields/Viewer/Record.vue
```diff
@@ -77,13 +79,19 @@ export default {
     value: {
       immediate: true,
       handler (value) {
+        this.relatedModuleID = undefined
```
Fajfa: Do we need to reset it?

client/web/compose/src/components/ModuleFields/Viewer/Record.vue
```diff
@@ -77,13 +79,19 @@ export default {
...
         this.formatRecordValues(value)
       },
     },
   },
 
   beforeDestroy () {
     this.setDefaultValues()
+    this.$root.$off('trigger-recordlist-refresh', this.refreshOnRelatedModuleUpdate)
```
Fajfa: create a destroyEvents method and do this there please

client/web/compose/src/store/record.js
```diff
@@ -46,7 +46,7 @@ export default function (ComposeAPI) {
 
         // exclude existing & make unique
         const existing = new Set(getters.set.map(({ recordID }) => recordID))
-        recordIDs = [...new Set(recordIDs.filter(recordID => !existing.has(recordID)))]
+        recordIDs = [...new Set(recordIDs)]
```
Fajfa: This will break record resolution(the whole point is we don't refetch the ones we don't have to).
You'll have to remove the updated resolution by hand for the updated records.

Clear the set on bulk edit.

client/web/compose/src/mixins/record.js
```diff
@@ -318,6 +318,10 @@ export default {
           this.fields = []
           this.record = new compose.Record(this.module, {})
           this.$emit('save')
+
+          this.$root.$emit('trigger-recordlist-refresh', {
```
Fajfa: this will have to be called something different since its not a refresh but a record update that triggers this.
So I would say you should call it something along the lines of `module-records-updated`.
You should also just emit the moduleID, if you don't need anything else.

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1076,6 +1076,24 @@ export default {
       this.$root.$on(`refetch-non-record-blocks:${pageID}`, () => {
         this.refresh(true)
       })
+
+      this.$root.$on('trigger-recordlist-refresh', this.refreshOnRelatedRecordsUpdate)
+    },
+
+    refreshOnRelatedRecordsUpdate (module) {
+      const recordFields = this.recordListModule.fields.filter((f) => f.kind === 'Record')
```
Fajfa: We discussed this, these are not the correct fields.

Fajfa (changes requested):

client/web/compose/src/components/ModuleFields/Viewer/Record.vue
```diff
@@ -194,6 +208,12 @@ export default {
       this.processing = false
       this.recordValues = {}
       this.relRecords = []
+      this.relatedModuleID = undefined
+    },
+
+    destroyEvents () {
+      this.setDefaultValues()
```
Fajfa: setDefaultValues should not be in destroyEvents, it should be after `destroyEvents` in the beforeDestroy hook

client/web/compose/src/store/record.js
```diff
@@ -38,16 +38,12 @@ export default function (ComposeAPI) {
       /**
        * Similar to fetchRecords but it only fetches unknown (not in set) ids
        */
-      async resolveRecords ({ commit, getters }, { namespaceID, moduleID, recordIDs }) {
+      async resolveRecords ({ commit, getters, state }, { namespaceID, moduleID, recordIDs }) {
```
Fajfa: You dont need getters or state here

Fajfa (changes requested):
1. All blocks that use fetchRecords (resolve), should be refreshed when a related record (with the same moduleID) is updated. If you could please check if there is any other page blocks that need this (record organizer does for example)
2. For some reason the record list refreshes twice when using bulk record edit. Guessing the saving and the emited event triggers the refresh

Fajfa (approved):
It seems good to me, maybe some comments on the bits with more code, so we know why it's there.

Fajfa (changes requested):
The double fetch still happens

Fajfa (approved):

katrinDY:
- if I have two comment blocks with the same config and add a comment in one, the other comment block isn't updated. I also have a RL pointing to the same module as the comment blocks and it's not updated after a comment was added

Other pbs I think we should update after a change has been made in a RL: (Discuss with Jože how to proceed)
- calendar
- record revision
- report - will not be addressed because it's not connected to an external moduleID
- progress bar
- geometry

Fajfa (approved):
Great job, looks good to me

katrinDY:
- RL isn't updated after a record has been added through the save as copy button from a modal => *no longer an issue*
- loading time has increased a lot when trying to add a record through a modal

katrinDY:
Loading time is fine. My bad

PR #1450, opened 2023-09-19
Add ability to determine number of records loaded at once.

### LINKED ISSUE #1403: Allow admin to enable end user to select how many rows to display in a list
Example  (e.g. display 25 rows, display 50 rows, display 100 rows, etc.)

EmmyMay:
TODO: Summary

 - [x] Code builds
 - [x] All existing tests pass

Closes https://github.com/cortezaproject/corteza/issues/1403

### DIFF
--- a/client/web/compose/src/components/PageBlocks/RecordListBase.vue
+++ b/client/web/compose/src/components/PageBlocks/RecordListBase.vue
@@ -620,24 +620,23 @@
       />
     </template>
 
-    <template
-      #footer
-    >
+    <template #footer>
       <div
         v-if="showFooter"
-        class="d-flex align-items-center justify-content-between p-2"
+        class="d-flex align-items-center flex-wrap justify-content-between p-2"
       >
-        <div class="text-truncate">
+        <div class="d-flex gap-col-3 align-items-center flex-wrap">
           <div
             v-if="options.showTotalCount"
-            class="ml-2 text-nowrap my-1"
+            class="ml-2 text-nowrap my-1 text-truncate"
           >
             <span
-              v-if="pagination.count > options.perPage"
+              v-if="pagination.count > recordsPerPage"
               data-test-id="pagination-range"
             >
               {{ $t('recordList.pagination.showing', getPagination) }}
             </span>
+
             <span
               v-else
               data-test-id="pagination-single-number"
@@ -645,11 +644,26 @@
               {{ $t('recordList.pagination.single', getPagination) }}
             </span>
           </div>
+
+          <div
+            v-if="options.showRecordPerPageOption"
+            class="d-flex align-items-center ml-2 my-1 gap-1 text-nowrap"
+          >
+            <span>
+              {{ $t('recordList.pagination.recordsPerPage') }}
+            </span>
+
+            <b-form-select
+              v-model="recordsPerPage"
+              :options="perPageOptions"
+              @change="handlePerPageChange"
+            />
+          </div>
         </div>
 
         <div
           v-if="showPageNavigation"
-          class="d-flex align-items-center justify-content-end"
+          class="d-flex align-items-center justify-content-end "
         >
           <b-pagination
             v-if="options.fullPageNavigation"
@@ -667,15 +681,19 @@
             <template #first-text>
               <font-awesome-icon :icon="['fas', 'angle-double-left']" />
             </template>
+
             <template #prev-text>
               <font-awesome-icon :icon="['fas', 'angle-left']" />
             </template>
+
             <template #next-text>
               <font-awesome-icon :icon="['fas', 'angle-right']" />
             </template>
+
             <template #last-text>
               <font-awesome-icon :icon="['fas', 'angle-double-right']" />
             </template>
+
             <template #elipsis-text>
               <font-awesome-icon :icon="['fas', 'ellipsis-h']" />
             </template>
@@ -691,6 +709,7 @@
             >
               <font-awesome-icon :icon="['fas', 'angle-double-left']" />
             </b-button>
+
             <b-button
               :disabled="!hasPrevPage || processing"
               data-test-id="previous-page"
@@ -704,6 +723,7 @@
               />
               {{ $t('recordList.pagination.prev') }}
             </b-button>
+
             <b-button
               :disabled="!hasNextPage || processing"
               data-test-id="next-page"
@@ -833,6 +853,7 @@ export default {
       selectedAllRecords: false,
 
       abortableRequests: [],
+      recordsPerPage: undefined,
     }
   },
 
@@ -847,12 +868,26 @@ export default {
     },
 
     showFooter () {
-      return this.showPageNavigation || this.options.showTotalCount
+      return this.showPageNavigation || this.options.showTotalCount || this.options.showRecordPerPageOption
+    },
+
+    perPageOptions () {
+      const defaultText = this.options.perPage === 0 ? this.$t('general:label.all') : this.options.perPage.toString()
+      return [
+        { text: defaultText, value: this.options.perPage },
+        { text: '25', value: 25 },
+        { text: '50', value: 50 },
+        { text: '100', value: 100 },
+      ].filter((v, i) => i === 0 || v.value !== this.options.perPage).sort((a, b) => {
+        if (a.value === 0) return 1
+        if (b.value === 0) return -1
+        return a.value - b.value
+      })
     },
 
     getPagination () {
       const { page = 1, count = 0 } = this.pagination
-      const { perPage = 10 } = this.options
+      const perPage = this.recordsPerPage
 
       return {
         from: ((page - 1) * perPage) + 1,
@@ -877,6 +912,10 @@ export default {
 
     showPageNavigation () {
       return this.items.length && !this.options.hidePaging
+    },
+
+    showPerPageSelector () {
+      return this.options.showRecordPerPageOption
     },
 
     disableSelectAll () {
@@ -1099,6 +1138,11 @@ export default {
       this.refresh(true)
     },
 
+    handlePerPageChange () {
+      this.filter.limit = this.recordsPerPage
+      this.refresh(true)
+    },
+
     onSaveFilterPreset (filter = []) {
       this.currentCustomPresetFilter = {
         filter,
@@ -1249,7 +1293,8 @@ export default {
     // Sanitizes record list config and
     // prepares prefilter
     prepRecordList () {
-      const { moduleID, presort, prefilter, editable, perPage, refField, positionField } = this.options
+      this.recordsPerPage = this.options.perPage
+      const { moduleID, presort, prefilter, editable, refField, positionField } = this.options
 
       // Validate props
       if (!moduleID) {
@@ -1298,7 +1343,7 @@ export default {
       }
 
       this.prefilter = filter.join(' AND ')
-      const limit = perPage
+      const limit = this.recordsPerPage
       this.filter = {
         limit,
         sort,
@@ -1594,7 +1639,7 @@ export default {
               if (!paginationOptions.incTotal) {
                 if (pages.length > 1) {
                   const lastPageCount = pages[pages.length - 1].items
-                  count = ((pages.length - 1) * this.options.perPage) + lastPageCount
+                  count = ((pages.length - 1) * this.recordsPerPage) + lastPageCount
                 } else {
                   count = records.length
                 }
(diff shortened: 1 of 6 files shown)


Fajfa (changes requested):
As discussed:
1. Make option for this in configurator
2. Move to the left side of the record list footer, and adjust the condition when footer is displayed
3. Style so you feel its good to use

EmmyMay:
The way this works is that at initial load of the RecordList, the current perPage option is the first option and it is disabled. However if there is ever any duplicate option, it is filtered out. We can't always show the initial per page option because we need to update that option to make sense of the "1-n of n records" text to the left of this. Happy to sync on this. Thanks for the review.

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -832,6 +854,12 @@ export default {
       selectedAllRecords: false,
 
       abortableRequests: [],
+      rowsOptions: [
```
Fajfa: Pretty sure you should delete this

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -849,6 +877,17 @@ export default {
       return this.showPageNavigation || this.options.showTotalCount
     },
 
+    rowOptions () {
+      return [
+        { text: this.options.perPage, value: this.options.perPage, disabled: true },
```
Fajfa: Shouldn't be disabled, that prevents you from going back to the default

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -849,6 +877,17 @@ export default {
...
+        { text: '25', value: 25 },
+        { text: '50', value: 50 },
+        { text: '100', value: 100 },
+      ]
+      // remove duplicates
+        .filter((v, i, a) => a.findIndex(t => (t.value === v.value)) === i)
```
Fajfa: You can just check if option is the same as the options.perPage, and index is not the first one
it should also be sorted, since you don't know if default option is the lowest value

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -849,6 +877,17 @@ export default {
       return this.showPageNavigation || this.options.showTotalCount
     },
 
+    rowOptions () {
```
Fajfa: to keep it consistent with the option name, you can call this `perPageOptions`

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -465,6 +467,7 @@ recordList:
     presortInputLabel: Presort
     presortPlaceholder: field1 DESC, field2 ASC
     showTotalCount: Show total record count
+    showRecordPerPageOption: Allow users to determine rows per page
```
Fajfa: Allow users to select records per page

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -626,24 +626,40 @@
         v-if="showFooter"
         class="d-flex align-items-center justify-content-between p-2"
       >
-        <div class="text-truncate">
+        <div class="d-flex gap-3 align-items-center">
```
Fajfa: It should wrap  instead of truncate

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -626,24 +626,40 @@
         v-if="showFooter"
```
Fajfa, line 625: As mentioned in previous review, you need to fix when footer is shown

lib/js/src/compose/types/page-block/record-list.ts
```diff
@@ -119,7 +120,8 @@ const defaults: Readonly<Options> = Object.freeze({
 
   bulkRecordEditEnabled: true,
   inlineRecordEditEnabled: false,
-  filterPresets: []
+  filterPresets: [],
+  showRecordPerPageOption: true,
```
Fajfa: Should be false by default, otherwise we'll impact all current record lists

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -626,24 +626,40 @@
...
+
+          <div
+            v-if="showPerPageSelector"
+            class="d-flex align-items-center gap-1 text-nowrap"
+          >
+            <span>
+              {{ $t('recordList.pagination.recordsPerPage') }}
+            </span>
+
+            <b-form-select
```
Fajfa, line 656: The select will need a permanent width, otherwise its value can be hidden if its responsive

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -626,24 +626,40 @@
...
+              v-model="filter.limit"
+              :options="rowOptions"
+              @input="updateTable"
```
Fajfa: Not sure if input is the correct event here

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -878,6 +910,10 @@ export default {
       return this.items.length && !this.options.hidePaging
     },
 
+    showPerPageSelector () {
```
Fajfa, line 917: No need for this, just use `options.showRecordPerPageOption`

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1046,6 +1082,7 @@ export default {
     if (!this.inlineEditing) {
       this.refreshBlock(this.refresh, false, true)
     }
+    this.recordsPerPage = this.options.perPage
```
Fajfa: This should be done in prepRecordList, as the name suggests, that's where all the prep happens
As you can see you're still using options.perPage in the prep method

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1100,6 +1137,11 @@ export default {
       this.refresh(true)
     },
 
+    showMoreRecords () {
```
Fajfa: can we call this something else. Show more records doesn't really describe it well.
For example `handlePerPageChange` would be more descriptive.

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -415,6 +415,8 @@ recordList:
     showing: '{{from}} - {{to}} of {{count}} records'
     single: One record
     single_plural: '{{count}} records'
+    recordsPerPage: 'Records per page:'
+    selectLabel: 'Select an option'
```
Fajfa: You dont need ' here, since it doesn't include any special characters

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -846,12 +867,23 @@ export default {
     },
 
     showFooter () {
-      return this.showPageNavigation || this.options.showTotalCount
+      return this.showPageNavigation || this.options.showTotalCount || this.options.showPerPageSelector
```
Fajfa: This is evaluating to undefined

client/web/compose/src/components/PageBlocks/RecordListBase.vue
Fajfa: This still needs to be resolved
Also, if the this.options.perPage value is 0, then label should be 'All' and be last in the array.

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -867,18 +867,34 @@ export default {
...
-      return this.showPageNavigation || this.options.showTotalCount || this.options.showPerPageSelector
+      return this.showPageNavigation || this.options.showTotalCount || this.options.showRecordPerPageOption
     },
 
     perPageOptions () {
-      return [
-        { text: this.options.perPage, value: this.options.perPage },
+      const caseZero = this.options.perPage === 0
+      let options = [
+        { text: this.options.perPage.toString(), value: this.options.perPage },
```
Fajfa: Why not just change the text if caseZero instead of slice/push?

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -867,18 +867,34 @@ export default {
...
-      // remove duplicates
-        .filter((v, i, a) => a.findIndex(t => (t.value === v.value)) === i)
+
+      if (caseZero) {
+        options = options.slice(1)
+        options.push({ text: this.$t('recordList.pagination.zero'), value: 0 })
+      } else {
+        options = options.filter((v, i, a) => {
+        // If it's the first occurrence, keep it
+          if (i === 0) return true
```
Fajfa: Don't think you need to complicate it this much. You can just filter it if its not first element and its the same as options.perPage.

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -867,18 +867,34 @@ export default {
...
+
+          // If there are previous items with the same value, eliminate the current item
+          return !a.slice(0, i).some(prevItem => prevItem.value === v.value)
+        })
+
+        // Sort the options by their value when it is all numbers
+        options.sort((a, b) => a.value - b.value)
```
Fajfa: Will this show 'All' as the last option?
If you could also chain the sort

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1292,6 +1307,7 @@ export default {
     // Sanitizes record list config and
     // prepares prefilter
     prepRecordList () {
+      this.recordsPerPage = this.options.perPage
       const { moduleID, presort, prefilter, editable, perPage, refField, positionField } = this.options
```
Fajfa: perPage is still in filter

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -416,7 +416,8 @@ recordList:
     single: One record
     single_plural: '{{count}} records'
     recordsPerPage: 'Records per page:'
-    selectLabel: 'Select an option'
+    selectLabel: Select an option
+    zero: All
```
Fajfa: all: All -> Move to general label

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -867,18 +867,36 @@ export default {
...
+      const defaultText = this.options.perPage === 0 ? this.$t('general:label.all') : this.options.perPage.toString()
+      let options = [
+        { text: defaultText, value: this.options.perPage },
         { text: '25', value: 25 },
         { text: '50', value: 50 },
         { text: '100', value: 100 },
       ]
-      // remove duplicates
-        .filter((v, i, a) => a.findIndex(t => (t.value === v.value)) === i)
+      options = options.filter((v, i) => {
```
Fajfa: You can reduce the size of all this by a lot
```
].filter((v, i) => i == 0 || v.value !== this.options.perPage).sort((a, b) => {
   if (a.value === 0) return 1
   if (b.value === 0) return -1
   return a.value - b.value
})
```

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -867,18 +867,36 @@ export default {
...
     showFooter () {
-      return this.showPageNavigation || this.options.showTotalCount || this.options.showPerPageSelector
+      return this.showPageNavigation || this.options.showTotalCount || this.options.showRecordPerPageOption
     },
 
     perPageOptions () {
-      return [
-        { text: this.options.perPage, value: this.options.perPage },
+      const defaultText = this.options.perPage === 0 ? this.$t('general:label.all') : this.options.perPage.toString()
+      let options = [
```
Fajfa: You dont need to create a let options, just return it

locale/en/corteza-webapp-compose/block.yaml
Fajfa: This is still here as i can see

Fajfa (changes requested):

Fajfa (approved):

PR #1464, opened 2023-10-03
Add spinner to c input confirm and apply necessary changes to all webapps

### LINKED ISSUE #1361: Add a reusable processing button component and apply it to appropriate places
## Description
Check if one already exists in corteza/vue

Use them in places where the buttons are disabled on click because there is processing.

## Plan of action
Add CButtonSubmit in:
- [x] compose
- [x] admin
- [x] wf
- [x] reporter
- [x] privacy
- [x] fix styles in all webapps

----

Extend CInputConfirm by adding processing after clicking on confirm btn in:
- [x] compose
- [x] admin
- [x] wf
- [x] reporter
- [x] privacy
- [x] fix styles in all webapps

katrinDY:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/admin/src/components/Apigw/Profiler/CProfilerRouteHits.vue
+++ b/client/web/admin/src/components/Apigw/Profiler/CProfilerRouteHits.vue
@@ -34,12 +34,11 @@
 
         <c-input-confirm
           :disabled="!items.length"
-          :borderless="false"
+          :processing="processingPurge"
+          :text="$t('purge.this')"
           variant="danger"
           @confirmed="purgeRequests"
-        >
-          {{ $t('purge.this') }}
-        </c-input-confirm>
+        />
       </div>
     </template>
 
@@ -127,6 +126,8 @@ export default {
 
   data () {
     return {
+      processingPurge: false,
+
       filter: {
         routeID: '',
         next: '',
@@ -254,6 +255,8 @@ export default {
     },
 
     purgeRequests () {
+      this.processingPurge = false
+
       const { routeID } = this.filter
       this.$SystemAPI.apigwProfilerPurge({ routeID })
         .then(() => {
@@ -261,6 +264,9 @@ export default {
           this.toastSuccess(this.$t('notification:gateway.profiler.purge.success'))
         })
         .catch(this.toastErrorHandler(this.$t('notification:gateway.profiler.purge.error')))
+        .finally(() => {
+          this.processingPurge = true
+        })
     },
 
     resetItems (sorting = this.sorting) {
--- a/client/web/compose/src/views/Admin/Pages/Edit.vue
+++ b/client/web/compose/src/views/Admin/Pages/Edit.vue
@@ -353,6 +353,7 @@
                       />
 
                       <c-input-confirm
+                        show-icon
                         @confirmed="removeLayout(index)"
                       />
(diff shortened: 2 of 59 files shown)


lib/vue/src/components/input/CInputConfirm.vue
```diff
@@ -13,13 +13,26 @@
...
+        <b-spinner
+          v-if="processing"
+          data-test-id="spinner"
+          class="align-middle"
+          small
+        />
+        <slot v-else>
           <font-awesome-icon
-            :icon="['far', 'trash-alt']"
+            v-if="isIconVisible"
```
Fajfa: probably makes sense to call this hideIcon, since it should be present by default.
Either that or make this true by default, and remove the prop where its not needed/add it where it is.

lib/vue/src/components/input/CInputConfirm.vue
```diff
@@ -61,39 +77,68 @@ export default {
...
+    },
 
     buttonClass: {
       type: String,
       default: '',
     },
+
+    iconClass: {
+      type: String,
+      default: '',
```
Fajfa: the default class is probably 'text-danger', or in general the one that is used the most.

client/web/admin/src/components/Apigw/Profiler/CProfilerRouteHits.vue
```diff
@@ -34,12 +34,11 @@
 
         <c-input-confirm
           :disabled="!items.length"
-          :borderless="false"
+          :processing="processingConfirm"
```
Fajfa: I would avoid calling it processingConfirm, instead it should denote what its processing.
In this case its the purge, so that would mean `processingPurge`
Same in all other places

client/web/compose/src/views/Admin/Pages/Edit.vue
```diff
@@ -353,6 +353,8 @@
                       />
 
                       <c-input-confirm
+                        v-if="layout.pageLayoutID !== '0'"
```
Fajfa: You should remove the ifs, since you can remove layouts that you just added aswell, same for layout actions below

client/web/privacy/src/components/Common/EditorToolbar.vue
```diff
@@ -38,16 +38,14 @@
         <template v-if="deleteShow">
           <c-input-confirm
             v-if="deleteConfirm"
-            :disabled="deleteDisabled || processing"
-            :borderless="false"
+            :processing="processingConfirm"
```
Fajfa: the disabled should be same as before, since it should be disabled if something else is processing or if delete is disabled

lib/vue/src/components/input/CInputConfirm.vue
```diff
@@ -13,13 +13,26 @@
         :class="`${buttonClass} ${borderless ? 'border-0' : ''}`"
         @click.stop.prevent="onPrompt"
       >
-        <slot>
+        <b-spinner
```
Fajfa, line 16: Button should also be disabled while processing, just in case

Fajfa (changes requested):

Fajfa (approved):

PR #1465, opened 2023-10-03
Add a continuous modal navigation flow for all page block

### LINKED ISSUE #1389: Properly handle all navigation (record/page) inside modal flows
Some record/page navigation in modals doesn't open the new view in the modal itself:
- Record viewers
- Calendar block click on event
- Geometry block click on event
- Others to be found

We need to keep the modal flow inside the modal itself

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/PageBlocks/CalendarBase.vue
+++ b/client/web/compose/src/components/PageBlocks/CalendarBase.vue
@@ -355,7 +355,12 @@ export default {
 
       const route = { name: 'page.record', params: { recordID, pageID: page.pageID } }
 
-      if (this.options.eventDisplayOption === 'newTab') {
+      if (this.inModal) {
+        this.$root.$emit('show-record-modal', {
+          recordID,
+          recordPageID: page.pageID,
+        })
+      } else if (this.options.eventDisplayOption === 'newTab') {
         window.open(this.$router.resolve(route).href)
       } else if (this.options.eventDisplayOption === 'modal') {
         this.$root.$emit('show-record-modal', {
--- a/client/web/compose/src/components/Public/Record/Modal.vue
+++ b/client/web/compose/src/components/Public/Record/Modal.vue
@@ -24,6 +24,7 @@
       :record-i-d="recordID"
       :values="values"
       :ref-record="refRecord"
+      :edit="edit"
       show-record-modal
       @handle-record-redirect="loadRecord"
       @on-modal-back="loadRecord"
@@ -74,6 +75,7 @@ export default {
       page: undefined,
       values: undefined,
       refRecord: undefined,
+      edit: false,
     }
   },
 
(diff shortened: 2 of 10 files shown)


client/web/compose/src/components/PageBlocks/CalendarBase.vue
```diff
@@ -355,6 +355,15 @@ export default {
...
       const route = { name: 'page.record', params: { recordID, pageID: page.pageID } }
 
+      if (this.inModal) {
+        this.$root.$emit('show-record-modal', {
+          recordID,
+          recordPageID: page.pageID,
+        })
+
+        return
+      }
```
Fajfa: Just a small thing, if you could chain this with the if below instead of returning after.
Since its one of the cases so it makes sense its if/else-if

Same for all other occurrences of this

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -24,6 +24,7 @@
       :record-i-d="recordID"
       :values="values"
       :ref-record="refRecord"
+      :in-edit-mode="modalInEditModal"
```
Fajfa: Just call it edit, should be good enough

client/web/compose/src/components/Public/Record/Modal.vue
```diff
@@ -126,10 +128,11 @@ export default {
       clearRecordIDs: 'ui/clearRecordIDs',
     }),
 
-    loadRecord ({ recordID, recordPageID, values, refRecord }) {
+    loadRecord ({ recordID, recordPageID, values, refRecord, editMode }) {
```
Fajfa: i would call all of these just edit, since it will make sense from the context what it is, same for this.modalInEditModal, can be called `edit`

client/web/compose/src/views/Public/Pages/Records/View.vue
```diff
@@ -177,6 +177,11 @@ export default {
       type: Boolean,
       required: false,
     },
+
+    inEditMode: {
```
Fajfa: Same here, just call it `edit`, since in certain edge cases, the prop will be unreliable.

Fajfa (changes requested):

katrinDY:
**Discuss with Jože if these changes should be added before starting to implement**

What I've noticed:
- after editing a record in a modal and closing it, the pb that displays the record isn't updated - RL, record organizer, calendar, geometry, chart. This can be addressed [here](https://github.com/cortezaproject/corteza/issues/1208)
- no 'x' btn in 'drill down' modal. Maybe add a bit of margin to modal title
Screenshot 2023-10-13 at 12 38 19
- no 'x' btn when chart is magnified(zoomed in)

Fajfa (approved):

katrinDY:
One small thing: when mag. modal opens for a chart, the mag. icon is present first and then replaced with the x btn. Not sure if this is an issue. Discuss with Jože, please => *update: this won't be addressed*

PR #1469, opened 2023-10-06
Add Corteza studio UI personalization feature

### LINKED ISSUE #1267: Corteza studio
1. UI/UX for styling (Studio)
2. Clean up UI to allow easier styling
3. Create a simple way to switch to dark mode.
4. Move themes to a central point on the root of repository
5. Add prop (showValue), to color picker. Which will show the color value in text next to the input

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1267

# Corteza Studio PR

This PR introduces a new feature that empowers users to personalize Corteza web applications according to their preferences using corteza main sass variables as _branding parameters_ `[white, black, primary, secondary, success, warning, danger, light,extra-light, dark, tertiary, gray-200, body-bg]`. Something to note, however, when a sass variable for example `$primary` is defined in the custom CSS editor, it will take precedence over `branding` variables.
In case of any sass-related error, the default CSS which is compiled and minified ` Bootstrap SASS, Bootstrap-Vue SASS and custom-webapp sass variables`  will be returned.

Additionally, SASS support has been integrated into the Corteza Server. However, users will need to install Dart-SASS locally or in the deployment environment, as detailed in the setup instructions provided [here](https://github.com/cortezaproject/corteza/issues/1267#issuecomment-1735272751).

### Major change introduced

Most of the SASS processing has been moved to the BE this includes; Bootstrap SASS, Bootstrap-Vue SASS, and custom webapp sass variables.

**It is important to note that this change hasn't affected Custom CSS functionality**

## Branding Editor Screenshot
Screenshot 2023-10-25 at 14 25 56

## Video demo

https://github.com/cortezaproject/corteza/assets/32476094/ad832c12-cfb4-4574-be12-a564c2073549

 - [ ] Code builds
 - [ ] All existing tests pass


lib/vue/src/components/input/CInputColorPicker.vue
```diff
@@ -1,53 +1,57 @@
...
+            />
+
+            <circle
+              cx="16"
+              cy="16"
+              r="16"
+            />
+          </svg>
+        </b-button>
+        <span v-show="showColorCodeText" class="ml-2">
```
Fajfa: I would do a v-if, also the text should look like a normal span, it looks dark here

lib/vue/src/components/navigation/CTopbar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="header-navigation d-flex flex-wrap align-items-center pr-3">
+  <div class="header-navigation d-flex flex-wrap align-items-center sticky-top mb-3 pr-3">
```
Fajfa: Rebase required, this will be removed then.

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -0,0 +1,180 @@
...
+      <h3 class="m-0">
+        {{ $t('title') }}
+      </h3>
+    </template>
+
+    <div class="row row-cols-2 mx-2">
+      <div
+        v-for="(colorInfo, key) in colorData"
+        :key="key"
+        class="p-2 border"
```
Fajfa: No need for border

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -0,0 +1,180 @@
...
+    header-bg-variant="white"
+    footer-bg-variant="white"
+  >
+    <template #header>
+      <h3 class="m-0">
+        {{ $t('title') }}
+      </h3>
+    </template>
+
+    <div class="row row-cols-2 mx-2">
```
Fajfa: If you can use b-row and b-col please, `cols="12" lg="6"`

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -0,0 +1,180 @@
...
+      <div
+        v-for="(colorInfo, key) in colorData"
+        :key="key"
+        class="p-2 border"
+      >
+        <label
```
Fajfa: this should be a b-form-group

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -0,0 +1,180 @@
...
+      type: Boolean,
+      required: true,
+    },
+  },
+
+  data () {
+    return {
+      colorData: {
+        white: {
+          label: 'White',
```
Fajfa: These require translations aswell

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -0,0 +1,180 @@
...
+    onSubmit () {
+      const brandVariables = Object.keys(this.colorData).map(key => {
+        return `$${key}: ${this.colorData[key].value}`
+      })
+
+      this.$emit('submit', { 'ui.branding.variables': brandVariables })
+    },
+
+    brandColorValue (brandingVariables, variable) {
+      const brandingVariable = brandingVariables.find(item => item.includes(`$${variable}:`))
```
Fajfa: Wouldn't it be easier to define the variable structure, so you can directly access the values?

Fajfa (changes requested):

server/assets/embed.go
```diff
@@ -67,3 +66,17 @@ func fromPath(path string) (assets fs.FS, err error) {
 
 	return
 }
+
+func FileNames(aPath string) (fileNames []string) {
```
tjerman: Minor thing but you could probably name `fileNames` to just `names`; just a thought, no need to change

server/assets/embed.go
```diff
@@ -67,3 +66,17 @@ func fromPath(path string) (assets fs.FS, err error) {
...
+	files, err := fs.ReadDir(ff, "src/"+aPath)
```
tjerman: This will only return names in the given directory; if you need to do a recursive job, you'll need to add another loop into this bad boy

server/pkg/webapp/serve.go
```diff
@@ -148,7 +151,67 @@ func serveConfig(r chi.Router, config webappConfig) {
...
+
+		// Boostrap, bootstrap-vue and custom variables scss files
+		fileNames := assets.FileNames("scss")
+		for _, fileName := range fileNames {
+			open, err := assetFiles.Open("scss/" + fileName)
+			if err != nil {
+				logger.Default().Error(fmt.Sprintf("failed to open asset %s file", fileName), zap.Error(err))
+				return
+			}
+			fileContent, err := io.ReadAll(open)
```
tjerman: I think you'll need to make a buffer over the file handle; IIRC go doesn't do that automagically and it greatly improves performance if you do it

server/pkg/webapp/serve.go
```diff
@@ -148,7 +151,67 @@ func serveConfig(r chi.Router, config webappConfig) {
 	r.Get(options.CleanBase(config.appUrl, "custom.css"), func(w http.ResponseWriter, r *http.Request) {
```
tjerman, line 152: I'm not sure I'm a big fan of us re-evaluating the SASS for each request to that custom CSS endpoint.

What I'd do is I'd create a small service (either in `pkg` or the same service we're using to manage custom SCSS) which would be responsible for re-computing the SASS as well as providing it on demand.

This would then solve 2 problems:

1. re-computing sass - CSS for each request
2. moving the evaluation logic out of here (because IMO, it doesn't belong here -- `pkg/webapp` handles the _simple logistics_ to serve webapps to the users; it shouldn't do such _compile time_ heavy lifting).

tjerman, line 152: Could we please also consider doing two additional bits?

1. make SASS compiler optional. For a lot of cases (mainly development and _light usage_), it's simply not needed and takes needles effort to set up
2. make SCSS asset path configurable to enable users uploading their own source files. Something similar we've done with auth templates -- it could come handy to users and I don't imagine this taking a lot of effort to enable.

tjerman (changes requested):
We'll need to do a bit of refactoring to clean it up and make it a bit _friendlier to use_. Also, please rebase so my changes don't show up in the diff (needles bloating).

Other then that, back-end stuff looking decent already

client/web/one/public/index.html
```diff
@@ -27,7 +27,7 @@
     <div id="app"></div>
     <!-- built files will be auto injected -->
     <style>
-      @import url("<%= BASE_URL %>custom.css");
+     @import url("http://localhost:3000/custom.css");
```
petergrlica: leftover?

client/web/compose/public/index.html
```diff
@@ -27,7 +27,7 @@
     <div id="app"></div>
     <!-- built files will be auto injected -->
     <style>
-      @import url("<%= BASE_URL %>custom.css");
+     @import url("<%= BASE_URL %>custom.css");
```
petergrlica: this should point to server url on dev env

server/pkg/webapp/serve.go
KinyaElGrande: I think we could make the SCSS asset path configurable but since we moved all the `bootstrap, boootstrap-vue and webapps variable`  SCSS files to be processed in the BE, maybe we could have a default, already compiled CSS from `[bootsrap,boootstrap-vue and variable]` SCSS that'll be served when a user hasn't installed dart-sass.

server/app/options/webapp.cue
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/cortezaproject/corteza/server/codegen/schema"
+)
+
+webapp: schema.#optionsGroup & {
+	handle: "webapp"
+
+	options: {
+		SCSS_dir_path: {
```
tjerman: I'd suggest `scss_dir_path`; all low caps

server/assets/embed.go
```diff
@@ -67,3 +66,27 @@ func fromPath(path string) (assets fs.FS, err error) {
 
 	return
 }
+
+func DirEntries(dir string) (fileNames, subDirs []string, err error) {
+	dirEntries, err := fs.ReadDir(ff, "src/"+dir)
```
tjerman: use `path.Join(...)` instead of `+` in `"src/"+dir`

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return nil
+}
+
+func defaultCSS() string {
+	var strBuilder strings.Builder
+
+	//read css contents from minified-custom.css
+	minifiedCSSFile, err := assets.Files(logger.Default(), "").Open("css/minified-custom.css")
```
tjerman: `path.Join` here also please

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,219 @@
...
+		stylesheet.set(
+			map[string]string{
+				"css":           processedCSS,
+				"branding-sass": str.HashStringSHA256(brandingSass),
+				"custom-css":    str.HashStringSHA256(customCSS),
+			},
+		)
+
+		return processedCSS, nil
+	}
```
tjerman: Prefer to use guard clauses instead of this.
The way you've done it is fine, but it's tricky to read since the if block does something special and prevents the rest of the code from executing.

Here, I would prefer to see something like this:

```go
if transpiler == nil {
  return somethingDefault(...)
}
return somethingTranspile(...)
```

or similar.

Basically, there is less going on and it's easier to understand.
Under-the-hood it's basically the same, but _my version_ is easier to understand/maintain

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if brandingSass != "" {
+		err := jsonToSass(brandingSass, &stringsBuilder)
+		if err != nil {
+			logger.Default().Error("failed to unmarshal branding sass variables", zap.Error(err))
+			return defaultCSS(), err
+		}
+	}
+
+	sassVariablesPattern := regexp.MustCompile(`(\$[a-zA-Z_-]+):\s*([^;]+);`)
```
tjerman: You can probably define this as a global since it's always the same.
Since this is all in a package it shouldn't hurt the overall codebase.

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	// compute sass content to CSS
+	args := godartsass.Args{
+		Source: stringsBuilder.String(),
+	}
+	execute, err := transpiler.Execute(args)
+	// if there's an error during compilation process, return contents from minified-custom.css
+	if err != nil {
+		logger.Default().Error("Sass syntax error", zap.Error(err))
```
tjerman: Is is really always a syntax error?

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,219 @@
...
+		)
+
+		return processedCSS, nil
+	}
+
+	if brandingSass != "" {
+		err := jsonToSass(brandingSass, &stringsBuilder)
+		if err != nil {
+			logger.Default().Error("failed to unmarshal branding sass variables", zap.Error(err))
+			return defaultCSS(), err
```
tjerman: You do this a bunch of times.
What you can do to reduce the amount of repetition here is this:

```
// ...
err = func() error {
  // .... do some work...
  if err != nil {
    return err
  }
  // ... do some more work
  return nil
}()
if err != nil {
  return defaultCSS(), err
}
```

Basically wrapping this all in a closure, doing what needs to be done and then handling the error with defaults once at the end.

tjerman (changes requested):
Some meta bits here and there

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	// if dart sass is not installed, or when the sass transpiler creation and startup process fails.
+	// return contents from default css
+	if transpiler == nil {
+		return defaultCSS(customCSS), nil
+	}
+
+	// transpile sass to css
+	err := func() error {
```
tjerman: I'd like to suggest this also gets moved to a separate function like the default CSS one.
In my opinion; you can leave it as is I won't fight you over it.

The actual code looks good to me

client/web/admin/src/components/Settings/UI/CUIBrandingEditor.vue
```diff
@@ -10,23 +10,24 @@
...
 
-    <b-row class="mx-2">
+    <b-row>
       <b-col
         v-for="key in Object.keys(brandingVariables)"
         :key="key"
-        cols="6"
+        md="6"
+        cols="12"
         class="p-2"
```
Fajfa: No need for the 'p-2'

lib/vue/src/components/input/CInputColorPicker.vue
```diff
@@ -47,10 +47,10 @@
               r="16"
             />
           </svg>
```
Fajfa: The alignment is off here aswell (whole button element is one tab too deep)

server/pkg/sass/processor.go
```diff
@@ -0,0 +1,230 @@
...
+// In case of an error, it will return default css and log out the error
+func GenerateCSS(brandingSass, customCSS, sassDirPath string) (string, error) {
+	var (
+		stringsBuilder strings.Builder
+
+		transpiler = dartSass()
+	)
+
+	//check if cache has already compiled css value and if branding-sass and custom-css have been updated
+	if !stylesheet.empty() && stylesheet.get("branding-sass") == str.HashStringSHA256(brandingSass) && stylesheet.get("custom-css") == str.HashStringSHA256(customCSS) {
```
tjerman: I'm really not a fan of you checking the values here.
You needlessly spend processing time hashing and comparing strings which could pose a bottleneck.

Why do we not move this when updating the setting in the system service package?

tjerman (changes requested):
1+1 bits left; do take a quick look please

Fajfa (changes requested):

tjerman (approved):
I think this looks decent now!

What I'd consider is making the system service not just clear out the cache, but do the actual re-computation. It'll probably not change much, but could be a bit better.

server/system/service/settings.go
```diff
@@ -246,7 +246,7 @@ func (svc *settings) BulkSet(ctx context.Context, vv types.SettingValueSet) (err
 
 	for _, v := range vv {
 		// if branding-sass or custom-css is updated, empty stylesheet cache
-		if v.Name == "ui.branding-sass" || v.Name == "ui.custom-css" {
+		if v.Name == "ui.studio.branding-sass" || v.Name == "ui.studio.custom-css" {
```
tjerman, line 249: do double check if all these references match now (since you renamed them here)

server/pkg/sass/processor.go
```diff
@@ -20,103 +20,38 @@ var (
...
-func dartSass() *godartsass.Transpiler {
-	transpiler, err := godartsass.Start(godartsass.Options{
-		DartSassEmbeddedFilename: "sass",
-	})
+// DefaultCSS contains css contents from minified-custom.css and custom-css editor content
+func DefaultCSS(customCSS string) string {
+	var strBuilder strings.Builder
 
+	//read css contents from minified-custom.css
+	minifiedCSSFile, err := assets.Files(logger.Default(), "").Open(path.Join("css", "minified-custom.css"))
```
tjerman, line 28: might be cleaner to pass in the logger via `DefaultCSS` parameters; up to you

tjerman (approved):
A few notes but lgtm as before :+1:

katrinDY:
- Getting this issue in compose:
Screenshot 2023-10-30 at 17 47 55
- Getting this is err when I open Custom CSS editor modal

- Think it's a good idea to add a way to undo a color change. Maybe add a set to default btn. If I change a color and wanna undo it, I gotta change the page and go back to UI settings. :( Discuss with Jože before hopping onto this
- Color changes in admin aren't applied to privacy, reporter, and wf
- Colors in privacy, reporter, and wf aren't the default ones
Screenshot 2023-10-30 at 18 15 07
Screenshot 2023-10-30 at 18 14 59
Screenshot 2023-10-30 at 18 14 50
- Please, rebase with .9 because the submit btn for custom css is missing. Can't submit changes without the btn

KinyaElGrande:
@katrinDY good observation I almost missed that..I've fixed that on my latest commit.

PR #1479, opened 2023-10-18
Corteza studio custom css url

petergrlica:
**What:**
 - fix custom css url on local dev env (vue cli service)
 - consistent vue config on all web clients

**How?**
Vue cli config fetches the `public/config.js` file and evals the contents to get the `window.CortezaAPI` value and uses it to set the proxying rule in the hotplug config.


Fajfa (approved):
Based on my untrained eye, looks good

KinyaElGrande:
It looks good to me.
Tested it and works as expected.

KinyaElGrande (approved):

PR #1516, opened 2023-11-09
Add reusable map component to vue

### LINKED ISSUE #1392: Make a reusable map component and standardize all options
Currently, we have multiple implementations of the map and markers.
We need to make one component that we use all over the place

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass


lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
+<template>
```
Fajfa, line 1: You can place this inside a div, that includes the geosearch container and the map itself.

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+<l-map
+  ref="map"
+  :zoom="map.zoom"
+  :center="map.center"
+  :min-zoom="map.zoomMin"
+  :max-zoom="map.zoomMax"
+  :bounds="map.bounds"
+  :max-bounds="map.maxBounds"
+  style="height: 75vh; width: 100%; cursor: pointer;"
```
Fajfa: You can remove the style implementation.

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+  @update:bounds="onBoundsUpdate"
+>
+  <l-tile-layer
+    url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+    :attribution="map.attribution"
+  />
+
+  <l-polygon
+    v-for="(geometry, i) in polygonGeometries"
+    :key="`polygon-${i}`"
```
Fajfa: For consistency you can just keep the key as `i`

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+
+  <l-marker
+    v-for="(marker, i) in markers"
+    :key="i"
+    :lat-lng="marker"
+    :opacity="localValueIndex === undefined || i == localValueIndex ? 1.0 : 0.6"
+    @click="onRemoveMarker(i)"
+  />
+
+  <slot name="markers"></slot>
```
Fajfa: Instead of having this everything should be handled through markers prop

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+    :lat-lngs="geometry.map(value => value.geometry)"
+    :color="polygonColors[i]"
+  />
+
+  <l-marker
+    v-for="(marker, i) in markers"
+    :key="i"
+    :lat-lng="marker"
+    :opacity="localValueIndex === undefined || i == localValueIndex ? 1.0 : 0.6"
+    @click="onRemoveMarker(i)"
```
Fajfa: I would implement onMarkerClick and handle this and the emit in that method

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+    :key="`polygon-${i}`"
+    :lat-lngs="geometry.map(value => value.geometry)"
+    :color="polygonColors[i]"
+  />
+
+  <l-marker
+    v-for="(marker, i) in markers"
+    :key="i"
+    :lat-lng="marker"
+    :opacity="localValueIndex === undefined || i == localValueIndex ? 1.0 : 0.6"
```
Fajfa: How about the :icon prop

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+import CInputSearch from '../input/CInputSearch.vue'
+
+export default {
+  props: {
+    hideCurrentLocationButton: {
+      type: Boolean,
+      default: false
+    },
+    
+    label: {
```
Fajfa: labels please

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+    map: {
+      type: Object
+    },
+    
+    markers: {
+      type: Array,
+      default: () => ([])
+    },
+
+    localValueIndex:{
```
Fajfa: This will have to be handled differently. Or at least named differently so name conveys the meaning

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+      type: Number,
+      default: undefined
+    },
+
+    styles: {
```
Fajfa: Rename this to mapStyle

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+      default: () => ([])
+    },
+
+    polygonColors: {
+      type: Array,
+      default: () => ([])
+    },
+  },
+
+  components: { LControl, CInputSearch, LPolygon },
```
Fajfa: Expand this for easier maintaining

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+  
+  data () {
+    return {
+      geoSearch: {
```
Fajfa, line 157: As mentioned above the template for geosearch should also be in this component, and it should emit an event with the Promise, so we can handle it where its used aswell

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+        query: '',
+        provider: new OpenStreetMapProvider(),
+        results: [],
+      },
+    }
+  },
+
+  mounted() {
+    this.$emit('on-map-ref', this.$refs.map);
```
Fajfa: We wont need refs where we're going

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+
+    onCenter (e) {
+      this.$emit('on-center', e)
+    },
+
+    onBoundsUpdate (e) {
+      this.$emit('on-bounds-update', e)
+    }
+  },
+};
```
Fajfa: No semicolons please

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+      type: Number,
+      default: undefined
+    },
+
+    styles: {
+      type: Object,
+      default: () => ({})
+    },
+
+    polygonGeometries: {
```
Fajfa: Can we call this polygons and package the colors with them? so we don't need two props

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+      type: Boolean,
+      default: false
+    },
+    
+    label: {
+      type: Object,
+      required: true
+    },
+    
+    map: {
```
Fajfa, line 129: If we can have a computedProperty that merges this with default options
`mapOptions ()`

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -313,7 +296,7 @@ export default {
       this.map.show = true
 
       setTimeout(() => {
-        this.$refs.map.mapObject.invalidateSize()
+        this.mapRef.mapObject.invalidateSize()
```
Fajfa: To avoid ref here we can instead have a watcher on map object inside the CMap, and we can invalidate the size when that changes

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -387,19 +370,19 @@ export default {
     },
 
     goToCurrentLocation () {
-      this.$refs.map.mapObject.locate()
+      this.mapRef.mapObject.locate()
```
Fajfa: This can be handled inside the map

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -387,19 +370,19 @@ export default {
...
     },
 
     onLocationFound ({ latitude, longitude }) {
-      const zoom = this.$refs.map.mapObject._zoom >= 15 ? this.$refs.map.mapObject._zoom : 15
+      const zoom = this.mapRef.mapObject._zoom >= 15 ? this.mapRef.mapObject._zoom : 15
```
Fajfa: This can also be handled inside the map, except for placing of the marker

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -387,19 +370,19 @@ export default {
...
       this.placeMarker({ latlng })
-      this.$refs.map.mapObject.flyTo([latitude, longitude], zoom)
+      this.mapRef.mapObject.flyTo([latitude, longitude], zoom)
     },
 
     placeGeoSearchMarker (result) {
-      const zoom = this.$refs.map.mapObject._zoom >= 15 ? this.$refs.map.mapObject._zoom : 15
-      this.$refs.map.mapObject.flyTo([result.latlng.lat, result.latlng.lng], zoom, { animate: false })
+      const zoom = this.mapRef.mapObject._zoom >= 15 ? this.mapRef.mapObject._zoom : 15
+      this.mapRef.mapObject.flyTo([result.latlng.lat, result.latlng.lng], zoom, { animate: false })
```
Fajfa: Same here

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -429,6 +412,10 @@ export default {
       this.map = {}
       this.geoSearch = {}
     },
+
+    onMapRefEmit (map) {
```
Fajfa: As mentioned we wont need this

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -257,6 +238,8 @@ export default {
         provider: new OpenStreetMapProvider(),
```
Fajfa: This can all be moved inside CMap

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+  ref="map"
+  :zoom="map.zoom"
+  :center="map.center"
+  :min-zoom="map.zoomMin"
+  :max-zoom="map.zoomMax"
+  :bounds="map.bounds"
+  :max-bounds="map.maxBounds"
+  style="height: 75vh; width: 100%; cursor: pointer;"
+  :style="styles"
+  @click="placeMarker"
```
Fajfa: This should be onMapClick, the actual marker placing needs to be handled where the c-map is used
The CMap only shows stuff, it shouldn't manipulate the markers/value

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -180,51 +180,32 @@
...
+        :map="map"
+        :on-location-found="onLocationFound"
+        :hide-current-location-button="!field.options.hideCurrentLocationButton"
+        :local-value-index="localValueIndex"
+        :markers="markers"
+        :label="{
+          tooltip: { 'goToCurrentLocation': $t('tooltip.goToCurrentLocation') }
+        }"
+        @on-go-to-current-location="goToCurrentLocation"
+        @on-place-marker="placeMarker($event)"
```
Fajfa: this should be on-map-click="placeMarker"
Also you dont need the $event if you don't use it with (), so `@on-map-click="placeMarker""

Fajfa (changes requested):
Map should only show and not modify the values, we should find a way to do it without refs aswell.
The geosearch needs to be moved inside the CMap aswell

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,151 @@
...
+    label: {
+      type: Object,
+      required: true
+    },
+    
+    map: {
+      type: Object
+    },
+    
+    markers: {
```
Fajfa, line 134: Watch this, and when it changes fly to the last marker

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+
+      if (markerColor) {
+        html = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 34.892337" height="60" width="40" style="margin-top: -40px;margin-left: -15px;height: 35px;">
```
Fajfa: Cant we just use this always and default the marker color?

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+      <l-polygon
+        v-for="(geometry, i) in polygons"
+        :key="i"
+        :lat-lngs="geometry.map(value => value.geometry)"
+        :color="geometry[0].color"
+      />
+    
+      <l-marker
+        v-for="(marker, i) in markers"
+        :key="`marker-${i}`"
```
Fajfa, line 60: Same here you can just use `i` as key, since only if its a nested v-for you should worry about keys

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+      default: false
+    },
+    
+    labels: {
+      type: Object,
+      required: true
+    },
+    
+    map: {
+      type: Object
```
Fajfa: the default is still needed here `default: () => ({})`, and any other Object prop

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+      this.geoSearch.provider.search({ query }).then(results => {
+        this.geoSearch.results = results.map(result => ({
+          ...result,
+          latlng: {
+            lat: result.raw.lat,
+            lng: result.raw.lon,
+          },
+        }))
+      }).catch(() => {
+        this.$emit('trigger-geosearch-error')
```
Fajfa: this should be called `on-geosearch-error`.
In general the reusable component doesn't know how the parent will handle the event, so the name should just indicate when the event is triggered.

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+    disableMap () {
+      if (this.disable) this.$refs.map.mapObject._handlers.forEach(handler => handler.disable())
+    },
+
+    enableMap () {
+      if (this.disable) this.$refs.map.mapObject._handlers.forEach(handler => handler.enable())
+    },
+
+    onMarkerClick(index, marker) {
+      this.$emit('on-marker-marker', { index, marker })
```
Fajfa: Probably `on-marker-click`

lib/vue/src/components/map/CMap.vue
```diff
@@ -1,57 +1,84 @@
...
+        :lat-lng="marker"
+        :icon="getIcon(marker.color)"
+        :opacity="activeMarkerIndex === undefined || i == activeMarkerIndex ? 1.0 : 0.6"
+        @click="onMarkerClick(i, marker)"
+      />
+    
+      <l-control class="leaflet-bar">
+        <a
+          v-if="!hideCurrentLocationButton"
+          :title="labels.tooltip && labels.tooltip.goToCurrentLocation"
```
Fajfa: Probably worth it to use a tooltip here so you don't have to do it on the other issue

lib/vue/src/components/map/CMap.vue
```diff
@@ -81,28 +108,37 @@ export default {
       default: () => ([])
     },
 
-    localValueIndex:{
+    activeMarkerIndex:{
       type: Number,
       default: undefined
```
Fajfa: If its not required, undefined will be the default anyway so this line isn't needed

lib/vue/src/components/map/CMap.vue
```diff
@@ -81,28 +108,37 @@ export default {
...
 
-    polygonColors: {
-      type: Array,
-      default: () => ([])
+    hideGeoSearch: {
+      type: Boolean,
+      default: false
     },
+
+    disable: {
```
Fajfa: should be called disabled

lib/vue/src/components/map/CMap.vue
```diff
@@ -81,28 +108,37 @@ export default {
...
+      type: Boolean,
+      default: false
+    }
   },
 
-  components: { LControl, CInputSearch, LPolygon },
+  components: {
```
Fajfa: components should be above the props

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -156,56 +156,30 @@
...
-      </div>
-
       <c-map
         ref="c-map"
         :field="field"
         :map="map"
-        :on-location-found="onLocationFound"
-        :hide-current-location-button="!field.options.hideCurrentLocationButton"
-        :local-value-index="localValueIndex"
+        :hide-geo-search="!field.options.hideGeoSearch"
```
Fajfa: No need to negate the value here, otherwise its gonna be the opposite of what you want to do

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -369,52 +326,14 @@ export default {
...
-          ...result,
-          latlng: {
-            lat: result.raw.lat,
-            lng: result.raw.lon,
-          },
-        }))
-      }).catch(() => {
-        this.toastErrorHandler(this.$t('notification:field-geometry.geolocationErrors.locationSearchFailed'))()
-      })
+    triggerGeoSearchError () {
```
Fajfa: onGeoSearchError

client/web/compose/src/components/ModuleFields/Configurator/Geometry.vue
```diff
@@ -19,41 +19,27 @@
...
-          >
-            <font-awesome-icon
-              :icon="['fas', 'location-arrow']"
-              class="text-primary"
-            />
-          </a>
-        </l-control>
-      </l-map>
+      <c-map
+        ref="c-map"
```
Fajfa: Probably no need for a ref here anymore, same in all other places except CMap

client/web/compose/src/components/ModuleFields/Viewer/Geometry.vue
```diff
@@ -24,62 +24,19 @@
...
-        </l-control>
-      </l-map>
+      <c-map
+        ref="c-map"
+        :field="field"
+        :map="map"
+        :active-marker-index="localValueIndex"
+        :markers="localValue"
+        :hide-current-location-button="field.options.hideCurrentLocationButton"
+        :hide-geo-search="!field.options.hideGeoSearch"
```
Fajfa: Same here, you shouldn't negate it

client/web/compose/src/components/PageBlocks/GeometryBase.vue
```diff
@@ -131,6 +103,16 @@ export default {
 
       return values
     },
+
+    polygons () {
```
Fajfa: Do we need this?

client/web/compose/src/components/PageBlocks/GeometryBase.vue
```diff
@@ -14,76 +14,48 @@
...
+        :map="{
+          ...map,
+          maxBounds: map.bounds
+        }"
+        :labels="{
+          tooltip: { 'goToCurrentLocation': $t('geometry.tooltip.goToCurrentLocation') }
+        }"
+        :markers="localValue"
+        :disable="editable"
+        :hide-geo-search="true"
```
Fajfa: you dont need to say `hide-geo-search="true"`. If you just do `hide-geo-search` it will be true

client/web/compose/src/components/ModuleFields/Editor/Geometry.vue
```diff
@@ -156,75 +156,30 @@
...
+        @on-map-click="placeMarker"
+        @on-marker-marker="removeMarker"
+        @trigger-geosearch-error="triggerGeoSearchError"
+      />
     </b-modal>
   </b-form-group>
 </template>
 <script>
 import base from './base'
 import { latLng } from 'leaflet'
```
Fajfa: Is there any way we can do this inside the CMap, if so we then can remove all imports from the webapp relating to the map

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+        v-for="(geometry, i) in polygons"
+        :key="i"
+        :lat-lngs="geometry.map(value => value.geometry)"
+        :color="geometry[0].color"
+      />
+    
+      <l-marker
+        v-for="(marker, i) in markers"
+        :key="`marker-${i}`"
+        :lat-lng="marker"
```
Fajfa: I would keep this as marker.value, and pass it through latLng

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/GeometryBase.vue
```diff
@@ -118,7 +90,7 @@ export default {
             value.markers.map(subValue => {
               if (this.getLatLng(subValue)) {
                 values.push({
-                  value: this.getLatLng(subValue) || {},
+                  ...(this.getLatLng(subValue) || {}),
```
Fajfa: Parent components shouldnt use latlng, the eval should be in the CMap itself.
We should have a consistent structure for markers, at least the .value part.

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
...
+        ...this.map
+      }
+    }
+  },
+
+  watch: {
+    mapOptions: {
+      handler (map) {
+        if (map.show && this.$refs.map) {
+          this.$refs.map.mapObject.invalidateSize()
```
Fajfa: Make sure that map fits block/modal on render. 
If setTimeout is not needed even better

lib/vue/src/components/map/CMap.vue
```diff
@@ -0,0 +1,307 @@
+<template>
+  <div class="position-relative" v-if="mapOptions.show">
```
Fajfa:
```
<div
 v-if
 class
 >
 ```

File: lib/vue/src/components/map/CMap.vue [hunk 45]
```diff
@@ -0,0 +1,307 @@
...
+    map: {
+      type: Object
+    },
+    
+    markers: {
+      type: Array,
+      default: () => ([])
+    },
+
+    activeMarkerIndex:{
```
Fajfa: space between :{

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -793,6 +793,7 @@ geometry:
     lowerRight: Bounds lower right
   tooltip:
       goToCurrentLocation: Go to current location
+      geosearchInputPlaceholder: Enter address, place or postcode
```
Fajfa: Remove this

client/web/compose/src/components/PageBlocks/GeometryConfigurator/Configurator.vue
```diff
@@ -1,39 +1,19 @@
...
-        </l-control>
-      </l-map>
+      <c-map
+        ref="c-map"
+        :map="mapOptions"
+        :labels="{
+          tooltip: { 'goToCurrentLocation': $t('geometry.tooltip.goToCurrentLocation') },
+          geosearchInputPlaceholder: $t('geometry.tooltip.geosearchInputPlaceholder')
+        }"
+        map-style="height: 45vh;"
```
Fajfa: you should probably hide-geo-search

Fajfa (changes requested):

Fajfa (changes requested):
Eeverything looks good to me, can you just double check every map related library is only in corteza-vue?
After that it will be time to cover other map instances in corteza(privacy, discovery, CLocation in corteza-vue)

client/web/admin/src/components/CLocation.vue
```diff
@@ -19,33 +19,29 @@
...
-        <l-marker
-          v-if="value && value.length"
-          :lat-lng="getLatLng(value)"
-          @click="removeMarker"
-        />
-      </l-map>
+      <c-map
+        map-style="height: 75vh; width: 100%;"
+        :map="map"
+        hide-geo-search
```
Fajfa: im thinking we can enable geo-search and current-location on all maps that it makes sense on

client/web/admin/package.json
```diff
@@ -34,6 +34,7 @@
     "fstream": "^1.0.12",
     "kind-of": "^6.0.3",
     "leaflet": "^1.7.1",
+    "leaflet-geosearch": "^3.7.0",
```
Fajfa: leaflet, vue-leaflet, leaflet-geosearch should all be only on lib/vue

client/web/admin/src/components/CLocation.vue
```diff
@@ -70,21 +66,15 @@ export default {
     }
   },
 
+  computed: {
```
Fajfa: Is this needed?

client/web/admin/src/components/CLocation.vue
```diff
@@ -19,33 +19,29 @@
...
-          url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
-          :attribution="map.attribution"
-        />
-        <l-marker
-          v-if="value && value.length"
-          :lat-lng="getLatLng(value)"
-          @click="removeMarker"
-        />
-      </l-map>
+      <c-map
```
Fajfa, line 22: Did you test the usage of this? To me it doesn't render correctly

client/web/discovery/src/components/DiscoveryMap.vue
```diff
@@ -1,29 +1,28 @@
...
-      @click="onMarkerClick(marker.id)"
-    />
-  </l-map>
+  <c-map
+    map-style="height: calc(100vh - 64px);"
+    :map="{
+      zoom,
+      center,
+    }"
+    hide-geo-search
```
Fajfa: Same here, both props can be removed (hide-geo-search, hide-current-location-button)
Make sure they work aswell

client/web/discovery/src/components/DiscoveryMap.vue
```diff
@@ -1,29 +1,28 @@
...
+    hide-current-location-button
+    :markers="[...makerValues]"
```
Fajfa: Do we need to do the `[...` ?

client/web/discovery/src/components/DiscoveryMap.vue
```diff
@@ -61,23 +72,15 @@ export default {
...
-    getLatLng (coordinates = [0, 0]) {
-      const [lat, lng] = coordinates
-
-      if (isNumber(lat) && isNumber(lng)) {
-        return latLng(lat, lng)
-      }
-    },
-
-    onMarkerClick (ID) {
+    onMarkerClick ({ index: ID }) {
```
Fajfa: I understand why you did this, but its probably better to just use index in this case, since that's what it is.

client/web/privacy/src/components/ConnectionMap.vue
```diff
@@ -1,61 +1,47 @@
 <template>
-  <l-map
-    :zoom="zoom"
+  <c-map
+    :map="{
+      zoom,
+    }"
+    hide-geo-search
+    hide-current-location-button
+    :markers="[...markerValues]"
```
Fajfa: Same here is the `[...` needed?

client/web/privacy/src/components/ConnectionMap.vue
```diff
@@ -1,61 +1,47 @@
...
-import { isNumber } from 'lodash'
-import { LMap, LTileLayer, LMarker, LTooltip } from 'vue2-leaflet'
+import { components } from '@cortezaproject/corteza-vue'
+
+const { CMap } = components
 
 export default {
-  i18nOptions: {
-    namespaces: 'map',
+  components: {
```
Fajfa: components should be under i18nOptions, weird that the linter didn't give an error

client/web/privacy/src/components/ConnectionMap.vue
```diff
@@ -84,26 +67,28 @@ export default {
         return coordinates && !!coordinates.length
       })
     },
+
+    markerValues () {
```
Fajfa: Instead of having two computeds you can rename the `validMarkers` into markerValues and do the mapping there aswell. Since `validMarkers` is only used in this computed.

client/web/privacy/src/components/ConnectionMap.vue
```diff
@@ -1,61 +1,47 @@
 <template>
-  <l-map
-    :zoom="zoom"
+  <c-map
```
Fajfa, line 2: If you compare how this looks locally and on the previous version you'll see the styles are a bit different

lib/vue/src/components/map/CMap.vue
```diff
@@ -62,9 +62,22 @@
         :key="`marker-${i}`"
         :lat-lng="marker.value"
         :icon="getIcon(marker.color)"
```
Fajfa, line 62: Im thinking that if you style the map to always expand to the parent div, you'll have a much easier time styling it for each use case. Might be worth a few minutes of trying it

Fajfa (changes requested):

client/web/admin/src/components/CLocation.vue
kelanik8: a mistake was debugging the issue with the values being updated from the map, but I've figured that out now

lib/vue/src/components/map/CMap.vue
```diff
@@ -307,6 +303,12 @@ export default {
 </script>
 
 <style scoped>
+.c-map-container {
+  height: 75vh;
```
Fajfa: Not sure about this, why do we have to have a height defined in the component.
Since the view using it is what should define the height.
The map should fill the space that the parent component provides.

lib/vue/src/components/map/CMap.vue
```diff
@@ -1,9 +1,8 @@
 <template>
   <div 
     v-if="mapOptions.show"
```
Fajfa: Shouldn't the map itself be hidden and not the parent element itself?
If its use case specific (it only happens on one place) then that view should have the v-if on the component itself.

Fajfa (changes requested):

Fajfa (approved):

katrinDY:
- full card not rendered when viewing/editing a record in a module


- full card doesn't always display when switching from page builder to the public view


- the current location button not working when editing/viewing records and in geo block config


- issue after switching from single to multi-values

Screenshot 2023-12-06 at 14 21 31

Fajfa (approved):

PR #1571, opened 2023-11-29
Added error wrapping for compose record wf functions

### LINKED ISSUE #1430: Return more info on failure while persisting records in workflows
When a user includes invalid or missing required values in record input in a workflow function, the error is displayed in server logs.

The error is not verbose enough, only the `invalid record value input` is presented to the user, along with the usual meta data on workflow and workflow step.

There needs to be more info presented, mainly the invalid field that is causing the error.

**Workflow error:**
```
15:54:19.674	WARN	workflow.session.exec	executed with errors	{"workflowID": "111", "runnerID": "111", "runnerRoles": [], "sessionID": "354354363533099009", "stateID": "354354363550007297", "stepID": "6", "functionRef": "composeRecordsCreate", "functionKind": "function", "execTime": "33.358159ms", "error": "invalid record value input"}
``` 

**API error (with more info):**
```
{"error":{"message":"1 issue(s) found","details":[{"kind":"invalidValue","message":"Invalid field value","meta":{"field":"LT","id":"354353949203103745","value":"new1"}}]}}
```

petergrlica:
Workflow record `create` and `update` functions have been wrapped with extra meta data when the values are invalid. It improves upon workflow developing, since before it was impossible to get info on which values were invalid for a specific module when a record was created.

It uses the `Safe()` method which is a stub for future secure values support, for now it always displays data.

Before
image

After
image

### DIFF
--- a/server/compose/automation/records_handler.go
+++ b/server/compose/automation/records_handler.go
@@ -2,7 +2,11 @@ package automation
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/cortezaproject/corteza/server/compose/types"
 	. "github.com/cortezaproject/corteza/server/pkg/expr"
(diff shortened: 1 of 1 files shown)


KinyaElGrande (approved):
LGTM

ghost:
Thanks for addressing this one. No more workarounds / stripping down steps to see what is causing it.

PR #1579, opened 2023-12-01
Add support for dark mode in Corteza studio

### LINKED ISSUE #1494: Corteza Studio v2 (Dark mode)
Implement dark mode, and extra configuration options for studio, alongside possible fixes(if needed).

### Todos
- [x] Ensure that dart-sass installation is not checked  on every request on `custom.css`.
- [x] Add an option to return to original setting 
- [x] Add a switch to light/dark mode
- [x] Add tabs for the themes, 2 for branding [light, dark] and 3 for Custom CSS [`general, light, dark`].
- [x] Add support for light/dark theme

KinyaElGrande:
# Dark mode / Multi-theme Support

Ref: https://github.com/cortezaproject/corteza/issues/1494

Add support for dark-mode theme and multiple themes on corteza studio.

### Additional Changes

1.  Dart-sass installation is checked on server boot up, instead on every request to `/custom.css`
2. FE option to return to default color setting on color-picker modal
3. CSS generation on server boot up.
4. `custom-css` has been provisioned to `custom-css.general` and `branding-sass` to both theme `light` and `dark` themes.


server/app/boot_levels.go
```diff
@@ -957,6 +966,20 @@ func updateSmtpSettings(log *zap.Logger, current *types.AppSettings) {
 	setupSmtpDialer(log, current.SMTP.Servers...)
 }
 
+func updateSassInstallSettings(ctx context.Context, log *zap.Logger) {
+	var sassInstalled bool
+
+	if sass.DartSassTranspiler(log) != nil {
+		sassInstalled = true
+	}
```
tjerman:
```suggestion
	sassInstalled :=  sass.DartSassTranspiler(log) != nil
```

<3

server/pkg/provision/stylesheet.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"encoding/json"
+	"github.com/cortezaproject/corteza/server/store"
+	"github.com/cortezaproject/corteza/server/system/types"
+	"go.uber.org/zap"
+	"strconv"
+	"unicode"
+)
+
+func updateWebappTheme(ctx context.Context, log *zap.Logger, s store.Storer) (err error) {
```
tjerman, line 16: It's a bit difficult to read this function. Please clean it up and add some notes on what exactly the point of it is.

server/pkg/provision/stylesheet.go
```diff
@@ -0,0 +1,97 @@
...
+					continue
+				}
+
+				themes = append(themes, types.Theme{
+					ID:     themeID,
+					Title:  string(title),
+					Values: "",
+				})
+				continue
+			}
```
tjerman: Why does this need to be in a loop? It'd be much easier to clean to have this case out side of the loop

server/system/rest.yaml
```diff
@@ -572,6 +572,10 @@ endpoints:
         name: labels
         title: Labels
         parser: label.ParseStrings
+      - name: meta
```
tjerman, line 579: Should the meta also go into the user create endpoint?

server/system/service/stylesheet.go
```diff
@@ -17,48 +18,160 @@ import (
...
 	)
 
-	//check if cache has already compiled css value
-	if !sass.StylesheetCache.Empty() {
-		return sass.StylesheetCache.Get("css"), nil
+	for _, customCSS := range studio.CustomCSS {
+		customCSSMap[customCSS.ID] = customCSS.Values
+	}
+
+	sass.DefaultCSS(log, customCSSMap["general"])
```
tjerman: These `general` strings are a bit spooky.
Can we have some global constant for these fixed theme names instead?

server/system/service/stylesheet.go
```diff
@@ -17,48 +18,160 @@ import (
...
+		customCSS := processCustomCSS(themeID, themeCustomCSS)
+		err := sass.Transpile(transpiler, log, themeID, themeSASS, customCSS, sassDirPath)
+		if err != nil {
+			log.Error("failed to transpile sass to css", zap.Error(err))
+		}
 	}
 
-	updateSassInstalledSetting(ctx, log, true)
+	for key := range currentThemesMap {
+		if str.HashStringSHA256(oldThemesMap[key]) == str.HashStringSHA256(currentThemesMap[key]) {
```
tjerman: I'd suggest using a faster algorithm. The only thing that matters here is if the bits match up or not. Generally this needs to be done as fast as possible so things like algorithm security doesn't matter.

server/system/service/stylesheet.go
```diff
@@ -17,48 +18,160 @@ import (
...
 
-	if err != nil {
-		return sass.DefaultCSS(log, customCSS), err
+			for themeID := range complimentaryStylesMap {
+				transpileSASS(themeID, complimentaryStylesMap[themeID], currentThemesMap)
+			}
+			continue
+		}
+
+		transpileSASS(key, complimentaryStylesMap[key], currentThemesMap)
```
tjerman, line 122: why are all these `transpileSASS` outputs ignored?

server/system/types/user.go
```diff
@@ -163,3 +164,13 @@ func (u *User) Clone() *User {
...
 func (meta *UserMeta) Value() (driver.Value, error) { return json.Marshal(meta) }
+
+func ParseUserMeta(ss []string) (p *UserMeta, err error) {
+	p = &UserMeta{}
+
+	if len(ss) == 0 {
+		return
+	}
+
+	return p, json.Unmarshal([]byte(ss[0]), p)
```
tjerman:
```suggestion
	return p, json.Unmarshal([]byte(ss[0]), &p)
```

Shouldn't the Unmarshal fnc get the pointer?

server/system/service/stylesheet.go
```diff
@@ -17,48 +18,160 @@ import (
 // If dart isn't installed on the host machine, customCustom css will continue to function, but without sass support.
 //
 // In case of an error, it will return default css and log out the error
-func GenerateCSS(ctx context.Context, brandingSass, customCSS, sassDirPath string) (string, error) {
+func GenerateCSS(settings *types.AppSettings, sassDirPath string) (err error) {
 	var (
-		log        = logger.Default()
-		transpiler = sass.DartSass(log)
+		log          = logger.Default()
```
tjerman: Should this get the logger from the caller?

tjerman (changes requested):
I reviewed it on the code level basis; the structure and stuff, is a bit too much to comprehend. We can take a look at it together at some later point.
So far so good but I'm finding it a bit confusing.

server/system/service/stylesheet.go
KinyaElGrande: you're right, I think `xxhash` will do the job

server/system/service/stylesheet.go
KinyaElGrande: `transpileSASS` is an inner function that doesn't return anything.
https://github.com/cortezaproject/corteza/blob/a9d72137697d2c4a67941d62835db0bb959178b9/server/system/service/stylesheet.go#L100-L106

tjerman: ah.. i see

server/system/service/stylesheet.go
```diff
@@ -1,64 +1,176 @@
...
-	if err != nil {
-		return sass.DefaultCSS(log, customCSS), err
+				customCSS := processCustomCSS(key, customCSSMap)
+				// transpile sass to css
+				err := sass.Transpile(transpiler, log, key, "", customCSS, sassDirPath)
+				if err != nil {
+					continue
+				}
+			}
+		}
```
tjerman: Might be worth splitting this into some functions just to reduce the amount if indentation

tjerman (approved):
Looks ok to me; good job man

PR #1580, opened 2023-12-04
Fix #1486

### LINKED ISSUE #1486: When try to build Corteza server in windows environemnt , build system fails because of error in users.go

### Version of Corteza

2023.3.7

### Current Behavior

following error rise when running Make Watch :
 cannot use syscall.Stdin (variable of type "syscall".Handle) as int value in argument to terminal.ReadPassword

### Expected Behavior

_No response_

### Steps To Reproduce

set your desired config for DB-DSN and other things then run "make watch" in command prompt .
you will see following error :
` cannot use syscall.Stdin (variable of type "syscall".Handle) as int value in argument to terminal.ReadPassword`

### Environment and versions

```markdown
windows (10)
Go 1.19
corteza 2023.3.x
```

### Anything else?

change file "user.go" file in following path : corteza\server\system\commands to : 
`if password, err = terminal.ReadPassword(int(os.Stdin.Fd())); err != nil {`
and error will gone.

moalamri:
`terminal.ReadPassword` requires an int.
Passing a string will make the building process fail on Windows.
After changing the type, I ran `make watch` and it compiled successfully.

 - [x] Code builds
 - [x] All existing tests pass

### DIFF
--- a/server/system/commands/users.go
+++ b/server/system/commands/users.go
@@ -118,7 +118,7 @@ func Users(ctx context.Context, app serviceInitializer) *cobra.Command {
 
 			if !flagNoPassword && !flagMakePasswordLink && len(password) == 0 {
 				cmd.Print("Set password: ")
-				if password, err = terminal.ReadPassword(syscall.Stdin); err != nil {
+				if password, err = terminal.ReadPassword(int(syscall.Stdin)); err != nil {
 					cli.HandleError(err)
 				}
 			}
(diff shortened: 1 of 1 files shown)


petergrlica:
Hi @moalamri , thanks for the fix.

Can you please use rebase strategy to update your branch and squash everything into 1 commit? We'd like to keep history clean that way, thank you.

moalamri:
👍

PR #1684, opened 2024-02-13
Fix new users theme assignment

KinyaElGrande:


server/system/service/user.go
```diff
@@ -388,14 +388,14 @@ func (svc user) Create(ctx context.Context, new *types.User) (u *types.User, err
...
-		//add default user's theme
-		new.Meta.Theme = sass.LightTheme
-
 		// Process avatar initials Image
 		if err = svc.generateUserAvatarInitial(ctx, new); err != nil {
 			return
 		}
 
+		//add default user's theme
+		new.Meta.Theme = sass.LightTheme
```
tjerman, line 401: I'd imagine this would be better of as

```suggestion
		if new.Meta == nil {
				new.Meta = &types.UserMeta{}
		}
		new.Meta.Theme = sass.LightTheme
```

server/system/service/auth.go
```diff
@@ -382,6 +383,12 @@ func (svc *auth) InternalSignUp(ctx context.Context, input *types.User, password
 			createUserHandle(ctx, svc.store, nUser)
 		}
 
+		if nUser.Meta == nil {
+			nUser.Meta = &types.UserMeta{
+				Theme: sass.LightTheme, //set default user's theme
+			}
+		}
```
tjerman:
```suggestion
		if nUser.Meta == nil {
			nUser.Meta = &types.UserMeta{}
		}
		nUser.Meta.Theme = sass.LightTheme //set default user's theme
```

tjerman (commented):
I think this would be better off like so

PR #1720, opened 2024-03-05
Fix multivalue fields deduplication

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1434

The following PR introduces  records multifield de-duplication check when multi-field values `equal` configuration is set.

### DIFF
--- a/server/compose/types/record_detect_duplicates.go
+++ b/server/compose/types/record_detect_duplicates.go
@@ -72,10 +72,16 @@ func (d deDup) CheckDuplication(ctx context.Context, rules DeDupRuleSet, rec Rec
 		if rule.HasAttributes() {
 			values := rr.GetValuesByName(distinct(rule.Attributes())...)
 
-			set := rule.validateValue(ctx, d.ls, rec, values)
+			for valID, value := range values {
+				if valID == rec.ID {
+					return nil
+				}
 
-			if !set.IsValid() {
-				out.Push(set.Set...)
+				set := rule.validateValue(ctx, d.ls, rec, value)
+
+				if !set.IsValid() {
+					out.Push(set.Set...)
+				}
 			}
 		}
 		return nil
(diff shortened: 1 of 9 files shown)


katrinDY:
Tested and pointed out a couple of issues in dms

Fajfa (approved):

server/compose/types/record_detect_duplicates.go
```diff
@@ -146,6 +156,30 @@ func (rule DeDupRule) checkCaseSensitiveDuplication(ctx context.Context, ls loca
...
+							"rule":          rule.String(),
+						},
+					})
+
+					return valErr
+				}
+			}
+
+			return nil
+		}
```
tjerman: Try to flatten this out a bit; if we go any more to the right we'll end up in space 😛

server/compose/types/record_detect_duplicates.go
```diff
@@ -146,6 +156,30 @@ func (rule DeDupRule) checkCaseSensitiveDuplication(ctx context.Context, ls loca
 			valErr = &RecordValueErrorSet{}
 		)
 
+		existingVv := vv.FilterByName(c.Attribute)
+		if c.IsAllEqual() {
+			if rvv.Len() == existingVv.Len() {
+				rvvmap := recordValueFrequencyMap(c.Modifier, rvv)
+				existingVvmap := recordValueFrequencyMap(c.Modifier, existingVv)
```
tjerman: Might be worth considering memory pressure; since this would have a very tight loop, it might impact performance due to heavy allocations cause of `map`.

It might all get inlined so it might not make sense so should be checked

tjerman (commented):
Consider my two comments; else looks good 👍

tjerman (approved):

katrinDY:
One more thing to be addressed and we should be good to go

tjerman (approved):
lgtm

PR #1726, opened 2024-03-07
Add open in new tab on redirect WF prompts

### LINKED ISSUE #1709: Redirect to URL in WF Prompts to open in a new tab
When redirecting to an external url from Corteza using a workflow prompt, it doesn't allow opening in a new tab. 
This sometimes result in losing progress on the current Corteza tab and also hinders user experience.

The new feature should allow to configure if it needs to be open in a new tab or same tab. Nice if it can allow selection of all html redirect 'target' options.

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1709

- [x] Add open in new tab and in modal redirects
- [x] Add prompt error notification.
- [x] Redirect in edit mode, If no recordID is provided in `recordPage` redirect.
- [x] Add default values on redirect and button variants

### DIFF
--- a/client/web/workflow/src/components/Configurator/Function.vue
+++ b/client/web/workflow/src/components/Configurator/Function.vue
@@ -132,7 +132,7 @@
                     @input="$root.$emit('change-detected')"
                     @search="searchWorkflows"
                   />
-
+                  <!-- Clearable -->
                   <c-input-select
                     v-else-if="a.input.type === 'select'"
                     v-model="a.value"
@@ -142,6 +142,7 @@
                     :filter="varFilter"
                     :reduce="a => a.value"
                     :placeholder="$t('steps:function.configurator.option-select')"
+                    :clearable="false"
                     @input="$root.$emit('change-detected')"
                   />
 
@@ -520,7 +521,7 @@ export default {
             target: param.name,
             type: arg.type || this.paramTypes[func.ref][param.name][0],
             valueType: this.getValueType(arg, arg.type || this.paramTypes[func.ref][param.name][0], input),
-            value: arg.value || null,
+            value: arg.value || input.default || null,
             expr: arg.expr || arg.source || null,
             required: param.required || false,
             input,
--- a/lib/vue/src/components/prompts/definitions.ts
+++ b/lib/vue/src/components/prompts/definitions.ts
@@ -11,6 +11,11 @@ const variants = [
   { value: 'dark', text: 'Dark' },
 ]
 
+const openModeVariants = [
+  { value: 'newTab', text: 'Open link in a new tab' },
+  { value: 'sameTab', text: 'Open link in the same tab' },
+]
+
 export const prompts = Object.freeze([
   {
     ref: 'redirect',
@@ -19,6 +24,7 @@ export const prompts = Object.freeze([
       { name: 'owner', types: ['User', 'ID'], required: false },
       { name: 'url', types: ['String'], required: true },
       { name: 'delay', types: ['Integer'], meta: { description: 'Redirection delay in seconds' } },
+      { name: 'openMode', types: ['String'], meta: { visual: { input: { type: 'select', properties: { options: openModeVariants }, default: 'sameTab' } } } },
     ],
   },
   {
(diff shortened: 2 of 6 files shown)


lib/vue/src/components/prompts/definitions.ts
```diff
@@ -30,6 +37,7 @@ export const prompts = Object.freeze([
       { name: 'params', types: ['KV'] },
       { name: 'query', types: ['KV'] },
       { name: 'delay', types: ['Integer'], meta: { description: 'Redirection delay in seconds' } },
+      { name: 'openLink', types: ['String'], meta: { visual: { input: { type: 'select', properties: { options: linkClickVariants }, default: 'sameTab' } } } },
```
Fajfa: openLink should be renamed to mode. Since there is no links here

lib/vue/src/components/prompts/definitions.ts
```diff
@@ -19,6 +25,7 @@ export const prompts = Object.freeze([
       { name: 'owner', types: ['User', 'ID'], required: false },
       { name: 'url', types: ['String'], required: true },
       { name: 'delay', types: ['Integer'], meta: { description: 'Redirection delay in seconds' } },
+      { name: 'openLink', types: ['String'], meta: { visual: { input: { type: 'select', properties: { options: linkClickVariants }, default: 'sameTab' } } } },
```
Fajfa, line 28: we need one more variable since the input shouldn't be clearable

lib/vue/src/components/prompts/kinds/index.ts
```diff
@@ -77,11 +84,20 @@ const definitions: Record<string, PromptDefinition> = {
...
+          if (openLink === 'newTab') {
+            // @ts-ignore
+            const url = this.$router.resolve(routeParams).href
+            // @ts-ignore
+            window.open(url, '_blank')
+          }
+
           // @ts-ignore
-          this.$router.push({ name, params, query })
+          this.$router.push(routeParams)
```
Fajfa: why are we doing 2 redirects?

lib/vue/src/components/prompts/kinds/index.ts
```diff
@@ -155,18 +176,16 @@ const definitions: Record<string, PromptDefinition> = {
...
 
       // @ts-ignore
       if (this.$root.$options.name === 'compose') {
-        if (!edit && !recordID) {
-          throw new Error('invalid record page prompt configuration')
-        }
-
         let name = 'page.record'
-        if (edit) {
+        if (edit || recordID === '') {
```
Fajfa: can we do !recordID || recordID === NoID, what if it is undefined or null?

Fajfa (changes requested):

lib/vue/src/components/prompts/definitions.ts
```diff
@@ -11,6 +11,12 @@ const variants = [
   { value: 'dark', text: 'Dark' },
 ]
 
+const linkClickVariants = [
```
Fajfa: Not all have the three modes

lib/vue/src/components/prompts/kinds/index.ts
```diff
@@ -179,8 +198,22 @@ const definitions: Record<string, PromptDefinition> = {
           if (reloadPage) {
             window.location.reload()
           } else {
```
Fajfa, line 213: No need for nested

lib/vue/src/components/prompts/kinds/index.ts
```diff
@@ -61,9 +61,16 @@ const definitions: Record<string, PromptDefinition> = {
...
       if (url !== undefined) {
         console.debug('redirect to %s via prompt in %d sec', url, delay)
         setTimeout(() => {
+          if (openLink === 'newTab') {
+            // @ts-ignore
+            window.open(url, '_blank')
+          }
+
           // @ts-ignore
           window.location = url
```
Fajfa: same here, if/else

client/web/workflow/src/components/Configurator/Function.vue
```diff
@@ -520,7 +520,7 @@ export default {
             target: param.name,
             type: arg.type || this.paramTypes[func.ref][param.name][0],
             valueType: this.getValueType(arg, arg.type || this.paramTypes[func.ref][param.name][0], input),
-            value: arg.value || null,
+            value: arg.value || input.default || null,
             expr: arg.expr || arg.source || null,
             required: param.required || false,
             input,
```
Fajfa, line 526: :clearable="properties.clearable"

Fajfa, line 527: make the c-input-select clearable

PR #1727, opened 2024-03-07
Add configuration to create and use global block on a namespace

### LINKED ISSUE #1598: Add option to make and use global blocks
https://forum.cortezaproject.org/t/global-blocks-for-navigation/1602/3

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass


client/web/compose/src/components/Admin/Page/Builder/Selector.vue
```diff
@@ -33,30 +33,74 @@
...
       >
         <b-input-group class="d-flex w-100">
           <c-input-select
-            v-model="selectedExistingBlock"
+            v-model="selectedGlobalBlock"
+            :get-option-label="getBlockLabel"
+            :get-option-key="getOptionKey"
+            :options="selectableNamespaceGlobalBlocks"
+            :reduce="b => b.blockID"
+            placeholder="Select global block from other pages"
```
Fajfa: Translations here

client/web/compose/src/components/Admin/Page/Builder/Selector.vue
```diff
@@ -33,30 +33,74 @@
...
+        v-if="selectableNamespaceGlobalBlocks.length"
         cols="12"
       >
         <b-input-group class="d-flex w-100">
           <c-input-select
-            v-model="selectedExistingBlock"
+            v-model="selectedGlobalBlock"
+            :get-option-label="getBlockLabel"
+            :get-option-key="getOptionKey"
+            :options="selectableNamespaceGlobalBlocks"
```
Fajfa: Can we just call this `selectableGlobalBlocks`

client/web/compose/src/components/Admin/Page/Builder/Selector.vue
```diff
@@ -33,30 +33,74 @@
...
+            placeholder="Select global block from other pages"
+          />
+
+          <b-input-group-append>
+            <b-button
+              v-b-tooltip.noninteractive.hover="{ title: $t('selector.tooltip.clone.noRef'), container: '#body' }"
+              variant="extra-light"
+              :disabled="!selectedGlobalBlock"
+              class="d-flex align-items-center"
+              @click="$emit('select', fetchBlockData(selectedGlobalBlock).clone())"
```
Fajfa: Since these @click are the same for both input groups, can you make them in a method, so its less code?

client/web/compose/src/components/Admin/Page/Builder/Selector.vue
```diff
@@ -225,9 +275,22 @@ export default {
...
     setDefaultValues () {
       this.current = undefined
-      this.selectedExistingBlock = undefined
+      this.selectedLayoutExistingBlock = undefined
+      this.selectedGlobalBlock = undefined
       this.types = []
     },
+
+    getOptionKey ({ blockID }) {
+      return blockID
```
Fajfa: No need for a separate function for this, since its just b => b.blockID

client/web/compose/src/components/Admin/Page/Builder/Selector.vue
```diff
@@ -33,30 +33,74 @@
...
+      </b-col>
+
+      <b-col
+        v-if="existingLayoutBlocks.length"
+        cols="12"
+        class="mt-3"
+      >
+        <b-input-group class="d-flex w-100">
+          <c-input-select
+            v-model="selectedLayoutExistingBlock"
```
Fajfa: I would call this `selectedLayoutBlock`, since both global and layout  blocks are existing

client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -128,6 +128,16 @@
           >
             {{ $t('general.hidden.label') }}
           </b-form-checkbox>
+          <b-form-checkbox
+            v-if="block.kind !== 'Tabs'"
+            v-model="block.meta.namespaceID"
+            :value="namespace.namespaceID"
+            :disabled="alreadyGlobalBlock"
```
Fajfa: Not sure we need the alreadyGlobalBlock.
You just need to properly provide the :value and @update events to toggle the property

client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -266,4 +298,4 @@ export default {
 .mh-tab {
   max-height: calc(100vh - 16rem);
 }
-</style>
+</style>import { watch } from 'fs'
```
Fajfa: Guessing this is a lapsus, strange the linter didn't catch it

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -485,10 +502,28 @@ export default {
...
     },
 
-    selectableExistingBlocks () {
+    selectableExistingLayoutBlocks () {
       return this.page.blocks.filter(({ blockID }) => !this.usedBlocks.some(b => b.blockID === blockID))
     },
 
+    selectableNamespaceGlobalBlocks () {
+      const { namespaceID } = this.namespace
+      const namespace = this.namespaces.find((n) => n.namespaceID === namespaceID)
```
Fajfa: I would add a method to the store that would get the namespace blocks and change the ID's like you do here.

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -485,10 +502,28 @@ export default {
...
+
+      if (!namespace) {
+        return []
+      }
+
+      return cloneDeep(namespace.blocks).map((b, i) => {
```
Fajfa: This triggers an error on my side `.map of undefined`

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -485,10 +502,28 @@ export default {
...
+        const block = compose.PageBlockMaker(b)
+        block.blockID = `${namespaceID}-${b.blockID}`
+
+        return block
+      }).filter(b => {
+        return !this.blocks.find(({ blockID }) => blockID === b.blockID)
```
Fajfa: Should we use the usedBlocks like the layout blocks uses in line 506?

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -869,25 +905,38 @@ export default {
       return Promise.all([
         this.findPageByID({ ...this.page, force: true }),
         this.findLayoutByID({ ...this.layout }),
-      ]).then(([page, layout]) => {
-        const blocks = [
+      ]).then(async ([page, layout]) => {
+        let blocks = [
```
Fajfa: Do you need a let?

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -916,6 +965,46 @@ export default {
...
 
+    async processGlobalBlocks (globalBlocks = []) {
+      if (!globalBlocks.length) return []
+
+      let { namespaceID, name, slug, enabled, meta } = this.namespace
+
+      // If new blocks are added to global blocks update the ID's
+      const newGlobalBlocks = globalBlocks
+        // We can't detect new global block through NOID because there can be an existing local block that
+        // was now made global previously and make a global block which means it does have the NoID added to it
```
Fajfa: Im not sure what this comment says 😅
In any case, cant you detect a global block if it has the current namespaceID and NOID ?

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -916,6 +965,46 @@ export default {
...
+        .map((block) => {
+          block.meta.namespaceID = namespaceID
+
+          return block
+        })
+
+      // Update existing global blocks
+      const existingGlobalBlocks = this.namespace.blocks
+
+      // Update the state of the existing global blocks so the include the updated state changes
```
Fajfa: Same here, not sure what that means.

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -1152,6 +1250,20 @@ export default {
       this.$root.$off('tab-createRequest', this.fulfilCreateRequest)
       this.$root.$off('tabChange', this.untabBlock)
     },
+
+    fetchBlockData ({ blockID, meta = {} }) {
+      blockID = fetchID({ blockID, meta })
+
+      if (meta.namespaceID) {
+        const namespace = this.namespaces.find((n) => n.namespaceID === this.namespace.namespaceID)
```
Fajfa: extract blocks from the namespace, fallback and then you don't need to have an if

lib/js/src/compose/types/page-block/base.ts
```diff
@@ -24,28 +24,32 @@ interface PageBlockStyle {
 interface PageBlockMeta {
   hidden?: boolean;
   tempID?: string;
+  isGlobal?: boolean;
```
Fajfa: isGlobal can be removed right?

lib/js/src/compose/types/page-block/base.ts
```diff
@@ -24,28 +24,32 @@ interface PageBlockStyle {
...
+  namespaceID?: boolean;
```
Fajfa: If you can add an inline comment what this is used for just in case

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -180,6 +180,8 @@ general:
   tooltip:
     dragAndDrop: Drag and drop to change order
     translations: Field translations
+  globalBlock:
+    label: Make a global block
```
Fajfa: Shouldnt be `make a global block` since its not a button.

Fajfa (changes requested):

client/web/compose/src/views/Admin/Pages/Builder.vue
```diff
@@ -1293,4 +1281,4 @@ div.toolbox {
     right: auto;
   }
 }
-</style>
+</style>, includes
```
Fajfa: ^

lib/js/src/compose/types/page-block/base.ts
```diff
@@ -24,7 +24,7 @@ interface PageBlockStyle {
 interface PageBlockMeta {
   hidden?: boolean;
   tempID?: string;
-  isGlobal?: boolean;
+  // `namespaceID` is used to identify what namespace the block belongs too and also if the block is a global block on the namespace
   namespaceID?: boolean;
```
Fajfa: Not sure this is a boolean, it should probably be a string

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/Configurator.vue
```diff
@@ -275,27 +269,25 @@ export default {
...
-    },
-  },
-
   methods: {
     updateRefresh (e) {
       // If value is less than 5 but greater than 0 make it 5. Otherwise value stays the same.
       this.block.options.refreshRate = e.target.value < 5 && e.target.value > 0 ? 5 : e.target.value
     },
+    updateGlobalState (value) {
+      this.block.blockID = NoID
```
Fajfa: Can you maybe save the old values so you can use them for the toggle

Remember original value and blockID

If original value matches current values -> set the blockID to the old one and you set the namespaceID to the old one

lib/js/src/compose/types/namespace.ts
```diff
@@ -38,6 +40,8 @@ export class Namespace {
 
   public meta: object = {}
 
+  public blocks: Array<any> = []
```
Fajfa: Make sure you use the same type as Page

server/compose/envoy/yaml_decode.go
```diff
@@ -103,6 +103,40 @@ func (d *auxYamlDoc) unmarshalPageBlocksNode(r *types.Page, n *yaml.Node) (refs
 	return
 }
 
+func (d *auxYamlDoc) unmarshalNamespaceBlocksNode(r *types.Namespace, n *yaml.Node) (refs map[string]envoyx.Ref, idents envoyx.Identifiers, err error) {
```
tjerman, line 81: I'm not a fan of this copy paste; since these and page blocks are basically the same; can we reuse the code?

server/compose/envoy/yaml_encode.go
```diff
@@ -107,7 +107,26 @@ func (e YamlEncoder) encodePageBlocksC(ctx context.Context, p envoyx.EncodeParam
 	return out, nil
 }
 
-func (e YamlEncoder) encodePageBlockC(ctx context.Context, p envoyx.EncodeParams, tt envoyx.Traverser, n *envoyx.Node, pg *types.Page, index int, b types.PageBlock) (_ any, err error) {
+func (e YamlEncoder) encodeNamespaceBlocksC(ctx context.Context, p envoyx.EncodeParams, tt envoyx.Traverser, n *envoyx.Node, pg *types.Namespace, bb types.PageBlocks) (_ any, err error) {
+	out, _ := y7s.MakeSeq()
+
+	var aux any
+	for i, b := range pg.Blocks {
+		aux, err = e.encodePageBlockC(ctx, p, tt, n, i, b)
```
tjerman: You've done it here; so why not at the decode also?

server/compose/types/namespace.go
```diff
@@ -15,6 +15,7 @@ type (
 		Slug    string        `json:"slug"`
 		Enabled bool          `json:"enabled"`
 		Meta    NamespaceMeta `json:"meta"`
+		Blocks  PageBlocks    `json:"blocks"`
```
tjerman: I'm not really a fan of reusing this struct like so.
I'd prefer it sticks consistent as we have it with [page layouts](https://github.com/cortezaproject/corteza/blob/2023.9.x/server/compose/types/page_layout.go#L31)

server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -242,6 +243,13 @@ func fix_2023_03_00_migrateComposePageMeta(ctx context.Context, s *Store) (err e
 	)
 }
 
+func fix_2024_03_00_migrateNamespacePageBlocks(ctx context.Context, s *Store) (err error) {
+	return addColumn(ctx, s,
+		"compose_namespace",
+		&dal.Attribute{Ident: "blocks", Type: &dal.TypeJSON{DefaultValue: "[{}]"}},
```
tjerman: Oof, I don't know how much I like you giving it an empty object; feels like this will cause more issues then it solves.
Why not just `[]`?

tjerman (changes requested):
In addition to this; what happens if we change the block definition (the name/id for example), delete it, etc? How do we assure references are ok and we don't just end up with broken pages?

server/compose/types/namespace.go
```diff
@@ -29,6 +29,28 @@ type (
...
+		// Warning: value of this field is now handled via resource-translation facility
+		//          struct field is kept for the convenience for now since it allows us
+		//          easy encoding/decoding of the outgoing/incoming values
+		Title string `json:"title,omitempty"`
+
+		// Warning: value of this field is now handled via resource-translation facility
+		//          struct field is kept for the convenience for now since it allows us
+		//          easy encoding/decoding of the outgoing/incoming values
+		Description string `json:"description,omitempty"`
+	}
```
tjerman, line 52: Seems fine but you could consider just doing this:

```go
GlobalBlock struct {
		PageBlock
	}
```

Since you're basically doing a full blown copy of the page block this would be fine.
For page layouts, I preserve just the reference and coordinates and size since that's all that we care about there.

tjerman (approved):
A consideration, else looks good to me 👍 ; wp

katrinDY:
A couple of questions:

- is the margin-top here intentional?
Screenshot 2024-04-15 at 10 46 43

- after marking a block as global and copying it **without** reference, the new block **isn't** marked as global. Is this intentional? **confirmed; it's fine**
-  after marking a block as global and copying it **with** reference, the new block **is** marked as global. Is this intentional? **confirmed; it's fine**

Fajfa (approved):

Issue #1646, opened 2024-01-23
Number precision ignored when encoding number as column or alias


martkaczmarek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version of Corteza

2023.9.2

### Current Behavior

When creating a number field with precision 3 for example, and encoding it as a column and not JSON field, the column created in SQL Server database is always of default type decimal(18,0). That means 18 decimal digits are stored, but 0 precision digits (no digits after comma). This is what application uses to add a column:
```markdown
ALTER TABLE "tableName" ADD "columnName" DECIMAL
```

### Expected Behavior

The solution is to use type decimal(18,n), where n is precision specified in module field configuration in Corteza:
```markdown
ALTER TABLE "tableName" ADD "columnName" DECIMAL(18,6)
```

### Steps To Reproduce

1. Create namespace
2. Create module with numerical field, precision 3; store it as a column in MSSQL database
3. Table is created with decimal(18,0) column, instead of decimal(18,3)

### Environment and versions

```markdown
Corteza 2023.9.2 Docker install
SQL Server 2019
```

martkaczmarek:
On a second thought, there may be a quick and dirty way to do this - just use 6 as precision value every time.

martkaczmarek:
No stale

Fajfa:
Afaik this has been resolved in one of the previous releases. Closing.
If not, feel free to reopen.

PR #1731, opened 2024-03-12
Add decimal places to number columns in MSSQL

martkaczmarek:
temporary fix to store numbers with decimal places in MSSQL - use DECIMAL(18,6) instead of DECIMAL

Use column type DECIMAL(18,6) instead of DECIMAL, which has no decimal places by default

Related to #1646

### DIFF
--- a/server/store/adapters/rdbms/drivers/mssql/dialect.go
+++ b/server/store/adapters/rdbms/drivers/mssql/dialect.go
@@ -193,6 +193,8 @@ func (mssqlDialect) AttributeToColumn(attr *dal.Attribute) (col *ddl.Column, err
 
 	case *dal.TypeNumber:
 		col.Type.Name = "DECIMAL"
+		// @todo temporary fix to store numbers with decimal places
+		col.Type.Name += "(18,6)"
 		// @todo precision, scale?
 		col.Default = ddl.DefaultNumber(t.HasDefault, t.Precision, t.DefaultValue)
 


tjerman:
Are you sure? [Here](https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16#arguments) it states that the default is 18

tjerman:
Ah... never mind; looked at the wrong thing.. my bad

martkaczmarek:
Yup, when defined as DECIMAL, the MSSQL engine assumes DECIMAL(18,0), meaning 18 digits but 0 to the right of the decimal point.

martkaczmarek:
No stale

petergrlica:
@tjerman @martkaczmarek 

https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16 says

Scale (the second param):
> The number of decimal digits that are stored to the right of the decimal point. This number is subtracted from p to determine the maximum number of digits to the left of the decimal point.

Which means if default is `18, 0`, now it should be `24, 6` no?

Check the examples section on what I mean:
https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16#examples

If we have historical data that was used with `18, 0` then imo `18,6` would incorrectly truncate it.

PR #1736, opened 2024-03-18
re-fetch on prefilter record value change

### LINKED ISSUE #1672: Make fields/blocks, that fetch based on record values reactive
If record value used in prefilter or any other configuration changes, then it should refetch with the current value.
This will make the experience with prefilters much more reactive

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/ModuleFields/Editor/Record.vue
+++ b/client/web/compose/src/components/ModuleFields/Editor/Record.vue
@@ -166,7 +166,7 @@ import base from './base'
 import { debounce } from 'lodash'
 import { compose, NoID } from '@cortezaproject/corteza-js'
 import { mapActions, mapGetters } from 'vuex'
-import { evaluatePrefilter } from 'corteza-webapp-compose/src/lib/record-filter'
+import { evaluatePrefilter, isFieldInFilter } from 'corteza-webapp-compose/src/lib/record-filter'
 import Pagination from '../Common/Pagination.vue'
 
 export default {
@@ -282,6 +282,11 @@ export default {
 
   beforeDestroy () {
     this.setDefaultValues()
+    this.destroyEvents()
+  },
+
+  mounted () {
+    this.createEvents()
   },
 
   methods: {
@@ -291,6 +296,20 @@ export default {
       resolveRecords: 'record/resolveRecords',
       updateRecords: 'record/updateRecords',
     }),
+
+    createEvents () {
+      this.$root.$on('record-field-change', this.refetchOnPrefilterValueChange)
+    },
+
+    refetchOnPrefilterValueChange ({ fieldName }) {
+      const { prefilter } = this.field.options
+
+      if (isFieldInFilter(fieldName, prefilter)) {
+        const namespaceID = this.namespace.namespaceID
+        const moduleID = this.field.options.moduleID
+        this.fetchPrefiltered({ namespaceID, moduleID })
+      }
+    },
 
     getRecord (index = undefined) {
       const recordID = index !== undefined ? this.value[index] : this.value
@@ -369,7 +388,7 @@ export default {
       }
     },
 
-    fetchPrefiltered (q) {
+    fetchPrefiltered (q = this.filter) {
       this.processing = true
 
       // Support prefilters
@@ -521,6 +540,10 @@ export default {
       }
       const { label } = v || {}
       return label
+    },
+
+    destroyEvents () {
+      this.$root.$off('record-field-change', this.refetchOnPrefilterValueChange)
     },
 
     setDefaultValues () {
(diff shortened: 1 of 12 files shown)


client/web/compose/src/components/PageBlocks/RecordEditor.vue
```diff
@@ -214,8 +214,11 @@ export default {
       return !expressions.value
     },
 
-    onFieldChange: debounce(function () {
+    onFieldChange: debounce(function (field) {
       this.evaluateExpressions()
+      this.$root.$emit('record-field-change', {
```
Fajfa, line 219: you can just emit the field.name itself like `this.$root.$emit('record-field-change', field.name)`

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -1105,6 +1105,17 @@ export default {
...
 
       this.$root.$on('module-records-updated', this.refreshOnRelatedRecordsUpdate)
+      this.$root.$on('record-field-change', this.refetchOnPrefilterValueChange)
+    },
+
+    refetchOnPrefilterValueChange ({ fieldName }) {
+      if (!fieldName) return
+
+      const { prefilter } = this.options
+      if (!(prefilter || '').includes('${record') && !(prefilter || '').includes(fieldName)) return
```
Fajfa: This could lead to false positives since there are 2 cases here
One where you're using ownerID or record.ownedBy, and the other where you are referencing ${record.values

So for example if my filter is `fieldName = ${record.values.foo}` it would still retrigger the fetch, when it shouldn't have. Similar for if you do `ownedBy = ${record.ownedBy}`. So i think you'll have to check for those three cases (`${fieldName}`, `${record.fieldName}`, `${record.values.fieldName}`)

client/web/compose/src/components/ModuleFields/Editor/Record.vue
```diff
@@ -292,6 +297,20 @@ export default {
...
 
+    createEvents () {
+      this.$root.$on('record-field-change', this.refetchOnPrefilterValueChange)
+    },
+
+    refetchOnPrefilterValueChange ({ fieldName }) {
+      if (!fieldName) return
+
+      const { prefilter } = this.field.options
+      if (!(prefilter || '').includes('${record') && !(prefilter || '').includes(fieldName)) return
```
Fajfa: Since this keeps repeating, would it make sense to turn it into a function inside the record-filter lib?
So you can just do if `isFieldInFilter(fieldName, filter)`. So you don't have to change the logic everywhere but instead just one place if there is an edge case

client/web/compose/src/components/ModuleFields/Editor/Record.vue
```diff
@@ -292,6 +297,20 @@ export default {
...
+
+      this.selected = { value: undefined }
```
Fajfa: I dont think you should reset the value on refetch

client/web/compose/src/components/ModuleFields/Editor/Record.vue
```diff
@@ -292,6 +297,20 @@ export default {
...
+      this.loadLatest()
```
Fajfa: shouldnt you use fetchPrefiltered? Since loadLatest doesnt have the updated filter as far as i can see.

Fajfa (changes requested):

client/web/compose/src/lib/record-filter.js
```diff
@@ -208,3 +208,18 @@ function trimChar (text = '', char = '') {
...
+export function isFieldInFilter (fieldName, filter) {
+  if (!fieldName) return
+
+  const filterCases = [
+    '${' + fieldName + '}',
+    '${record.' + fieldName + '}',
+    '${record.values.' + fieldName + '}',
+  ]
+
+  if ((filter || '').includes(filterCases[0]) ||
```
Fajfa: You can just do `return filterCases.some(f => filter.includes(f))`

client/web/compose/src/lib/record-filter.js
```diff
@@ -208,3 +208,18 @@ function trimChar (text = '', char = '') {
...
+// Helper function that checks if field name is included in a filter
+export function isFieldInFilter (fieldName, filter) {
+  if (!fieldName) return
+
+  const filterCases = [
+    '${' + fieldName + '}',
+    '${record.' + fieldName + '}',
+    '${record.values.' + fieldName + '}',
+  ]
+
```
Fajfa, line 221: Instead of falling back 3 times just do it where you define the filter param

Fajfa (changes requested):

PR #1783, opened 2024-05-02
Replicate record list filter ui to record list configurator prefilter

kelanik8:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass


locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -289,6 +289,9 @@ recordList:
   addRecord: Add
   cancelSelection: Cancel
   customFilter: Custom filter
+  prefilter:
+    toggleInputType: Toggle input type
+    toggleFilterBuilderType: Toggle filter builder type
```
Fajfa: No need for this translation, (line 294)
The first one is enough

client/web/compose/src/components/PageBlocks/RecordList/Components/Prefilter.vue
```diff
@@ -0,0 +1,593 @@
+<template>
+  <c-form-table-wrapper hide-add-button>
```
Fajfa, line 2: The prefilter component itself should only emit the value if it changes. 
That way it can be used in any place where we do record list filters.
It would also mean the RecordListFilter component could use it (which removes duplicate code)

It should also not be wrapped inside a c-form-table here, but instead just in the configurator(where its used), since we just want to style that form

In essence this component is a black box that parses/emits the record filter value which we can use.
So it shouldn't concern itself with such things as the block options. That should be achieved through props

client/web/compose/src/components/PageBlocks/RecordList/Base.vue
```diff
@@ -743,12 +743,12 @@
 import axios from 'axios'
 import { isEqual } from 'lodash'
 import { mapGetters, mapActions } from 'vuex'
-import base from './base'
+import base from '../base'
```
Fajfa: You can reference it with and absolute path(corteza-webapp-compose), so that we don't need to do this in the future. Same for automation buttons below

Fajfa (changes requested):
Some duplicate code can be removed with a bit of a rework. Otherwise looks and works great!

client/web/compose/src/components/Common/RecordListFilter.vue
```diff
@@ -246,16 +88,18 @@
   </div>
 </template>
 <script>
-import FieldEditor from '../ModuleFields/Editor'
+// import FieldEditor from '../ModuleFields/Editor'
```
Fajfa: Guessing you can remove this

client/web/compose/src/components/Common/RecordListFilter.vue
```diff
@@ -246,16 +88,18 @@
...
+import FilterToolbox from 'corteza-webapp-compose/src/components/Common/FilterToolbox.vue'
 
 export default {
   i18nOptions: {
     namespaces: 'block',
   },
 
   components: {
-    FieldEditor,
+    // FieldEditor,
```
Fajfa: Same here

client/web/compose/src/components/PageBlocks/RecordList/Base.vue
```diff
@@ -739,16 +739,17 @@
     </template>
   </wrap>
 </template>
+
 <script>
 import axios from 'axios'
 import { isEqual } from 'lodash'
 import { mapGetters, mapActions } from 'vuex'
-import base from './base'
+import base from 'corteza-webapp-compose/src/components/PageBlocks/base.vue'
```
Fajfa: I dont think you need the .vue suffixes (i don't care really, just saying)

Fajfa (changes requested):

PR #1802, opened 2024-05-23
Fix MFA authentication page error: handle user read permissions

### LINKED ISSUE #1782: When I log in with a user who has enabled 2 factor authentication I get a template: mfa.html.tpl:1:34 error

### Version of Corteza

2023.9.6

### Current Behavior

When I log in with a user who has enabled 2 factor authentication I get the following error:

`template: mfa.html.tpl:1:34: executing “mfa.html.tpl” at <set . “hideNav” true>: error calling set: assignment to entry in nil map`

image

### Expected Behavior

I expect that I can use MFA and that the template is rendered.

### Steps To Reproduce

1. Use the following settings:

image

2. Log in with a user with MFA enabled.

### Environment and versions

_No response_

### Anything else?

Issue mentioned on Taiga and on the forum: https://forum.cortezaproject.org/t/mfa-errors-on-2023-9-6-release/2063

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1782

### DIFF
--- a/server/auth/handlers/handler.go
+++ b/server/auth/handlers/handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/gob"
 	"fmt"
+	"github.com/cortezaproject/corteza/server/pkg/errors"
 	"html/template"
 	"io"
 	"mime/multipart"
@@ -335,10 +336,21 @@ func (h *AuthHandlers) enrichTmplData(req *request.AuthReq) interface{} {
 	d := req.Data
 	d["theme"] = "light"
 	if req.AuthUser != nil {
+		maskEmail := service.CurrentSettings.Privacy.Mask.Email
+		maskName := service.CurrentSettings.Privacy.Mask.Name
+
+		service.CurrentSettings.Privacy.Mask.Email = false
+		service.CurrentSettings.Privacy.Mask.Name = false
+
 		// fetch current user with updated fields
 		user, err := h.UserService.FindByAny(req.Context(), req.AuthUser.User.ID)
-		if err != nil {
-			return nil
+
+		service.CurrentSettings.Privacy.Mask.Email = maskEmail
+		service.CurrentSettings.Privacy.Mask.Name = maskName
+
+		// check if err is not nil and cater for MFA by checking if the error is not allowed to read
+		if err != nil && !errors.Is(err, service.UserErrNotAllowedToRead()) {
+			return err
 		}
 
 		d["user"] = user


server/auth/handlers/handler.go
```diff
@@ -337,11 +338,13 @@ func (h *AuthHandlers) enrichTmplData(req *request.AuthReq) interface{} {
...
-			return nil
+
+		// check if err is not nil and cater for MFA by checking if the error is not allowed to read
+		if err != nil && !errors.Is(err, service.UserErrNotAllowedToRead()) {
+			return err
 		}
 
-		d["user"] = user
+		d["user"] = req.AuthUser.User
 		d["theme"] = user.Meta.Theme
```
petergrlica: is there a reason you did not change to `req.AuthUser` here as well?

server/auth/handlers/handler.go
```diff
@@ -337,11 +338,13 @@ func (h *AuthHandlers) enrichTmplData(req *request.AuthReq) interface{} {
 	if req.AuthUser != nil {
 		// fetch current user with updated fields
 		user, err := h.UserService.FindByAny(req.Context(), req.AuthUser.User.ID)
-		if err != nil {
-			return nil
+
+		// check if err is not nil and cater for MFA by checking if the error is not allowed to read
+		if err != nil && !errors.Is(err, service.UserErrNotAllowedToRead()) {
```
petergrlica, line 352: what happens when there is a `UserErrNotAllowedToRead` error?

server/auth/handlers/handler.go
KinyaElGrande: Yes, the reason is `req.AuthUser` doesn't have the most updated data.

server/auth/handlers/handler.go
KinyaElGrande: In this case, since we have the User details from `req.AuthUser` and we're searching for the most updated `users` data, the  UserErrNotAllowedToRead is ignored.

KinyaElGrande: Unless maybe we don't want the current user's theme to be applied on MFA login screen, then we can handle it the same way we handle other errors.

server/auth/handlers/handler.go
```diff
@@ -335,10 +336,18 @@ func (h *AuthHandlers) enrichTmplData(req *request.AuthReq) interface{} {
...
 	if req.AuthUser != nil {
+		service.CurrentSettings.Privacy.Mask.Email = false
+		service.CurrentSettings.Privacy.Mask.Name = false
+
 		// fetch current user with updated fields
 		user, err := h.UserService.FindByAny(req.Context(), req.AuthUser.User.ID)
-		if err != nil {
-			return nil
+
+		service.CurrentSettings.Privacy.Mask.Email = true
```
petergrlica: how do you know this was `true` before in all cases? maybe just use a temp variable

PR #1812, opened 2024-06-05
Null values on datetime fields

petergrlica:
Handling null values of `expr.Datetime` on db layer and in expr
Added more tests


tjerman (approved):
lgtm

PR #1844, opened 2024-07-01
Fix Postgres timestamp fields sort inconsistencies

### LINKED ISSUE #1843: Sorting bug
Currently, Corteza supports time precision only up to the second mark but saves the timestamp field up to the milliseconds mark (Postgres). This limitation causes a sorting issue when sorting fields with timestamps such as `created_at`, especially when multiple records have the same timestamp up to the second mark but differ by milliseconds.

This inconsistency affects in pagination and record list count on sort(**ASC/DESC)**.

KinyaElGrande:
Fixes: https://github.com/cortezaproject/corteza/issues/1843


tjerman (approved):
lgtm but what about the other dialects? You've only updated psql from what I can see

PR #1877, opened 2024-08-08
Add support for custom CDN javascripts on webapps

### LINKED ISSUE #1875: Add Code Snippets similar to customCSS

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/1875

Add support for custom javascript CDNs on webapps.

#### Admin UI screenshot:

Screenshot 2024-08-15 at 15 39 57


server/pkg/webapp/serve.go
```diff
@@ -152,6 +157,57 @@ func serveConfig(r chi.Router, config webappConfig) {
...
+        doc, err := html.Parse(strings.NewReader(customJS))
+        if err != nil {
+            log.Fatal(err)
+        }
+
+        var scriptsAttrs []scriptAttrs
+        traverseScriptsNode(doc, &scriptsAttrs)
+
+        //create a javascript array of objects
+        jsonScripts, err := json.Marshal(scriptsAttrs)
```
tjerman: `jsonScripts`?

server/pkg/webapp/serve.go
```diff
@@ -183,3 +239,29 @@ func replaceBaseHrefPlaceholder(buf []byte, app, baseHref string) []byte {
 
 	return fixed
 }
+
+func traverseScriptsNode(n *html.Node, scripts *[]scriptAttrs) {
```
tjerman: Nooo; utilize returns over this pointer stuff; recursive functions can still return correctly

tjerman (changes requested):
What exactly is the point of this feature?

We're letting users provide arbitrary javascript yet we want them to work with HTML (add appropriate `script` tags and such).
Yet, we're yanking out most of the custom HTML in the preprocessing on server-side.
In server code, you keep stating these are CDN scripts, yet we allow arbitrary code to be added.

I would love to see the following:

* **External script inclusion section**: a section in the UI 100% dedicated for external resource inclusion. A simple list where you add entries and specify all the required parameters (no effort from the user needed).
* **Custom JavaScript section**: a section in the UI 100% dedicated for arbitrary JS which gets injected. I'd love to see an option to add multiple _JS files_ so users can have multiple smaller files instead of a gigantic one.

___

A note on custom JS.

Since Corredor already has client scripts, would it make more sense to utilize/extend that? How much functionality do we want this different JS editor to have? Should this JS section primarily focus on UI & design?

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+            OK
+          </b-button>
+        </template>
+      </b-modal>
+
+      <template #footer>
+        <c-button-submit
+          :disabled="!canManage"
+          :processing="processing"
+          :success="success"
```
Fajfa: success and processing should be prefixed with the cdn since that's where you store them

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+        :items="providerItems"
+        :fields="cdnProviderFields"
+        head-variant="light"
+        hover
+        class="mb-0"
+      >
+        <template #cell(provider)="{ item }">
+          {{ item.provider || item.tag }}
+        </template>
+
```
Fajfa, line 36: Add empty state ("No snippets yet")

locale/en/corteza-webapp-admin/navigation.yaml
```diff
@@ -9,6 +9,7 @@ system:
     applications: Applications
     authclients: Auth Clients
     connections: Connections
+    cdns: CDNs
```
Fajfa: All user facing things should be called "Code Snippets", we can use cdn internally.

client/web/admin/src/components/Settings/System/CDNs/CSystemCdnProviderEditor.vue
```diff
@@ -0,0 +1,64 @@
+<template>
```
Fajfa: I would prefer if you just keep this in the original file. (CDNS/Index)

Also the whole components approach is not needed here since they'll all be the same.

client/web/reporter/vue.config-builder.js
```diff
@@ -125,7 +125,7 @@ module.exports = ({ appFlavour, appLabel, version = process.env.BUILD_VERSION, t
           target: fetchBaseUrl(),
         },
 
-        '^/custom.js': {
+        '^/cdns-provider.js': {
```
Fajfa: Sorry to make you change this but i would say code-snippets instead of cdns-provider.
Same for everywhere else.

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+      <b-modal
+        id="modal-cdn"
+        v-model="modal.open"
+        :title="modal.title"
+        scrollable
+        size="lg"
+        title-class="text-capitalize"
+        @ok="modal.updater(modal.data)"
+      >
+        <component
```
Fajfa: This can be replaced with static html. Just copy the component here

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+          :is="modal.component"
+          v-model="modal.data"
+        />
+        <template #modal-footer="{ ok, cancel}">
+          <b-button
+            size="sm"
+            variant="danger"
+            @click="deleteCDN(modal.index)"
+          >
+            Delete
```
Fajfa: Translation strings. Also the delete button should only be visible for existing code snippets.
It should also be aligned left.

Make sure its an input confirm aswell

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+        title-class="text-capitalize"
+        @ok="modal.updater(modal.data)"
+      >
+        <component
+          :is="modal.component"
+          v-model="modal.data"
+        />
+        <template #modal-footer="{ ok, cancel}">
+          <b-button
+            size="sm"
```
Fajfa: Remove size="sm" from the buttons, please.

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+            @click="cancel()"
+          >
+            Cancel
+          </b-button>
+          <b-button
+            size="sm"
+            variant="primary"
+            @click="ok()"
+          >
+            OK
```
Fajfa: Consult other modal footers. In general we have (Delete, Cancel, Save)

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+
+      // deref
+      this.modal.data = data
+    },
+
+    newCDN () {
+      this.openEditor({
+        component: 'cdn-provider-editor',
+        title: this.$t('cdns-provider.add'),
+        data: {},
```
Fajfa: New CDNS should have `<script` as default value, since placeholders are a bit tricky here.

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+        updater: (changed) => {
+          this.cdns.push(changed)
+        },
+      })
+    },
+    fetchSettings () {
+      this.incLoader()
+      this.$Settings.fetch()
+      return this.$SystemAPI.settingsList({ prefix: 'cdns' })
+        .then(settings => {
```
Fajfa, line 247: settings[0] could throw an error, make sure that's handled

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+        .finally(() => {
+          this.decLoader()
+        })
+    },
+
+    settingsUpdate (action) {
+      this.cdn.processing = true
+      this.$SystemAPI.settingsUpdate({ values: [{ name: 'cdns', value: this.cdns }] })
+        .then(() => {
+          if (action === 'delete') {
```
Fajfa, line 265: This can be a bit cleaner. You repeat the same line in both ifs.

client/web/admin/src/views/System/Cdns/Index.vue
```diff
@@ -0,0 +1,257 @@
...
+          {{ $t('cdns-provider.add') }}
+        </b-button>
+      </div>
+
+      <b-table
+        :items="providerItems"
+        :fields="cdnProviderFields"
+        head-variant="light"
+        hover
+        class="mb-0"
```
Fajfa, line 30: Some min height is necessary.

Fajfa (changes requested):
So in general, rename CDNs to Code Snippets (at least from the user perspective)
Clean up some unneeded code and stylistic details.

PR #1894, opened 2024-08-28
Add record from record selector field

### LINKED ISSUE #1795: Quick way to add records from record selector fields
https://forum.cortezaproject.org/t/add-shortcut-to-add-record-on-record-selector-field-type/2068/4

katrinDY:
TODO: Summary

 - [ ] Code builds
 - [ ] All existing tests pass

### DIFF
--- a/client/web/compose/src/components/ModuleFields/Editor/Record.vue
+++ b/client/web/compose/src/components/ModuleFields/Editor/Record.vue
@@ -40,86 +40,119 @@
       :show-list="field.options.selectType !== 'multiple'"
     >
       <template #single>
-        <c-input-select
-          v-if="field.options.selectType === 'multiple'"
-          v-model="multipleSelected"
-          :options="options"
-          :get-option-key="getOptionKey"
-          :get-option-label="getOptionLabel"
-          :disabled="!module"
-          :loading="processing"
-          :clearable="false"
-          :filterable="false"
-          :searchable="searchable"
-          :selectable="isSelectable"
-          :placeholder="placeholder"
-          multiple
-          @search="search"
-        >
-          <pagination
-            v-if="showPagination"
-            slot="list-footer"
-            :has-prev-page="hasPrevPage"
-            :has-next-page="hasNextPage"
-            @prev="goToPage(false)"
-            @next="goToPage(true)"
-          />
-        </c-input-select>
+        <b-input-group class="d-flex w-100">
+          <c-input-select
+            v-if="field.options.selectType === 'multiple'"
+            v-model="multipleSelected"
+            :options="options"
+            :get-option-key="getOptionKey"
+            :get-option-label="getOptionLabel"
+            :disabled="!module"
+            :loading="processing"
+            :clearable="false"
+            :filterable="false"
+            :searchable="searchable"
+            :selectable="isSelectable"
+            :placeholder="placeholder"
+            multiple
+            @search="search"
+          >
+            <pagination
+              v-if="showPagination"
+              slot="list-footer"
+              :has-prev-page="hasPrevPage"
+              :has-next-page="hasNextPage"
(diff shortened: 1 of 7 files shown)


client/web/compose/src/components/ModuleFields/Editor/Record.vue
```diff
@@ -499,6 +549,43 @@ export default {
       })
     },
 
+    addRecordThroughRecordSelectField () {
+      const { page } = this.getRecordSelectorPage()
```
Fajfa, line 560: You need to show the button only if record can be added for that module (same as record list +Add button check)

That means if recordPage exists and user can creat a record for that module

Fajfa, line 560: page can be undefined here which then breaks this

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -296,6 +297,7 @@ record:
     condition: Condition
     tooltip:
       performance: Using conditional fields will impact performance
+  recordSelectorAddRecordDisplayOption: Add record via selector field
```
Fajfa: On selector record create
While you're here can you also rename the On record selector click to On record value click

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -262,6 +262,7 @@ metric:
     valueLabel: Value style
   label: Metric
 record:
+  add: Users will be able to add new records
```
Fajfa: Can create records via selector

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -277,6 +311,11 @@ export default {
         this.options.inlineRecordEditEnabled = v
       },
     },
+
+    isRecordConfigured () {
+      return !!this.module.fields
+        .find(f => f.kind === 'Record' && (f.options.labelField || f.options.recordLabelField))
```
Fajfa: i think just kind is ok, since it might confuse people if they don't see it. 
Also use some instead of find so it returns a boolean

client/web/compose/src/components/PageBlocks/RecordConfigurator.vue
```diff
@@ -90,6 +90,40 @@
             />
           </b-form-group>
         </b-col>
+
```
Fajfa, line 57: Can you also add the "recordSelectorDisplayOptions and referenceField" under this template please

client/web/compose/src/components/PageBlocks/RecordEditor.vue
```diff
@@ -222,6 +235,16 @@ export default {
         fieldName: field.name,
       })
     }, 500),
+
+    refreshOnRelatedRecordsUpdate ({ moduleID, notPageID }) {
```
Fajfa: I think we misunderstood each other, I said to add this to the record editor field, and refresh its options when this event is called.
We don't refresh record editor block so that users don't lose their changes.

lib/js/src/compose/types/page-block/record.ts
```diff
@@ -11,8 +11,10 @@ interface FieldCondition {
 interface Options {
   fields: unknown[];
   fieldConditions: FieldCondition[];
+  showAddRecordButton: boolean;
```
Fajfa: since others are prefixed with record selectors lets do the same for this one
`recordSelectorShowAddRecordButton`

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordEditor.vue
katrinDY: Sorry, my bad

PR #2016, opened 2024-10-31
Add records limit count per namespace

KinyaElGrande:
This PR adds a Limit to the number of records a user could have in a Namespace.
The configuration can be adjusted by `LIMIT_RECORD_COUNT_PER_NAMESPACE` environment key.


tjerman (approved):
I think it looks good!

PR #2110, opened 2025-04-22
Fix discovery mappings, indexing and allowed roles

### LINKED ISSUE #2106: Discovery flow check

KinyaElGrande:
This PR resolves issues in the corteza-discovery flow by :
1.  Fixing the reindexing process to ensure data is actually reindexed on opensearch/elastic search
2. Adds proper mapping for the security field in indexed compose documents
3.  Adds the super-admin role in the allowed roles list for all the documents.


tjerman (approved):
looks good

PR #2126, opened 2025-06-06
2024.9.x federation issue

### LINKED ISSUE #2085: Federation flow check
- Are docs enough to set it up?
- Does it work?
- Fix https://forum.cortezaproject.org/t/federation-mapping-does-not-work/1367/4

KinyaElGrande:
Ref: https://github.com/cortezaproject/corteza/issues/2085

This PR Fixes Federation  issues and ensures that the federation service is back in operation

### DIFF
--- a/server/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
+++ b/server/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
@@ -162,14 +162,23 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
 		}
 	}
 
-	// Make sure results are always sorted at least by primary keys
-	if f.Sort.Get("id") == nil {
+	{{- $sortColumn := "id" }}
+
+	{{ if (eq .expIdentPlural "FederationModuleMappings") }}
+        {{ $sortColumn = "node_id" }}
+    {{ end }}
+
+    {{ if (eq .expIdentPlural "FederationNodeSyncs") }}
+        {{ $sortColumn = "rel_node" }}
+    {{ end }}
+
+    // Make sure results are always sorted at least by primary keys
+	if f.Sort.Get("{{ $sortColumn }}") == nil {
 		f.Sort = append(f.Sort, &filter.SortExpr{
-			Column:     "id",
+			Column:     "{{ $sortColumn }}",
 			Descending: f.Sort.LastDescending(),
 		})
 	}
-
 
 	// Cloned sorting instructions for the actual sorting
 	// Original are passed to the etchFullPageOf{{ .expIdentPlural }} fn used for cursor creation;
--- a/server/federation/node_sync.cue
+++ b/server/federation/node_sync.cue
@@ -12,15 +12,17 @@ nodeSync: {
 	model: {
 		ident: "federation_nodes_sync"
 		attributes: {
-			node_id: {
-				sortable: true,
+			rel_node: {
+			  sortable: true,
 				ident: "nodeID",
+				storeIdent: "rel_node",
 				goType: "uint64",
 				dal: { type: "ID" }
 			}
-			module_id: {
+			rel_module: {
 				sortable: true,
 				ident: "moduleID",
+				storeIdent: "rel_compose_module",
 				goType: "uint64"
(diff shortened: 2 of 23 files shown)


server/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
```diff
@@ -162,13 +162,23 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
...
+		}
+	{{ else }}
+	    // Make sure results are always sorted at least by primary keys
+		if f.Sort.Get("id") == nil {
+			f.Sort = append(f.Sort, &filter.SortExpr{
+				Column:     "id",
+				Descending: f.Sort.LastDescending(),
+			})
+		}
+	{{ end }}
```
tjerman: I'm not the biggest fan of this. I see some resources don't have an `ID` (unsure why)...
We either need to ensure all resources have a consistent ID, or we make this generic.
Perhaps you could define (generate) a constant with the primary key field for each resource

For now, due to time pressure, this is good enough, but please add some comments or open an issue to indicate that this should be looked into.

server/federation/node_sync.cue
```diff
@@ -38,14 +38,14 @@ nodeSync: {
 		}
 
 		indexes: {
-			"primary": { attribute: "node_id" }
```
tjerman, line 41: why the rename?

server/federation/node_sync.cue
```diff
@@ -38,14 +38,14 @@ nodeSync: {
...
+			"node_id_idx": { attribute: "node_id" }
 		}
 	}
 
 	filter: {
 		struct: {
-			node_id:     { goType: "uint64", storeIdent: "rel_node",   ident: "nodeID" }
```
tjerman, line 47: Columnst that point to a different table; we used a `rel_{NAME}` as a standard template -- please preserve it unless there is a good reason

server/federation/rest/sync_data.go
```diff
@@ -99,6 +99,8 @@ func (ctrl SyncData) ReadExposedAll(ctx context.Context, r *request.SyncDataRead
 			Paging:      filter.Paging{Limit: 1},
 		}
 
+		// TODO:: I think this is where a fix could potentially be implemented
```
tjerman: What fix?

tjerman (changes requested):

server/federation/node_sync.cue
KinyaElGrande: Because `node_id` shouldn't be a Primary key  within the `federation_nodes_sync` table, since there can be multiple records with the same node_id.
I renamed it to `...idx` to clearly show that it's a regular index.

server/codegen/assets/templates/gocode/store/rdbms/rdbms.go.tpl
```diff
@@ -162,15 +162,24 @@ func (s *Store) Search{{ .expIdentPlural }}(ctx context.Context, {{ template "ex
...
+    {{ if (eq .expIdentPlural "FederationNodeSyncs") }}
+        {{ $sortColumn = "rel_node" }}
+    {{ end }}
+
+    // Make sure results are always sorted at least by primary keys
+	if f.Sort.Get("{{ $sortColumn }}") == nil {
 		f.Sort = append(f.Sort, &filter.SortExpr{
-			Column:     "id",
+			Column:     "{{ $sortColumn }}",
 			Descending: f.Sort.LastDescending(),
```
tjerman, line 179: not exactly what I had in mind but I guess it does the trick just fine; please just fix indentation <3

server/federation/node_sync.cue
```diff
@@ -12,15 +12,17 @@ nodeSync: {
...
-			node_id: {
-				sortable: true,
+			rel_node: {
+			  sortable: true,
 				ident: "nodeID",
+				storeIdent: "rel_node",
 				goType: "uint64",
 				dal: { type: "ID" }
 			}
 			module_id: {
```
tjerman: This one should be `rel_module` also?

tjerman (approved):
lgtm; just do check the last few comments and make sure your SQL changes are compatible with all databases we support

PR #2134, opened 2025-06-13
Add compose record cloner function to workflows

### LINKED ISSUE #1778: When I copy record values in a workflow to a new record, I cannot update the values without it updating the value in both records.

### Version of Corteza

2023.9.5

### Current Behavior

I have a workflow that creates a new record and a step where I copy the values. Now, if I update any value in the new record, it updates also the value in the main record of which the values were copied.

image

image

image

### Expected Behavior

If I clone values in a new record, I should be able to update them without it updating the data of the values of the record I cloned the data from.

The last screenshot above should show:

```
record.values.AccountName = account name (from the record)
recordB.values.AccountName = modified
recordC.values.AccountName = modified again
```

### Steps To Reproduce

1. You can test it on QC in the CRM instance, 
2. Go to the account module. There is a "CRM - Clone" button. The workflow it calls has ID 383351435045109761.
3. Hit the button.

### Environment and versions

_No response_

### Anything else?

Reported first on the forum: https://forum.cortezaproject.org/t/copying-a-record-in-wf/2015/3

Fajfa:


server/compose/types/record.go
```diff
@@ -185,6 +185,13 @@ func (r *Record) GetModule() *Module {
 
 func (r Record) Clone() *Record {
 	c := &r
+	c.ID = 0
+	c.CreatedAt = time.Time{}
+	c.UpdatedAt = nil
+	c.DeletedAt = nil
+	c.CreatedBy = 0
+	c.UpdatedBy = 0
+	c.DeletedBy = 0
```
tjerman, line 194: This is too brave; I'd remove this

server/compose/automation/records_handler.go
```diff
@@ -436,3 +437,29 @@ func wrapRecordValueErrorSet(rr *types.Record, res *types.RecordValueErrorSet, e
...
+	}
+
+	// Clone the record
+	cloned := rec.Clone()
+
+	// Clear ID and timestamps to make it a new record
+	cloned.ID = 0
+	cloned.CreatedAt = time.Time{} // Non-pointer time.Time
+	cloned.UpdatedAt = nil         // Pointer to time.Time
+	cloned.DeletedAt = nil         // Pointer to time.Time
```
tjerman: Remove these GPT comments 🙃

tjerman (approved):
Looks fine to me; just address the 2 comments

PR #2140, opened 2025-07-01
Add support for configuring inline editable fields

### LINKED ISSUE #2131: Feature Request: Specify which fields in a record list can be inline-edited
### Discussed in https://github.com/orgs/cortezaproject/discussions/2092

<div type='discussions-op-text'>

<sup>Originally posted by **cmrusna97** March  4, 2025</sup>
To further tailor the user experience and constrain the user to actions desired by the administrator/developer, I suggest the following feature be added:

In the Record List configuration screen, allow the administrator to specify which fields displayed in the record list may be inline-edited when the inline editing option is selected (e.g. - if a field is selected, the “edit” pencil icon appears when mousing over the value; if not selected, the pencil icon does not appear). This should function independent of permissions - it’s a user-interface issue.

My use case is that I have a record list that displays four fields, but I want the user to only be able to edit one of the fields using inline editing.</div>

KinyaElGrande:
This PR adds the ability to configure inline editable fields

Ref: https://github.com/cortezaproject/corteza/issues/2131


client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -2135,7 +2135,10 @@ export default {
...
 
     showInlineActions (field) {
-      return this.showInlineEdit(field) || this.showInlineFilter(field)
+      const canInlineEdit = this.options.inlineEditFields.some(f => f.fieldID === field.moduleField.fieldID)
+      const canShowActions = this.showInlineEdit(field) || this.showInlineFilter(field)
+
+      return canInlineEdit && canShowActions
     },
 
     showInlineEdit (field) {
```
Fajfa, line 2283: Only change should be in showInlineEdit function

Fajfa, line 2283: Only change should be in showInlineEdit function

client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
```diff
@@ -689,10 +689,19 @@
               cols="12"
               lg="6"
             >
-              <b-form-group
-                :label="$t('recordList.inlineEdit.enabled')"
-                label-class="text-primary"
-              >
+              <b-form-group>
+                <label class="text-primary mr-2"> {{ $t('recordList.inlineEdit.enabled') }}</label>
```
Fajfa: Small thing.
Consider using the label slot instead (to keep in consistent with other places where label is used like this).
Do a quick #label search for examples.

client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
```diff
@@ -689,10 +689,19 @@
...
+                <b-button
```
Fajfa: Make sure to use the column-picker

locale/en/corteza-webapp-compose/block.yaml
```diff
@@ -335,6 +335,10 @@ recordList:
   inlineEdit:
     enabled: Inline value editing enabled
     allowAddField: Allow adding fields to inline edit
+    configure:
```
Fajfa: Make sure to include `fields` somewhere to make it obvious its for fields.

Fajfa (changes requested):

client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
```diff
@@ -689,19 +689,26 @@
...
+                <template #label>
+                  {{ $t('recordList.inlineEdit.enabled') }}
+
+                  <column-picker
+                    size="sm"
+                    button-class="text-secondary border-0"
+                    variant="outline-extra-light"
+                    :button-tooltip="inlineEditButtonTooltip"
+                    :module="recordListModule"
+                    :fields="options.inlineEditFields || []"
```
Fajfa: No need for the fallback here

client/web/compose/src/components/PageBlocks/RecordListBase.vue
```diff
@@ -2135,14 +2135,13 @@ export default {
...
-      const canInlineEdit = this.options.inlineEditFields.some(f => f.fieldID === field.moduleField.fieldID)
-      const canShowActions = this.showInlineEdit(field) || this.showInlineFilter(field)
-
-      return canInlineEdit && canShowActions
+      return this.showInlineEdit(field) || this.showInlineFilter(field)
     },
 
     showInlineEdit (field) {
-      return this.options.inlineRecordEditEnabled && field.canEdit && !this.showingDeletedRecords
+      const isFieldInlineEditable = this.options.inlineEditFields.some(fieldID => fieldID === field.moduleField.fieldID || fieldID === field.moduleField.name)
```
Fajfa: Please make this a function and call it at the end of the 2144 line (&& isFieldInlineEditable)
Also make sure if no fields are selected, that means all of them have it enabled.

client/web/compose/src/components/PageBlocks/RecordListConfigurator.vue
```diff
@@ -998,6 +986,13 @@ export default {
     isInlineEditorAllowed () {
       return this.recordListModule && (this.onRecordPage || this.options.editable)
     },
+
+    inlineEditButtonTooltip () {
```
Fajfa: Its fine if you define the tooltip  in the template itself (since it like this for other places)

Fajfa (changes requested):

PR #2150, opened 2025-07-25
Fix MySQL federation_nodes_sync  primary key constraint drop error

KinyaElGrande:
This PR resolves MySQL syntax error when dropping `federation_nodes_sync` primary key.

### DIFF
--- a/server/store/adapters/rdbms/upgrade_fixes.go
+++ b/server/store/adapters/rdbms/upgrade_fixes.go
@@ -808,29 +808,87 @@ func fix_2024_09_03_dropFederationNodeSyncPrimaryKey(ctx context.Context, s *Sto
 		return err
 	}
 
-	if !strings.HasPrefix(s.DB.DriverName(), "sqlite") {
-		query := `ALTER TABLE federation_nodes_sync DROP CONSTRAINT IF EXISTS federation_nodes_sync_pkey`
-		if _, err = s.DB.ExecContext(ctx, query); err != nil {
-			if strings.Contains(err.Error(), "does not exist") {
-				return nil
-			}
-			return fmt.Errorf("failed to drop primary key constraint on federation_nodes_sync: %w", err)
+	var (
+		alterQry   string
+		rows       *sql.Rows
+		tableName  = "federation_nodes_sync"
+		driverName = s.DB.DriverName()
+	)
+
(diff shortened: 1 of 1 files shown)


server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -808,29 +808,70 @@ func fix_2024_09_03_dropFederationNodeSyncPrimaryKey(ctx context.Context, s *Sto
...
+			fmt.Sprintf("CREATE TABLE %s AS SELECT * FROM %s WHERE 1=0", tempTable, tableName),
+			fmt.Sprintf("INSERT INTO %s SELECT * FROM %s", tempTable, tableName),
+			fmt.Sprintf("DROP TABLE %s", tableName),
+			fmt.Sprintf("ALTER TABLE %s RENAME TO %s", tempTable, tableName),
+		}
+
+		for _, sql := range sqlStatements {
+			if _, err = s.DB.ExecContext(ctx, sql); err != nil {
+				return err
 			}
```
tjerman: Perhaps worth doing this in a transaction just in case? Else, we might need to check if any of these queries executed already

server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -808,29 +808,70 @@ func fix_2024_09_03_dropFederationNodeSyncPrimaryKey(ctx context.Context, s *Sto
...
+		}
+		return
+	}
+
+	// handle drop primary key for postgres, mysql and mssql
+	alterQry = `ALTER TABLE federation_nodes_sync DROP CONSTRAINT IF EXISTS federation_nodes_sync_pkey`
+	if _, err = s.DB.ExecContext(ctx, alterQry); err != nil {
+		if strings.Contains(err.Error(), "does not exist") {
+			return nil
+		}
```
tjerman: Do you need this if if the query already states `IF EXISTS`?

tjerman (changes requested):
Please make sure you've tested running Corteza on all supported database drivers

KinyaElGrande:
I have tested with MySql 8, Mssql Server 2017 and Postgres 15.

tjerman (approved):
lgtm 👍

PR #2198, opened 2025-09-24
2024.9.6 feature wf reminders

### LINKED ISSUE #2191: Add workflow functions and triggers for reminders
## References

The reminders service is implemented here: https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/service/reminder.go
You can take a look at the roles service (https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/service/role.go) to see how events are integrated into a service.

Events (system events) that get dispatched on the event bus are defined here https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/service/event/events.yaml

Workflow functions (system ones) are defined here https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/automation/roles_handler.yaml
The generated functions (https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/automation/roles_handler.gen.go) provide workflow metadata (name, arguments, responses, ...)
Here is an example implementation https://github.com/cortezaproject/corteza/blob/2024.9.x/server/system/automation/roles_handler.go

## What next

Update this issue (add comment, update original; up to you) to outline the changes you're going to do and make an action plan.

You can follow what I've done in the past (no need to do it exactly how I do it).

* https://github.com/cortezaproject/corteza/issues/1867
* https://github.com/cortezaproject/corteza/issues/2143

Mahjabin08:
Introduced reminder events in events.yaml and implemented initial lookup handler in reminders_handler.go and reminders_handler.yaml. Work in progress for full reminder workflow support.

### DIFF
--- a/server/system/automation/expr_types.go
+++ b/server/system/automation/expr_types.go
@@ -69,6 +69,30 @@ func CastToRole(val interface{}) (out *types.Role, err error) {
 	}
 }
 
+func CastToReminder(val interface{}) (out *types.Reminder, err error) {
+	switch val := val.(type) {
+	case expr.Iterator:
+		out = &types.Reminder{}
+		return out, val.Each(func(k string, v expr.TypedValue) error {
+			return assignToReminder(out, k, v)
+		})
+	}
+
+	switch val := expr.UntypedValue(val).(type) {
+	case *types.Reminder:
+		return val, nil
+	case map[string]interface{}:
+		out = &types.Reminder{}
+		m, _ := json.Marshal(val)
+		_ = json.Unmarshal(m, out)
+		return out, nil
+	case nil:
+		return &types.Reminder{}, nil
+	default:
+		return nil, fmt.Errorf("unable to cast type %T to %T", val, out)
+	}
+}
+
 func CastToTemplate(val interface{}) (out *types.Template, err error) {
 	switch val := val.(type) {
 	case expr.Iterator:
@@ -241,6 +265,11 @@ func (v *RenderOptions) Clone() (expr.TypedValue, error) {
 func (v *RenderedDocument) Clone() (expr.TypedValue, error) {
 	aux := *v.value
 	return NewRenderedDocument(&aux)
+}
+
+func (v *Reminder) Clone() (expr.TypedValue, error) {
+	aux := *v.value
+	return NewReminder(&aux)
 }
 
 func (v *Role) Clone() (expr.TypedValue, error) {
--- a/server/system/automation/expr_types.go
+++ b/server/system/automation/expr_types.go
@@ -268,6 +268,9 @@ func (v *RenderedDocument) Clone() (expr.TypedValue, error) {
 }
 
 func (v *Reminder) Clone() (expr.TypedValue, error) {
+	if v.value == nil {
+		return NewReminder(nil)
+	}
 	aux := *v.value
 	return NewReminder(&aux)
 }
--- a/server/system/automation/expr_types.go
+++ b/server/system/automation/expr_types.go
@@ -82,8 +82,13 @@ func CastToReminder(val interface{}) (out *types.Reminder, err error) {
 		return val, nil
 	case map[string]interface{}:
 		out = &types.Reminder{}
-		m, _ := json.Marshal(val)
-		_ = json.Unmarshal(m, out)
+		// m, _ := json.Marshal(val)
+		// _ = json.Unmarshal(m, out)
+		for k,v := range val {
+			if err = assignToReminder(out, k,v);err != nil{
+				return nil, err
+			}
+		}
 		return out, nil
 	case nil:
 		return &types.Reminder{}, nil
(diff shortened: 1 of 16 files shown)


server/system/automation/expr_types.go
```diff
@@ -68,6 +68,29 @@ func CastToRole(val interface{}) (out *types.Role, err error) {
...
+		})
+	}
+
+	switch val := expr.UntypedValue(val).(type) {
+	case *types.Reminder:
+		return val, nil
+	case map[string]interface{}:
+		out = &types.Reminder{}
+		m, _ := json.Marshal(val)
+		_ = json.Unmarshal(m, out)
```
tjerman: This is ok and it's how we usually did it...
I'd prefer we start avoiding this as it brings quite a bit of overhead.
Lets try to manually decode this arbitrary map into the Reminder type

server/system/service/event/events.yaml
```diff
@@ -130,3 +130,16 @@ system:data-privacy-request:
       immutable: true
   constraints:
     - name: role.name
+
+system:reminder:
+  # on: ['manual']
+  ba: ['create','update','delete','dismiss','snooze']
```
tjerman, line 136: Perhaps we could also add a hook to catch before/after a reminder is assigned

tjerman (commented):
A few notes from this intermediate code check

server/system/automation/expr_types.go
Mahjabin08: okay, on it

server/system/service/event/events.yaml
Mahjabin08: sure

Mahjabin08: Hi Tomaž, could you kindly mention an usecase for this, wouldn't this be already handled by the update hook?If the goal is to have hooks that trigger only when the assignment changes, should i create a separate assign method like the dismiss and snooze or just add the assigned hooks within the existing Update method when AssignedTo changes.

server/system/automation/reminders_handler.yaml
```diff
@@ -0,0 +1,134 @@
...
+  create:
+    meta:
+      short: Reminder create
+    labels:
+      <<: *labels
+      create: "step"
+    params:
+      # reminder:
+      #   <<: *reminder
+      #   required: true
```
tjerman: What's up with this?

server/system/automation/reminders_handler.go
```diff
@@ -0,0 +1,245 @@
...
+	return i, i.loader()
+}
+
+ func (h remindersHandler) create(ctx context.Context, args *remindersCreateArgs) (results *remindersCreateResults, err error) {
+	
+  	results = &remindersCreateResults{}
+	currentUser := intAuth.GetIdentityFromContext(ctx).Identity()
+
+	reminder := &types.Reminder{
+		Resource: "workflow-reminder",
```
tjerman: The resource states _what thing_ this reminder is attached to.
For example, if you're creating a reminder for records (how it was originally intended)

Have the resource provided via arguments

server/system/automation/reminders_handler.go
```diff
@@ -0,0 +1,245 @@
...
+		AssignedTo: args.AssignedTo,
+		AssignedBy: currentUser,
```
tjerman: This needs to be provided either via workflow arguments or defaulted to currentUser in case it was not provided via arguments.

server/system/service/event/events.yaml
```diff
@@ -130,3 +130,16 @@ system:data-privacy-request:
       immutable: true
   constraints:
     - name: role.name
+
+system:reminder:
+  # on: ['manual']
```
tjerman: Keep this one in just for consistency

server/system/service/reminder.go
```diff
@@ -24,6 +26,7 @@ type (
 		actionlog      actionlog.Recorder
 		store          store.Reminders
 		reminderSender reminderSender
+		eventbus 	   eventDispatcher //adding eventbus
```
tjerman:
```suggestion
		eventbus 	   eventDispatcher
```

tjerman (changes requested):
Looks nice! Take a look at the few comments, please!
After that's sorted, do make sure you test it thuraly

server/system/automation/reminders_handler.yaml
Mahjabin08: At first, I used this to take the args, but that was not very user friendly, so defined the args directly, but kept it if I needed it later, going to remove in the next commit.

server/system/automation/reminders_handler.go
Mahjabin08: Noted!

server/system/service/event/events.yaml
Mahjabin08: Okay

server/system/service/reminder.go
Mahjabin08: sure

tjerman (approved):
lgtm!

PR #2207, opened 2025-10-02
2024.9.6 fix length function arr workflow

### LINKED ISSUE #2183: Length function not working on Array type in workflows
https://forum.cortezaproject.org/t/length-array-error/2840/2

Mahjabin08:
Based on upstream/2024.9.6. Fixes #2183 by processing typed values in the pkg/expr/func_generic.go.


server/compose/automation/records_handler.gen.go
tjerman, line 1: What's going on with this file? Did you modify it manually, or is there a formatting mismatch?

server/pkg/expr/func_generic.go
```diff
@@ -36,6 +37,7 @@ func length(i interface{}) int {
 	if isEmpty(i) {
 		return 0
 	}
+	i = UntypedValue(i)
```
tjerman: This sounds reasonable, but what about the `isEmpty` function above? As far as I see, it also runs on the original type, which could be problematic.

```go
func isEmpty(i interface{}) bool {
	if isNil(i) {
		return true
	}

	if c, ok := i.(empty); ok {
		return c.IsEmpty()
	}

	switch reflect.TypeOf(i).Kind() {
	case reflect.Slice, reflect.Array, reflect.Map:
		return reflect.ValueOf(i).Len() == 0
	}

	return reflect.ValueOf(i).IsZero()
}
```

tjerman (changes requested):
See comments.
Additionally, let's think about adding some extra tests for these cases

server/compose/automation/records_handler.gen.go
Mahjabin08: I didn't actually do anything there, but when I ran the codegen-legacy command, it just changed, and from what i see, the importing lines sequence changed only, nothing else changed here

server/pkg/expr/func_generic.go
Mahjabin08: Right, I should've unwrapped it before passing it to the isEmpty function, thanks

Mahjabin08: fixing it

Mahjabin08:
> See comments.
> Additionally, let's think about adding some extra tests for these cases

Sure, just to be sure,are we talking about adding test cases to the func_generic_test.go or any validations?

tjerman (approved):
Looks good but please remove the unchanged files (records_handler gen go and nodeHandshake go)
We'll have to do something about formatting though :/

Mahjabin08:
> Looks good but please remove the unchanged files (records_handler gen go and nodeHandshake go) We'll have to do something about formatting though :/

Got it, I will remove those files. Just to confirm, what exactly do you mean by formatting here? Want to make sure I am seeing it the same way :)

PR #2240, opened 2025-11-17
Add label references for namespace and module in workflow

Mahjabin08:
- **Add rest api definition for labels listing endpoint**
- **Add label listing endpoint**
- **Add functionality for handling single and multiple label value(wip)**
- **Change label value type(WIP)**
- **Add LabelValue expression type**
- **Fix value type for labels in wf yaml files**
- **Fix label search for LabelValue type change**
- **Refactor label value type to support single and multiple values**

### DIFF
--- a/server/automation/service/trigger.go
+++ b/server/automation/service/trigger.go
@@ -106,7 +106,7 @@ func (svc *trigger) Search(ctx context.Context, filter types.TriggerFilter) (rr 
 		if len(filter.Labels) > 0 {
 			labelStrings := make(map[string]string,len(filter.Labels))
 			for k, v := range filter.Labels {
-				labelStrings[k] = v.Value
+				labelStrings[k] = v.Val
 			}
 			filter.LabeledIDs, err = label.Search(
 				ctx,
--- a/server/automation/service/trigger.go
+++ b/server/automation/service/trigger.go
@@ -104,11 +104,15 @@ func (svc *trigger) Search(ctx context.Context, filter types.TriggerFilter) (rr 
 		}
 
 		if len(filter.Labels) > 0 {
+			labelStrings := make(map[string]string,len(filter.Labels))
+			for k, v := range filter.Labels {
+				labelStrings[k] = v.Value
+			}
 			filter.LabeledIDs, err = label.Search(
 				ctx,
 				svc.store,
 				types.Trigger{}.LabelResourceKind(),
-				filter.Labels,
+				labelStrings,
 			)
 
 			if err != nil {
--- a/server/automation/service/trigger.go
+++ b/server/automation/service/trigger.go
@@ -104,15 +104,11 @@ func (svc *trigger) Search(ctx context.Context, filter types.TriggerFilter) (rr 
 		}
 
 		if len(filter.Labels) > 0 {
-			labelStrings := make(map[string]string,len(filter.Labels))
-			for k, v := range filter.Labels {
-				labelStrings[k] = v.Val
-			}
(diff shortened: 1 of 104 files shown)


server/automation/service/trigger.go
```diff
@@ -104,11 +104,15 @@ func (svc *trigger) Search(ctx context.Context, filter types.TriggerFilter) (rr
...
+			for k, v := range filter.Labels {
+				labelStrings[k] = v.Val
+			}
 			filter.LabeledIDs, err = label.Search(
 				ctx,
 				svc.store,
 				types.Trigger{}.LabelResourceKind(),
-				filter.Labels,
+				labelStrings,
 			)
```
tjerman: I'd prefer we change the label.search function signature so it expects a slice of labels instead of a map.
This would remove all of the slice -> map mappings

tjerman (commented):
Looking good; a potential structural change to consider

server/pkg/label/types/type.go
```diff
@@ -9,14 +20,16 @@ type (
...
 	}
 
 	LabelFilter struct {
 		Kind       string
 		ResourceID []uint64
-		Filter     map[string]string
+		Filter     map[string][]string
 		Limit      uint
+		Name      string
+		Value 	  []string //use a slice of string
```
tjerman:
```suggestion
		Value 	  []string
```

server/pkg/label/types/type.go
```diff
@@ -33,13 +46,61 @@ func (set LabelSet) ResourceIDs() (rr []uint64) {
...
 }
 
-func (set LabelSet) FilterByResource(kind string, ID uint64) map[string]string {
-	var kv = make(map[string]string)
+func (set LabelSet) FilterByResource(kind string, ID uint64) map[string]LabelValue {
+	var kv = make(map[string]LabelValue)
 	for _, label := range set {
 		if kind == label.Kind && ID == label.ResourceID {
-			kv[label.Name] = label.Value
+					kv[label.Name] = label.Value
```
tjerman: It seems like your code formatter is not configured correctly.
Can you please take a look if you have it enabled? Make sure you're using whatever standard the built-in VSCode plugin has

server/pkg/label/label.go
```diff
@@ -2,38 +2,41 @@ package label
...
 )
 
 // Changed checks if label maps are same or different
-func Changed(old, new map[string]string) bool {
+func Changed(old, new map[string]types.LabelValue) bool {
 	for k := range old {
 		if _, has := new[k]; !has {
 			return true
-		} else if new[k] != old[k] {
+		} else if !reflect.DeepEqual(new[k],old[k]) {
```
tjerman: No.
Write out a comparison function.

server/pkg/label/label.go
```diff
@@ -2,38 +2,41 @@ package label
...
 			return true
 		}
 	}
 
 	for k := range new {
 		if _, has := old[k]; !has {
 			return true
-		} else if new[k] != old[k] {
+		} else if !reflect.DeepEqual(new[k],old[k]) {
```
tjerman: Same as above

tjerman (changes requested):
Almost there!

server/pkg/label/types/type.go
Mahjabin08: Hi, you are right, my formatter wasnt configured properly, so I have turned it on, it formatted the file. But for this specific line 53 its showing the existing formatting is correct

Mahjabin08:
Created a new pr

PR #2245, opened 2025-11-27
Add label references for namespace and module in workflow

### LINKED ISSUE #2209: Add an endpoint to list and search over system-defined labels
We need to allow the user interface to get a list of all labels in order to do some things...

Currently, RBAC does not explicitly check access to specific labels.
Labels are attached directly to a resource, which essentially makes them RBAC-protected.

We never explicitly stated what labels should be used for so we might need to be careful we don't leak sensitive information.
On the other hand, we might need to allow users access to all labels.

Here are some options I can think of:
* Expose labels _publicly_, disclose the fact that labels can be accessed by anyone who has access to Corteza.
* Expand label resource identifier to use the full resource identifier, which is better than nothing ,but it won't support contextual roles.
* Define an endpoint per component (compose, system, ...) and define a "can search labels" for a specific component. This gives us at least some control.

@Mahjabin08, Can you please consider this and provide some feedback (in the comments here)?
We'll probably tag team this issue.

Mahjabin08:
- **Add rest api definition for labels listing endpoint**
- **Add label listing endpoint**
- **Add functionality for handling single and multiple label value(wip)**
- **Change label value type(WIP)**
- **Add LabelValue expression type**
- **Fix value type for labels in wf yaml files**
- **Fix label search for LabelValue type change**
- **Refactor label value type to support single and multiple values**
- **fix label query in filter.go**
- **fix path for list endpoint**
- **fix filter.go and type.go**
- **Refactor label.search to accept LabelValue map directly**
- **Fix tests for LabelValue Type compatibility**
- **Fix label filtering and standardize label value format for API**
- **Fix suspended and deleted user FE resolution**
- **Add ignored private folder**
- **Remove chatty development log**
- **Fix module issues tab color not applying**
- **Add Equal method for LabelValue comparison and fix code formatting**

### DIFF
--- a/server/store/adapters/rdbms/upgrade_fixes.go
+++ b/server/store/adapters/rdbms/upgrade_fixes.go
@@ -54,6 +54,7 @@ var (
 		fix_2024_09_03_renameFederationNodeSyncNodeID,
 		fix_2024_09_03_renameFederationNodeSyncComposeID,
 		fix_2024_09_03_addFederationNodeSyncNodeIDIndex,
+		fix_2024_9_7_migrateLabelsValueToJsonb,
 	}
 
 	fixesPost = []func(context.Context, *Store) error{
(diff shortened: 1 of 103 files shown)


server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -1020,7 +1021,133 @@ func fix_2024_09_03_addFederationNodeSyncNodeIDIndex(ctx context.Context, s *Sto
...
+func fix_2024_9_7_migrateLabelsValueToJsonb(ctx context.Context, s *Store) (err error) {
+	var (
+		log        = s.log(ctx)
+		driverName = s.DB.DriverName()
+		columnType string
+		rows       *sql.Rows
+	)
+
+	switch {
+	case strings.HasPrefix(driverName, "postgres"):
```
tjerman: Have each case body in a separate function

server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -1020,7 +1021,133 @@ func fix_2024_09_03_addFederationNodeSyncNodeIDIndex(ctx context.Context, s *Sto
...
+  			WHERE value IS NOT NULL 
+  			  AND value NOT LIKE '{%'`
+
+		if _, err = s.DB.ExecContext(ctx, updateQuery); err != nil {
+			return err
+		}
+
+		log.Info("successfully migrated labels.value data format")
+
+	case strings.HasPrefix(driverName, "sqlserver"):
```
tjerman: This one doesn't check if it's already ok; intentional?

server/store/adapters/rdbms/upgrade_fixes.go
```diff
@@ -1020,7 +1021,133 @@ func fix_2024_09_03_addFederationNodeSyncNodeIDIndex(ctx context.Context, s *Sto
...
+		if _, err = s.DB.ExecContext(ctx, alterQuery); err != nil {
+			return err
+		}
+
+		log.Info("successfully migrated labels.value column to json")
+
+	case strings.HasPrefix(driverName, "sqlite"):
+		log.Info("migrating labels.value data format for sqlite")
+
+		updateQuery := `
```
tjerman: This one doesn't check if it's already ok; intentional?

tjerman (changes requested):

PR #2248, opened 2025-12-04
Fix Totp not updating without server restart

### LINKED ISSUE #2239: When I try to update a TOTP Issuer, the changes do not reflect until a server restarts

### Version of Corteza

2024.9.6

### Current Behavior

In the admin webapp, under auth settings, "Enable TOTP" section; the changes to the "Issuer" do not reflect unless I restart the server.

The most probable cause:

This line https://github.com/cortezaproject/corteza/blob/2024.9.6/server/auth/handlers/handle_mfa_totp.go#L112 is referencing settings which get passed in over here https://github.com/cortezaproject/corteza/blob/2024.9.6/server/auth/auth.go#L198 which get reloaded over here https://github.com/cortezaproject/corteza/blob/2024.9.6/server/auth/auth.go#L350

Since https://github.com/cortezaproject/corteza/blob/2024.9.6/server/app/boot_levels.go#L641 does not reuse the previous struct value, https://github.com/cortezaproject/corteza/blob/2024.9.6/server/auth/handlers/handle_mfa_totp.go#L112 does not see the changed value as it relies on pointers

### Expected Behavior

This change should not expect a server restart.

### Steps To Reproduce

1. Go to the admin webapp and navigate to the settings in question
2. Make sure your mobile TOTP is enabled
3. in a new tab, go to `/auth/mfa/totp/setup` and inspect the QR code (you can use an online decoder)
4. Update the issuer option
5. Refresh the page with the QR code and re-inspect it

### Environment and versions

```markdown

```

### Anything else?

@Mahjabin08 let's fix this one together, a nice _palate cleanser_.
Since our code base listens to setting changes and calls an update function, we should probably respect that decision and provide additional hooks to reflect the changes everywhere.

The issue in question refers to just one place, but it most likely occurs elsewhere.

Mahjabin08:


tjerman (approved):
If tested and works as intended, looks good to me 👍

PR #2263, opened 2025-12-23
Fix uint64 values to be converted to strings in response to frontend

### LINKED ISSUE #2257: Fix ID's rounding in revision changes values.
If you look at ownedBy field or other values that are ID's, they get rounded up, therefore resulting in wrong values.
Those need to be strings like elsewhere so we can properly use them.

Mahjabin08:
- **Add step option to number field**
- **Fix uint64 values to be converted to strings in response to frontend**

### DIFF
--- a/server/pkg/revisions/revisions.go
+++ b/server/pkg/revisions/revisions.go
@@ -1,6 +1,7 @@
 package revisions
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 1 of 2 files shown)


tjerman (changes requested):
The idea sounds fine, but I'm not a fan of you bringing the code from `dal.IteratorEncodeJSON` directly into the rest middleware.
I'm also not a fan of you defining a receiver function to handle resource -> response encoding -- can't we utilise JSON tags on the structs?

tjerman (approved):
lgtm; just double-check if everything works as expected. You could probably add some unit tests to check JSON encoding

Mahjabin08:
okay, will do that asap

PR #2269, opened 2026-01-04
Fix websocket panics on session disconnect

### LINKED ISSUE #2223: Websocket write panic on redirect to login screen

### Version of Corteza

2024.9.5

### Current Behavior

I am seeing periodic websocket DEBUG panics when running Corteza with `LOG_DEBUG=true` and `WEBSOCKET_LOG_ENABLED=true`. This seems to be occurring when a user is logged out and Corteza redirects them from the base domain (e.g. http://local.cortezaproject.org:18080/) to the login page (e.g. http://local.cortezaproject.org:18080/auth/login).

Using the [example docker-compose setup](https://docs.cortezaproject.org/corteza-docs/2024.9/devops-guide/examples/deploy-offline/multi-mysql.html) with the only .env modifications being changing `VERSION` to 2024.9.5 and setting `WEBSOCKET_LOG_ENABLED=true` and `LOG_DEBUG=true` , I get the following logs after browsing to http://local.cortezaproject.org:18080/:

```
server-1  | 09:15:36.324        DEBUG   http    entering active state
server-1  | 09:16:00.000        DEBUG   scheduler       started
server-1  | 09:16:12.317        DEBUG   websocket.session       recovering from websocket read panic    {"id": "465506884581982209", "recovered-error": "repeated read on failed websocket connection"}
server-1  | 09:16:17.242        INFO    websocket.session       closing unidentified connection {"id": "465506884581982209"}
server-1  | 09:16:17.242        INFO    websocket.session       disconnected    {"id": "465506884581982209"}
server-1  | 09:16:17.242        DEBUG   websocket.session       recovering from websocket write panic   {"id": "465506884581982209", "recovered-error": "runtime error: invalid memory address or nil pointer dereference"}
```

### Expected Behavior

Websocket errors are fine, but there probably shouldn't be a `invalid memory address or nil pointer dereference` error in the server.

### Steps To Reproduce

1. Grab docker-compose.yaml and .env from https://docs.cortezaproject.org/corteza-docs/2024.9/devops-guide/examples/deploy-offline/multi-mysql.html
2. Modify .env to set `VERSION=2024.9.5` and add `WEBSOCKET_LOG_ENABLED=true` and `LOG_DEBUG=true`
3.  `docker compose up` and wait for the server to come up
4.  browse to http://local.cortezaproject.org:18080/

There should be an immediate websocket read panic, followed a few seconds later by the websocket write panic that references `runtime error: invalid memory address or nil pointer dereference`

### Environment and versions

```markdown
########################################################################################################################
# docker-compose supports environment variable interpolation/substitution in compose configuration file
# (more info: https://docs.docker.com/compose/environment-variables)

########################################################################################################################
# General settings
DOMAIN=local.cortezaproject.org:18080
VERSION=2024.9.5

########################################################################################################################
# Database connection
DB_DSN=dbuser:dbpass@tcp(db:3306)/dbname?collation=utf8mb4_general_ci

########################################################################################################################
# Server settings

# Running all-in-one and serving web applications directly from server container
HTTP_WEBAPP_ENABLED=true

# Disabled, we do not need detailed persistent logging of actions in local env
ACTIONLOG_ENABLED=false

########################################################################################################################
# SMTP (mail sending) settings

# Point this to your local or external SMTP server if you want to send emails.
# In most cases, Corteza can detect that SMTP is disabled and skips over sending emails without an error
#SMTP_HOST=smtp-server.example.tld:587
#SMTP_USER=postmaster@smtp-server.example.tld
#SMTP_PASS=this-is-your-smtp-password
#SMTP_FROM='"Demo" <info@your-demo.example.tld>'
WEBSOCKET_LOG_ENABLED=true
LOG_DEBUG=true
```

### Anything else?

I am also seeing this in my production Corteza install, so it does not seem to be specific to the `DOMAIN` setting or having HTTPS or not. I have reproduced this on multiple browsers (Brave, Firefox, Safari) in normal and private mode, so it does not seem to be related to browser or extensions.

Browsing directly to the login page (http://local.cortezaproject.org:18080/auth/login in my example) does NOT trigger the websocket error, so I suspect it's related to the redirect. Browsing to http://local.cortezaproject.org:18080/ from a freshly opened browser or from a browser with website data cleared still triggers the websocket error, so it should be unrelated to any previous websocket connection.

Admittedly this is only a minor annoyance to me (a few panic errors cluttering up my logs, no noticeable impact to users) but maybe there are larger implications to hitting an invalid memory address or nil pointer dereference, especially for a logged out user.

jackmaninov:
## Summary

Fixes multiple race conditions and error handling issues in websocket session management that caused panics during unauthenticated visitor access and login redirects.

## Problem

When unauthenticated users connect via WebSocket (e.g., during redirect from base URL to login page), several issues can occur:

1. **Nil pointer dereference**: `disconnect()` sets `conn = nil` while `read()`/`write()` operations are pending, causing panic when they try to use the connection
2. **Read loop spinning**: `errHandler()` returned `nil` for close errors, causing the read loop to continue instead of exiting. After ~1000 iterations, gorilla/websocket triggers a safety panic
3. **Send on closed channel**: `Write()` could attempt to send to `s.send` channel after `disconnect()` closed it
4. **Noisy error logs**: Expected close conditions (1006 abnormal closure, "close sent") were logged as errors

## Changes

- Add nil checks in `read()` and `write()` after acquiring lock to prevent nil pointer dereference
- Return `net.ErrClosed` from `errHandler()` for close errors to exit read/write loops cleanly
- Add `atomic.Bool` closed flag to session, set before closing channels in `disconnect()`
- Check closed flag in `Write()` before attempting channel send
- Handle additional expected close codes (1005, 1006) and "close sent" error in `errHandler()`

## Testing

- Existing websocket tests pass
- Manually verified with unauthenticated visitor access and login redirect flow

Fixes #2223

### DIFF
--- a/server/pkg/websocket/session.go
+++ b/server/pkg/websocket/session.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/cortezaproject/corteza/server/pkg/auth"
@@ -39,9 +40,10 @@ type (
 	session struct {
 		l sync.RWMutex
 
-		id   uint64
-		once sync.Once
-		conn conection
+		id     uint64
+		once   sync.Once
+		conn   conection
+		closed atomic.Bool
 
 		ctx       context.Context
 		ctxCancel context.CancelFunc
@@ -91,6 +93,9 @@ func (s *session) Identity() auth.Identifiable {
 func (s *session) disconnect() {
 	s.l.Lock()
 	defer s.l.Unlock()
+
+	// Mark session as closed before closing channels
+	s.closed.Store(true)
 
 	// Cancel context
 	s.ctxCancel()
@@ -189,6 +194,11 @@ func (s *session) read() (raw []byte, err error) {
 
 	s.l.RLock()
 	defer s.l.RUnlock()
+
+	// Check if connection was closed by disconnect()
+	if s.conn == nil {
+		return nil, net.ErrClosed
+	}
 
 	if _, raw, err = s.conn.ReadMessage(); err != nil {
 		return nil, errHandler("websocket read failed", err)
@@ -291,6 +301,11 @@ func (s *session) write(t int, msg []byte) (err error) {
 		}
 	}()
 
+	// Check if connection was closed by disconnect()
+	if s.conn == nil {
+		return net.ErrClosed
+	}
+
 	if err = s.conn.SetWriteDeadline(time.Now().Add(s.config.Timeout)); err != nil {
 		return fmt.Errorf("deadline error: %w", err)
 	}
@@ -324,6 +339,11 @@ func (s *session) authenticate(p *payloadAuth) error {
 
 // sendBytes sends byte to channel or timeout
 func (s *session) Write(p []byte) (int, error) {
+	// Check if session is closed before attempting to send
+	if s.closed.Load() {
+		return 0, net.ErrClosed
+	}
+
 	defer func() {
 		if recovered := recover(); recovered != nil {
 			s.logger.Debug("recovering from websocket write panic", zap.Any("recovered-error", recovered))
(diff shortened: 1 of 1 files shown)


jackmaninov:
Not sure if you have an AI generated code policy, but this was done supervised and tested on my production environment.

Fajfa:
If the code is good and resolves the issue, then it will be accepted. Thanks for contributing.

Fajfa:
Merged into 2024.9.10

PR #2270, opened 2026-01-04
feat(auth): OIDC compliance improvements

jackmaninov:
## Summary

This PR improves OIDC (OpenID Connect) compliance for Corteza's OAuth2/OIDC implementation:

- **UserInfo endpoint** (`/auth/oauth2/userinfo`): Implements OIDC Core 1.0 Section 5.3 compliant endpoint that returns user claims based on token scopes (openid, profile, email)
- **Discovery document**: Adds spec-compliant mount point at `/auth/.well-known/openid-configuration` (the original `/.well-known/openid-configuration` is preserved for backward compatibility)
- **Silent authentication** (`prompt=none`): Implements OIDC Core 1.0 Section 3.1.2.1 for silent authentication flows, returning appropriate `login_required` or `interaction_required` errors
- **Enhanced discovery metadata**: Adds `userinfo_endpoint`, `claims_supported`, and `scopes_supported` to the discovery document

## Related Issues

- Forum request: https://forum.cortezaproject.org/t/openid-connect-userinfo/816

## Test Plan

- [x] Added comprehensive tests for UserInfo endpoint (7 test cases covering full claims, scope filtering, error handling)
- [x] Added tests for `prompt=none` error redirects (4 test cases)
- [x] Added tests for UserID serialization
- [x] All existing tests pass

## Breaking Changes

None. The original `/.well-known/openid-configuration` endpoint is preserved for backward compatibility.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/app/servers.go
+++ b/server/app/servers.go
@@ -155,6 +155,12 @@ func (app *CortezaApp) mountHttpRoutes(r chi.Router) {
 	}()
 
 	func() {
+		// Mount OIDC Discovery document at root level for backward compatibility
 		r.Handle("/.well-known/openid-configuration", app.AuthService.WellKnownOpenIDConfiguration())
+
+		// Also mount at /auth/.well-known/openid-configuration for OIDC compliance
+		// Per OIDC Discovery 1.0, the well-known endpoint must be at {issuer}/.well-known/openid-configuration
+		// Since our issuer URL is {baseURL}/auth, clients need to discover at /auth/.well-known/...
+		r.Handle("/auth/.well-known/openid-configuration", app.AuthService.WellKnownOpenIDConfiguration())
 	}()
 }
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -436,19 +436,94 @@ func dirCheck(path string) (err error) {
 	return
 }
 
+// WellKnownOpenIDConfiguration returns the OIDC Discovery document
+// per OpenID Connect Discovery 1.0 specification
 func (svc service) WellKnownOpenIDConfiguration() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		_ = json.NewEncoder(w).Encode(map[string]interface{}{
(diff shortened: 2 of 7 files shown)


jackmaninov:
Same story as my previous PR, this was written with claude-code but has been heavily tested on my production environment. Could probably use some sanity testing with a client other than go-oidc.

Fajfa:
Oh wow, thanks for the contribution, will be reviewed and merged if all is good.

Fajfa:
Thanks — the UserInfo endpoint and the discovery document were genuinely missing, and we've merged that part (plus the ID-token iat/email_verified additions) for 2024.9.10. https://github.com/cortezaproject/corteza/commit/d7a635f274fbcbcf052150f0bd600b625b67d5e2

Three changes we made while integrating:
1. Issuer. ID tokens are issued with iss = AUTH_BASE_URL (https://host/auth). Stripping /auth from the issuer at the root discovery URL makes it disagree with the tokens, so existing clients — which all use that URL — would start rejecting them. Both endpoints now report AUTH_BASE_URL unchanged. For the same reason we kept deriving the endpoint URLs from AUTH_BASE_URL directly, since it isn't guaranteed to end in /auth.
2. Scopes. auth.CheckScope(scope, "profile", "openid") is an OR, so a token scoped only openid was getting name, email, picture and locale. UserInfo now requires profile for profile claims and email for email claims. ID-token claims that shipped before (email, user_id) stay unconditional so existing integrations don't break.
3. prompt=none is held back. As written, it redirects to the redirect_uri from the query string before anything checks that URI belongs to the client — go-oauth2 only validates it later in GetAuthorizeToken. On a local build, an unauthenticated request produced 302 → https://evil.example.com/steal?error=login_required&state=… with nothing but a valid client_id. That's an open redirect on the authorize endpoint, and OIDC 3.1.2.6 requires not redirecting to an unvalidated URI. 

Also dropped picture (the URL is built off the auth base rather than the API base, and avatar attachments need a signed URL, so it 404s either way)

PR #2271, opened 2026-01-06
Fix OIDC role security and invalidate tokens on provider/role changes

### LINKED ISSUE #2204: OIDC user can bypass "permitted roles"

### Version of Corteza

2024.9.4

### Current Behavior

It's seems the front end is unable to retrieve the user roles. This make my page layouts hang forever if I define a required role.

### Expected Behavior

The front end should be able to retrieve user roles and apply conditional display.

### Steps To Reproduce

1. Create a module
2. Create a page
3. Under page layout, set any allowed roles (except null)
4. Check if the layout load

### Environment and versions

```markdown
I tested this issue with both chrome and firefox
```

### Anything else?

Related console log :
Image

Remark : 
It seems to apply to both page layout role and block.

I also realized that the user.roles conditional expression also return undefined. I don't know if it should work normally but it's still a suggested auto-complete.

here is the related UI component : 
Image

Mahjabin08:


Mahjabin08:
will remove the debug log soon

PR #2285, opened 2026-01-29
Fix workflow delay step label and add format hint

### LINKED ISSUE #2250: Workflow - Delay function - Confusing expected format

### Version of Corteza

2024.9.6

### Current Behavior

The delay function doesn't show clearly how it's meant to be used. It require trial and error from the user in it's current state.

Image

### Expected Behavior

The expected format is a string following the Duration type format, according to my finding at least. 

I propose reusing the workflow argument block with a Duration type indicator : 

Image

### Steps To Reproduce

_No response_

### Environment and versions

```markdown

```

### Anything else?

_No response_

Mahjabin08:


client/web/workflow/src/components/Configurator/Delay.vue
```diff
@@ -18,6 +18,7 @@
       >
         <b-form-group
           :label="$t('general:offset-expression')"
```
Fajfa: $t('configurator:delay.duration.label')

client/web/workflow/src/components/Configurator/Delay.vue
```diff
@@ -18,6 +18,7 @@
...
+          :description="$t('configurator:delay.format-hint')"
```
Fajfa: $t('configurator:delay.duration.description')

locale/en/corteza-webapp-workflow/general.yaml
```diff
@@ -24,7 +24,7 @@ new-workflow: New Workflow
 none: None
 now: Now
 offset: Offset
-offset-expression: Offset (expression)
+offset-expression: Duration
```
Fajfa: Can we actually remove this (if its not used anywhere else)

locale/en/corteza-webapp-workflow/configurator.yaml
```diff
@@ -25,3 +25,5 @@ run-as:
 sub-workflow:
   label: Subworkflow
   description: Subworkflows can't have enabled triggers
+delay:
```
Fajfa, line 28: Match this file to the above comments

Fajfa (changes requested):

locale/en/corteza-webapp-workflow/general.yaml
Mahjabin08: I did a search, its not used anywhere else, so I think we can safely remove it

PR #2309, opened 2026-03-21
Created collapsible record and collapsible record list

Toocky:

New block types

Collapsible Record:

configurable header with title and subtitle ( subtitle is selectable to include when minimised)
able to specify title alignment left centre right
3 choices for the main part of the record,
Single body of rich text the full width of the block
A collection of record fields ie current record
both body of text and selection of fields, with configurable position above below
can default collapsed
Collapsible Record List:

As the above collapses into a single header its possible to create a list of collapsed items
all the other features of a record list, sorting paging etc
This could allow for the creation of effective history record lists, ie a case history with the records sorted by date

https://github.com/cortezaproject/corteza/issues/2308
https://forum.cortezaproject.org/t/collapsible-record-collapsible-record-list-blocks/3080

image

to allow manipulating the alignment in the title and subtitle i have modified the expressions to handle L[], C[], R[] to adjust alignment. 
i.e. 
L[${record.values.title}]     to align left 
C[example text]   to align centre
R[${record.values.status} ${record.values.date}]  to align right 

all of these can be used in a single title, but they must be used once each and in the correct order, you can not use C[], R[],   L[],  it will ignore them and default to all left alignment 

image

image
image
image
image

The single record block when collapsed will also redraw the page and draw up blocks below it assuming there is nothing else blocking the blocks below it

the list block is intended to be identical to the above just a list of the records collapsed to be expanded as the user wishes 
image

image

There is room for improvement theres currently no editing options but for now i'm putting the PR in for any comments / suggestions as i want to get this pushed for my own use to see how it works in production also i need to rebase now there's been a new release 

 - [x] Code builds
 - [ ] All existing tests pass


Toocky:
Rebased to 2024.9.8

